Issue #3, opened 2019-12-20
xy_7


clyne:
Description: Line plotting with two axes

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/xy_7.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/xy.shtml

khallock:
`xy_7_2.py` was added in #28, but only one of the two plots from the [NCL xy_7.ncl example](https://www.ncl.ucar.edu/Applications/xy.shtml#ex7) is being generated: https://geocat-examples.readthedocs.io/en/latest/auto_examples/XY/xy_7_2_lg.html

In the interest of being consistent with the original NCL example, we may want to generate the second plot and also consider renaming the Python script to `xy_7.py` (or `NCL_xy_7.py`) to match the original script.

PR #28, opened 2020-01-23
Add XY_7_2 plot

andersy005:
Towards #3

- Original NCL:

- Python Version:

sphx_glr_xy_7_2_lg_001

### DIFF
--- a/Plots/XY/xy_7_2_lg.py
+++ b/Plots/XY/xy_7_2_lg.py
@@ -3,30 +3,31 @@ xy_7_2_lg.py
 ===============
 An example of a double y plot: two separate line with their own unique axis.
 
-https://www.ncl.ucar.edu/Applications/Images/xy_7_2_lg.png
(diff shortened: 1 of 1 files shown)


erogluorhan:
Comparison of plots:
1) Python version should divide x-axis into 5 minor-ticks instead of 3 between adjacent major ticks (e.g. 1970.0 and 1970.5

2) Python plot is missing the very last curve change (which is the closest point to 1973.0) on the red dashed line plot that is apparent in NCL plot.

3) If identical dash-spaces with NCL plot is of interest, the following code:
`ax2.plot(ds.time, ds.P, color=color, linestyle=linestyle, linewidth=2.0)`
can be exchanged with the following one:
`ax2.plot(ds.time, ds.P, color=color, linestyle=linestyle, dashes=[6.5, 3.7], linewidth=2.0)`

andersy005:
sphx_glr_xy_7_2_lg_001 (1)

Issue #14, opened 2019-12-20
vector_1


clyne:
Description: Vectors and contours

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/vector_1.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/vector_1_lg.png

mnlevy1981:
closed via #29

PR #29, opened 2020-01-23
Python version of vector1.ncl

mnlevy1981:
@maboualidev and I converted [vector1.ncl](https://www.ncl.ucar.edu/Applications/Scripts/vector_1.ncl) to the python template. We use `contourf()` for the color contour of SST and `quiver()` for the vector plot.

### DIFF
--- a/Plots/Vectors/vector_1.py
+++ b/Plots/Vectors/vector_1.py
@@ -58,7 +58,7 @@ def add_lat_lon_ticklabels(ax):
     ax.xaxis.set_major_formatter(lon_formatter)
     ax.yaxis.set_major_formatter(lat_formatter)
 
-def nclize_axis(ax):
+def nclize_axis(ax, minor_per_major=3):
     """
     Utility function to make plots look like NCL plots
     """
@@ -66,8 +66,8 @@ def nclize_axis(ax):
 
     ax.tick_params(labelsize="small")
     ax.minorticks_on()
-    ax.xaxis.set_minor_locator(tic.AutoMinorLocator(n=3))
-    ax.yaxis.set_minor_locator(tic.AutoMinorLocator(n=3))
+    ax.xaxis.set_minor_locator(tic.AutoMinorLocator(n=minor_per_major))
+    ax.yaxis.set_minor_locator(tic.AutoMinorLocator(n=minor_per_major))
 
     # length and width are in points and may need to change depending on figure size etc.
     ax.tick_params(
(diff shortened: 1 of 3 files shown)


mnlevy1981:
Fixes #14

mnlevy1981:
Requires #37 to build correctly

Plots/Vectors/vector_1.py
```diff
@@ -0,0 +1,149 @@
...
+# Read in grid information
+lat_sst = sst_in['lat']
+lon_sst = sst_in['lon']
+lat_uv = uv_in['lat']
+lon_uv = uv_in['lon']
+
+# Read SST and U, V
+sst = sst_in['SST'].isel(time=ind_sst)
+u = uv_in['U'].isel(time=ind_uv, lev=0)
+v = uv_in['V'].isel(time=ind_uv, lev=0)
```
kmpaul: Can we use `sel` here, instead of `isel`?  I realize that this matches explicitly what is done in the NCL script, but that's just because NCL doesn't have this functionality.

kmpaul:
@mnlevy1981 Nice!

Plots/Vectors/vector_1.py
mnlevy1981: I've made this change -- it reduces the python line count (and avoids variables like `ind_sst`), though it adds a little complication in that we need `date` to be a dimension. Honestly, that's an issue with the formatting of the original `nc` file rather than anything code-related...

Plots/Vectors/vector_1.py
```diff
@@ -0,0 +1,148 @@
...
+               width=.0025, scale=(4.0/.045), zorder=2)
+
+# Draw legend for vector plot
+qk = ax.quiverkey(Q, 0.85, 0.9, 4, r'4 $m/s$', labelpos='N',
+                  coordinates='figure', color='black')
+
+# Draw SST contours
+plt.cm.register_cmap('BlAqGrYeOrReVi200', truncate_colormap(cmaps.BlAqGrYeOrReVi200, minval=0.08, maxval=0.96, n=len(levels)))
+cmap = plt.cm.get_cmap('BlAqGrYeOrReVi200', 50)
+cf = ax.contourf(lon_sst, lat_sst, sst, extend='both', levels=levels,
```
dcherian: really minr: This could be `sst.plot.contourf(x="lon", y="lat", ...)` and avoid extra unnecessary variables (which also separate the metadata from the data...)

PR #34, opened 2020-01-23
Vector_3

### LINKED ISSUE #16: vector_3
Description: basic vectors

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/vector_3.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/vector_3_lg.png

mnlevy1981:
@maboualidev and I converted [vector_3.ncl](https://www.ncl.ucar.edu/Applications/Scripts/vector_3.ncl) to the python template. We use `quiver()` for the vector plot and rely on @dcherian's `nclize_axis()` and `add_lat_lon_ticklabels()` functions.

Fixes #16 

Based on the same branch used in #29 -- after that PR is accepted, this should show just a single commit / new file.

### DIFF
--- a/Plots/Vectors/vector_1.py
+++ b/Plots/Vectors/vector_1.py
@@ -128,8 +128,8 @@ Q = plt.quiver(lon_uv, lat_uv, u, v, color='white',
                width=.0025, scale=(4.0/.045), zorder=2)
 
 # Draw legend for vector plot
-qk = ax.quiverkey(Q, 0.85, 0.9, 4, r'4 $m/s$', labelpos='N',
-                  coordinates='figure', color='black')
+qk = ax.quiverkey(Q, 94, 26, 4, r'4 $m/s$', labelpos='N', zorder=2,
+                  coordinates='data', color='black')
 
 # Draw SST contours
 plt.cm.register_cmap('BlAqGrYeOrReVi200', truncate_colormap(cmaps.BlAqGrYeOrReVi200, minval=0.08, maxval=0.96, n=len(levels)))
--- a/Plots/Vectors/vector_3.py
+++ b/Plots/Vectors/vector_3.py
@@ -22,7 +22,7 @@ import geocat.viz as gcv
 # This choice was made because ``geocat.viz`` doesn't offer an
 # equivalent function to ncl's ``vcMinDistanceF`` yet.
 file_in = xr.open_dataset('../../data/netcdf_files/uv300.nc')
-ds = file_in.isel(time=1, lon=slice(0,-1,3), lat=slice(2,-1,3))
+ds = file_in.isel(time=1, lon=slice(0,-1,3), lat=slice(1,-1,3))
 
 ###############################################################################
 # Make the plot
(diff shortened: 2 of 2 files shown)


clyne:
@mnlevy1981 Is this still a work in progress? It was approved by the reviewer. Thanks!

mnlevy1981:
Let me take one last look at where things stand -- I think this one will be okay to merge (after I add a note about the kludgy "only display every third vector in each direction while we figure out `vcMinDistanceF`")

clyne:
Sounds good. Getting some input from the user community would be helpful for figuring out a best path forward.

mnlevy1981:
Okay, I think this is ready for re-review and then merging. Note that it also uses `geocat.viz` as `gcv`, so it will need to be updated when #50 is addressed.

clyne:
@bonnland please re-review when you have time. Thanks!

bonnland:
@mnlevy1981   Only a couple of minor issues that I see, and I'm not sure how worth mentioning they are.   

First, the vectors along the bottom don't point in the same direction as the NCL version, which may be due to the vcMinDistanceF() function not being used, but still the vectors are large enough in the NCL version to maybe warrant another look, or a slightly more detailed explanation as to why the difference.     

Also, I wonder if the eventual equivalent function to vcMinDistanceF() will end up in geocat.comp, instead of geocat.viz, if it is some kind of number crunching function.    But perhaps not; I don't know if it even matters that much right now. 

Lastly, I tried manually resizing the figure, when running "python vector3.py",  and everything changed properly except the quiver key.   I'm not sure if this is a bug in quiverkey(), or if it is the wrong axis transform being used to describe its location.  

Anyway, that's all I have, and I think it looks awesome outside of these little things.

mnlevy1981:
#bonnland -- thanks for the comments!

> the vectors along the bottom don't point in the same direction as the NCL version, which may be due to the vcMinDistanceF() function not being used

I'll change the subset of `lat` that I use to get the bottom row of vectors to look more like the NCL field

> I wonder if the eventual equivalent function to vcMinDistanceF() will end up in geocat.comp, instead of geocat.viz, if it is some kind of number crunching function

That could be - I was picturing it as a wrapper to `plt.quiver`, which seems like it would go in `gcv`, but I could also see implementing it in a way where `geocat.comp` makes more sense

> Lastly, I tried manually resizing the figure, when running "python vector3.py", and everything changed properly except the quiver key.

Oh, I see what you mean -- it looks like `quiverkey()` is tied to the figure frame (`[0, 1] x [0, 1]`) instead of the axes being plotted. I'll see if I can find a work-around for that... I did notice that changing `figsize` in `plt.subplots()` required coming up with new coordinates for the key which is not ideal

mnlevy1981:
@bonnland This is ready for review again -- I was able to set `coordinates='data'` for `quiverkey()` so the location is tied to the grid (which is much smarter than how I was doing it before), and I modified the `lat` bands we plot to include a slightly lower band that looks closer to the NCL bottom band.

mnlevy1981:
Note that this PR will also include a change to `vector_1`, which suffered from the same resizing issue for the `quiverkey()`.

PR #40, opened 2020-01-24
Use geocat-datafiles for downloading data files

khallock:
Added geocat-datafiles as conda dependency and modified NCL_coneff_16.py to use geocat.datafiles.get to fetch data files instead of using the `/data` submodule directory of this repository.

### DIFF
--- a/.rtd_conda_environment.yml
+++ b/.rtd_conda_environment.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3
   - geocat-comp
+  - geocat-datafiles
   - netcdf4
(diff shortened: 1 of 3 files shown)


khallock:
Oops, I should've added a note about the new dependency.

You may need to run
```conda env update -f conda_environment.yml```
from the root dir of the repository with this branch checked out in order to get `geocat-datafiles` installed.

Alternatively, you can run
```conda install -c conda-forge -c ncar geocat-datafiles```

to install `geocat-datafiles` if you only downloaded the script and didn't checkout this branch.

Issue #13, opened 2019-12-20
mask_1


clyne:
Description: Masking off data values

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/mask_1.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/mask.shtml

PR #42, opened 2020-01-24
Add Masking plots

andersy005:
Towards #13

- NCL Version:

- Python Version:

sphx_glr_mask_1_002

- NCL Version:

- Python Version:
sphx_glr_mask_1_001

### DIFF
--- a/Plots/Masking/mask_1.py
+++ b/Plots/Masking/mask_1.py
@@ -4,12 +4,9 @@ mask_1.py
 Demonstrates the use of the xarray's where function and 
 a masking array to mask out land or ocean.
 
-https://www.ncl.ucar.edu/Applications/Scripts/mask_1.ncl
-
-https://www.ncl.ucar.edu/Applications/Images/mask_1_1_lg.png
-
-
-https://www.ncl.ucar.edu/Applications/Images/mask_1_2_lg.png
+- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/mask_1.ncl
+- https://www.ncl.ucar.edu/Applications/Images/mask_1_1_lg.png
+- https://www.ncl.ucar.edu/Applications/Images/mask_1_2_lg.png
 """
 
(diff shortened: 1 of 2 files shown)


Plots/Masking/mask_1.py
```diff
@@ -0,0 +1,187 @@
...
+import cartopy.crs as ccrs
+import cmaps
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+import numpy as np
+import xarray as xr
+
+# Read in the netCDF file
+
+ds = (
```
clyne: Could you please comment why the decode_times keyword is needed?

Plots/Masking/mask_1.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+# Use xarray's where function to mask out land and then ocean data
+land_only = ds.TS.where(ds.ORO == 1.0)
+ocean_only = ds.TS.where(ds.ORO == 0.0)
+
+
+# Define a few utility functions
+
+def truncate_colormap(cmap, minval=0.0, maxval=1.0, n=100):
```
clyne: Could you please explain  a little more about what truncate_colormap is doing (and maybe why it is needed)

Plots/Masking/mask_1.py
```diff
@@ -0,0 +1,187 @@
+"""
```
clyne: The plots have a gap along 0 degrees longitude (wraparound). Can this be fixed? I think @jukent had a solution.

andersy005:
New plots:

sphx_glr_mask_1_001 (1)
sphx_glr_mask_1_002 (1)

Plots/Masking/mask_1.py
andersy005: This is fixed

Plots/Masking/mask_1.py
andersy005: Added a note in the documentation string of the function

PR #43, opened 2020-01-24
bar_7

bonnland:


Plots/Bar/NCL_bar_7.py
```diff
@@ -0,0 +1,125 @@
...
+###############################################################################
+# Create the second bar chart with a legend.
+
+
+## NOTE: you may need to close the first figure window to see the second figure.
+# Figure size is (width, height) inches.
+plt.figure(2, figsize=(6, 6))
+
+bar_handle = plt.bar(x, data, color=color_list, edgecolor='black')
+plt.ylabel("Number of Deaths", fontsize = 16)
```
clyne: For some reason Y axis label doesn't appear to be showing up. Do you think you could figure out why?

bonnland: @clyne  The script creates two plots, where the first does not have a Y axis label, and the second one does (as in the NCL examples).   Can you confirm that the Y axis label is not showing up on the second plot?

clyne: I just confirmed that the Y axis label does not show up. @khallock, @erogluorhan can you please see if you can reproduce this problem? The NCL plot contains a Y-axis label and should look like this:

https://www.ncl.ucar.edu/Applications/Images/bar_7_2_lg.png

erogluorhan: Yes, I also confirm that Y axis label does not show up on Geocat-Examples webpage. I think that this is due to the aspect ratio management of the Geocat-examples webpage rather than the source codes because you can also see in the below screenshot that the Y-axis of my `NCL_xy_2` plot shows up as half-cut:

image

@khallock Could you please have a look at this?

bonnland:
@khallock @erogluorhan @clyne  If it is inconvenient to modify the code producing images in Sphinx, it is also possible to "shrink" the figure using subplot_adjust(), in order to prevent clipping. 

For now, I will adjust the left border of the plot this way:

```python
# Move the figure left border, so Y Label appears without manually adjusting the viewport.
plt.subplots_adjust(left=0.2)
```

bonnland:
> I haven't see this with other examples. Any idea what is going on here?

@clyne  I am seeing it with other examples, like the Contour example coneff_16.   From discussions like [this one](https://github.com/tctianchi/pyvenn/issues/3), the error makes sense because Sphinx is not intended to display output on a console.

@khallock Any idea what the sphinx-gallery backend is set to?   See https://github.com/sphinx-gallery/sphinx-gallery/issues/153 for some context.

bonnland:
> I haven't see this with other examples. Any idea what is going on here?

@clyne  Could you please point me to an example where this message is not showing?   Then I can try to figure out how this was prevented.

khallock:
@bonnland I'm looking into a global setting (in `conf.py`) to suppresses the Matplotlib "agg" warning message we're getting with this example. Ultimately, I'd prefer to avoid putting any code directly in these example scripts to ignore warnings, I think that might distract a user who's trying to make sense of the script.

bonnland:
> Ultimately, I'd prefer to avoid putting any code directly in these example scripts to ignore warnings, I think that might distract a user who's trying to make sense of the script.

@khallock  Thanks for the heads up.   I will remove the warning suppression code for now.
@clyne  I just pull the latest version of the coneff_16 example, which does not contain the plt.show() warning message.   My guess is that this is gone because there are no Jupyter notebook cells in this example any more.

bonnland:
@clyne If you pull Kevin's latest commit, delete the _build directory and rebuild with "make html", the plt.show() warning message should be gone on all notebooks.

clyne:
> @clyne If you pull Kevin's latest commit, delete the _build directory and rebuild with "make html", the plt.show() warning message should be gone on all notebooks.

Yes indeed it all looks great now!

PR #45, opened 2020-01-24
Overlay 11

### LINKED ISSUE #4: overlay_11
Description: Combines both vectors and contours

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/overlay_11.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/overlay_11_lg.png

kmpaul:
The original `overlay_11.ncl` script produces the following output:

overlay_11_lg

Closes #4


kmpaul:
Noticed that there are 2 ways of doing this with matplotlib:

a. Lay "filled" geometric objects over top of a plot (i..e, with `zorder`)
b. Clip a plot with an "empty" geometric boundary

I've decided to add both examples.  The first will be called `NCL_overlay_11a.py` and the second will with `NCL_overlay_11b.py`.

kmpaul:
The plot from `NCL_overlay_11a.py` is shown below:

Screen Shot 2020-01-28 at 3 37 53 PM

**NOTE:** The color bar is obviously different, but I think the matplotlib colorbar is better.  Also, the quiver key cannot be replicated with matplotlib, so I moved it to the top of the plot (above the plot region, next to the "Wind" label).

kmpaul:
Also, it seems that the arrowheads in the quiver diagram do not scale like the NCL plot.

mnlevy1981:
I added the quiverkey in an NCL-esque fashion in #34. You could do something like

```
ax.add_patch(plt.Rectangle((137.5, 52,5), 2.5, 2.5, facecolor='magenta', edgecolor=None, zorder=1))
```

To draw the box in the upper-right corner and then you just need to get the `quiverkey` coordinates correct.

kmpaul:
@mnlevy1981 I thought about doing something like that and thought I was "trying too hard."  😄  But since you have the code, I could easily add that.

kmpaul:
@mnlevy1981 I just added your recommendation.  Thanks!

kmpaul:
The plot from `NCL_overlay_11a.py` *now* looks like this:

Screen Shot 2020-01-28 at 5 08 11 PM

and the plot from `NCL_overlay_11b.py` looks like this:

Screen Shot 2020-01-28 at 5 07 29 PM

kmpaul:
I think this is now ready for review.

Plots/Overlays/NCL_overlay_11a.py
```diff
@@ -0,0 +1,213 @@
...
+# Create a subselection of the color map
+# --------------------------------------
+#
+# We create a new color map that is a subselection of an existing color map
+
+trunc_cm = LinearSegmentedColormap.from_list(
+    name="BkBlAqGrYeOrReViWh200",
+    colors=cmaps.BkBlAqGrYeOrReViWh200(np.linspace(0.1, 0.6, len(clevs))),
+)
+plt.cm.register_cmap('BkBlAqGrYeOrReViWh200', trunc_cm)
```
mnlevy1981: This can be done with `geocat.viz`'s `truncate_colormap()` function:

```python
import geocat.viz as gcv

cmap = gcv.util.truncate_colormap(cmaps.BkBlAqGrYeOrReViWh200, 0.1, 0.6, len(clevs), "BkBlAqGrYeOrReViWh200"
```

and I think it's a good idea to advertise `gcv` as much as possible

Plots/Overlays/NCL_overlay_11a.py
```diff
@@ -0,0 +1,213 @@
...
+# Create the figure and the Cartopy GeoAxes object
+fig = plt.figure(figsize=(12,12))
+ax = plt.axes(projection=crs)
+
+# Define the axis tick parameters and labels
+lon_formatter = LongitudeFormatter(zero_direction_label=False,
+                                   dateline_direction_label=False)
+lat_formatter = LatitudeFormatter()
+ax.xaxis.set_major_formatter(lon_formatter)
+ax.yaxis.set_major_formatter(lat_formatter)
```
mnlevy1981:
```python

gcv.util.add_lat_lon_ticklabels(ax)
```

(defaults to setting both `direction_label` variables to `False`)

Plots/Overlays/NCL_overlay_11a.py
```diff
@@ -0,0 +1,213 @@
...
+ax.minorticks_on()
+ax.xaxis.set_minor_locator(AutoMinorLocator(n=4))
+ax.yaxis.set_minor_locator(AutoMinorLocator(n=5))
+ax.tick_params("both", length=10, width=1.0, which="major", bottom=True, left=True, labelsize=20)
+ax.tick_params("both", length=7, width=0.5, which="minor", bottom=True, left=True, labelsize=20)
```
mnlevy1981:
```python

gcv.util.nclize_axis(ax, minor_per_major=4)
```

Note: you'll still need to update `ax.yaxis.set_minor_locator(AutoMinorLocator(n=5))`, as the function will set both `x` and `y` to use 4 minor ticks per major (default for the function is 3)

Plots/Overlays/NCL_overlay_11a.py
```diff
@@ -0,0 +1,213 @@
...
+ax.set_extent([100, 145, 15, 55], crs=crs)
+ax.set_xticks([100, 120, 140])
+ax.set_yticks([20, 30, 40, 50])
+
+# Draw the temperature contour plot with the subselected colormap
+# (Place the zorder of the contour plot at the lowest level)
+cf = ax.contourf(lon, lat, T, extend='both', levels=clevs, cmap=cmap, zorder=1)
```
mnlevy1981: Do you need `extend='both'` here?

Plots/Overlays/NCL_overlay_11a.py
```diff
@@ -0,0 +1,213 @@
...
+
+# Draw the color bar for the contour plot
+cax = plt.axes((0.14, 0.08, 0.74, 0.02))
+fig.colorbar(cf, ax=ax, cax=cax, ticks=clevs, drawedges=True, orientation='horizontal')
```
mnlevy1981: To stay truer to the NCL plot, I'd go with `ticks=range(232,270,4)` or `ticks=clevs[1:-2]` or something

Plots/Overlays/NCL_overlay_11a.py
mnlevy1981: This one may be worth leaving as-is, since `nclize_axis()` will result in slightly different `tick_params` than you are using and those values can not be overridden with optional arguments

kmpaul: Yeah.  That's why I didn't use the `nclize_axis()` function.  I'll leave it as-is.

Plots/Overlays/NCL_overlay_11a.py
kmpaul: Maybe not.  With the clipping, I don't think you can see the regions that extend beyond the colorbar range.  I'll try removing it and see if there is a noticeable difference.

Plots/Overlays/NCL_overlay_11a.py
kmpaul: Ok.  I'll try it.

Plots/Overlays/NCL_overlay_11a.py
kmpaul: This is excellent!  I'll give that a shot.

kmpaul: Done.

Plots/Overlays/NCL_overlay_11a.py
kmpaul: Done.

Plots/Overlays/NCL_overlay_11a.py
kmpaul: Looks like you do not need `extend='both'` due to the clipping.  Nice!

(Maybe you didn't need it at all.  Can't remember, actually.)

Plots/Overlays/NCL_overlay_11a.py
kmpaul: Yep.  That worked.  Good catch!

kmpaul:
Incidentally, the new plots from version (a) and (b) are identical and look like this:

NCL_overlay_11a

Issue #9, opened 2019-12-20
polyg_4


clyne:
Description: Plots with polygons

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/polyg_4.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/polyg.shtml

jukent:
Not sure that the polyg plot corresponds to the plyg_4 script.

khallock:
Closed by #48

PR #48, opened 2020-01-24
Add polyg_4 example

bonnland:
This is a work in progress.   Hatch patterns are hard to duplicate in matplotlib.

### DIFF
--- a/Plots/Polygons/NCL_polyg_4.py
+++ b/Plots/Polygons/NCL_polyg_4.py
@@ -1,0 +1,172 @@
+"""
+NCL_polyg_4.py
+==============
+ Concepts illustrated:
+   - Drawing a cylindrical equidistant map
+   - Zooming in on a particular area on a cylindrical equidistant map
+   - Attaching an outlined box to a map plot
+   - Attaching filled polygons to a map plot
+   - Filling in polygons with a shaded pattern
+   - Changing the color and thickness of polylines
+   - Changing the color of a filled polygon
+   - Labeling the lines in a polyline
+   - Changing the density of a fill pattern
+   - Adding text to a plot
+
+This Python script reproduces the NCL plot script found here: https://www.ncl.ucar.edu/Applications/Scripts/polyg_4.ncl
+"""
+
+###############################################################################
+# Import the necessary python libraries
+import numpy as np
+import xarray as xr
+import cartopy
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+
+
+ds = xr.open_dataset("../../data/netcdf_files/uv300.nc").isel(time=1)
+
+###############################################################################
+# These next two functions add nice axes decorations and make the plot look more
+# like NCL
+
+
+def add_lat_lon_ticklabels(ax):
+    """
+    Nice latitude, longitude tick labels
+    """
+    from cartopy.mpl.ticker import LongitudeFormatter, LatitudeFormatter
+
+    lon_formatter = LongitudeFormatter(
+        zero_direction_label=False, dateline_direction_label=False
+    )
+    lat_formatter = LatitudeFormatter()
+    ax.xaxis.set_major_formatter(lon_formatter)
+    ax.yaxis.set_major_formatter(lat_formatter)
+
+
+def nclize_axis(ax):
+    """
+    Utility function to make plots look like NCL plots
+    """
+    import matplotlib.ticker as tic
+
+    ax.tick_params(labelsize="small")
+    ax.minorticks_on()
+    ax.xaxis.set_minor_locator(tic.AutoMinorLocator(n=3))
+    ax.yaxis.set_minor_locator(tic.AutoMinorLocator(n=3))
+
+    # length and width are in points and may need to change depending on figure size etc.
+    ax.tick_params(
+        "both",
+        length=8,
+        width=1.5,
+        which="major",
+        bottom=True,
+        top=True,
+        left=True,
+        right=True,
+    )
+    ax.tick_params(
+        "both",
+        length=5,
+        width=0.75,
+        which="minor",
+        bottom=True,
+        top=True,
+        left=True,
+        right=True,
+    )
+
+
+
+
+###############################################################################
+# Another text/markdown cell
+
+def xr_add_cyclic(da, coord):
+    from cartopy.util import add_cyclic_point
+
+    cyclic_data, cyclic_coord = cartopy.util.add_cyclic_point(da.values, coord=da[coord])
+
+    coords = da.coords.to_dataset()
+    coords[coord] = cyclic_coord
+    return xr.DataArray(cyclic_data, dims=da.dims, coords=coords.coords)
+
+
+
+###############################################################################
+# Note the inline comment in the previous code cell. This was not converted to
+# its own markdown cell because it was not contiguously preceeded with a
+# comment line beginning with 20+ '#' characters.
+
+# first add continents
+continents = cartopy.feature.NaturalEarthFeature(
+    name="coastline",
+    category="physical",
+    scale="50m",
+    edgecolor="None",
+    facecolor="lightgray",
+)
+
+#$map_extent = [-130, 0, -8, 25]
+map_extent = [-130, 0, -20, 40]
+ax = plt.axes(projection=ccrs.PlateCarree())
+ax.add_feature(continents)
+ax.set_extent(map_extent, crs=ccrs.PlateCarree())
+
+
+levels = np.arange(-12, 40, 4)
+# Using a dictionary makes it easy to reuse the same keyword arguments twice for the contours
+kwargs = dict(
+    levels=levels,  # contour levels specified outside this function
+    xticks=[-120, -90, -60, -30, 0],  # nice x ticks
+    yticks=[-20, 0, 20, 40],  # nice y ticks
+    transform=ccrs.PlateCarree(),  # ds projection
+    add_colorbar=False,  # don't add individual colorbars for each plot call
+    add_labels=False,  # turn off xarray's automatic Lat, lon labels
+    colors="k",  # note plurals in this and following kwargs
+    linestyles="-",
+    linewidths=0.5,
+)
+
+# Create contour plot
+hdl = ds.U.plot.contour(
+    x="lon",  # not strictly necessary but good to be explicit
+    y="lat",
+    ax=ax,  # this is the axes object we want to plot to
+    **kwargs,
+)
+
+# Add contour labels
+ax.clabel(hdl,
+          np.arange(-8, 32, 8),  # only label these contour levels
+          fontsize="small",
+          fmt="%.0f")  # Turn off decimal points
+
+# Create square polygon
+x_points = [-90.0, -45.0,-45.0, -90.0,-90.0]
+y_points = [30.0,  30.0,  0.0,   0.0, 30.0]
+
+# Plot the hatch pattern first
+ax.fill(x_points, y_points,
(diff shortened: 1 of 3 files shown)


bonnland:
The basic map and square polygon are present in the example.

bonnland:
@khallock  This example should work, but it probably needs to be refactored to make calls to geocat-viz.   I don't know if geocat-viz is ready yet.   I hope the refactoring will be quick and easy, given that I've copy-pasted from one of the vector examples.

khallock:
@bonnland `geocat-viz` is available with the `nclize_axis` and `add_lat_lon_ticklabels` functions you're using, although it's very much a WIP still. If you want to try using `geocat-viz`, you can install it by running the following command with your conda environment already activated:
```
conda install -c conda-forge -c ncar geocat-viz
```

Once `geocat-viz` is installed, you can `import geocat.viz as gcv` (we're going to use a different shorthand name for `geocat.viz` eventually, but `gcv` is fine for now). Then the two functions should be available under the `gcv.util` namespace as `gcv.util.nclize_axis` and `gcv.util.add_lat_lon_ticklabels`.

bonnland:
polyg_4

This is the current plot.

bonnland:
@khallock   Your suggestions worked like a charm.   Do you want me to push changes that make calls to geocat-viz?   I ask because it means that reviewers will have to follow the same steps as I did to install geocat-viz, but it worked without a hitch for me.

khallock:
@bonnland Great, I'm glad that worked! Yes, please push your changes that use `geocat-viz`, it'll be one less thing to have to update later. I can help anyone fix their environment if needed.

khallock:
@bonnland The plot you attached above looks great! However, when I generate the actual gallery pages using `make html`, my output plot doesn't seem to be complete:
sphx_glr_NCL_polyg_4_001

I think @erogluorhan has been working through a similar issue with one of his example scripts, I'll check with him to see if he's found a solution yet.

clyne:
I have the same issue as @khallock: only a single plot is generated and it's incomplete. However, if I run the script directly I get a plot that matches 

https://www.ncl.ucar.edu/Applications/Images/polyg_4_2_lg.png

almost perfectly. The only difference is that it is missing the "CONTOUR FROM..." label in the bottom right. Moreover, once again only a single plot is generated.

erogluorhan:
All, 

1) I had this same issue with my functions `NCL_xy_2.py` and `NCL_conwomap_2.py` and resolved it later. 

2) How I resolved it: The issue was happening when I had my `xArray.DataArray.plot() or contour()` lines first, then all other plot-related lines (e.g. tick labeling/limiting, title, `plot.show()` etc.) afterwards. I moved my `xArray.DataArray.plot() or contour()` lines as close as possible to `plt.show()` line all after other plot-related stuff.

3) My understanding of the issue: I cannot make a comprehensive ruling to fully prevent this issue, but my understanding is that PyCharm-like IDEs generate the plot when they interpret `plt.show()` line. This is why we do not get any artifact when we are running code in our IDEs. On the other hand, Sphynx-generated Jupyter notebooks might be creating separate plots as soon as they interpret various lines of codes in changing orders. To come up with a complete ruling, the issue requires a bit deeper search.

mnlevy1981:
An alternative to @erogluorhan's suggestion is to create a function `make_base_plot()` (or similar) that returns `ax`. Then start each new cell with `ax = make_base_plot()` and the `plt.show()` command will work as expected.

mnlevy1981:
Specifically, I did this:

```python
def make_base_plot():
    # first add continents
    continents = cartopy.feature.NaturalEarthFeature(
        name="coastline",
        category="physical",
        scale="50m",
        edgecolor="None",
        facecolor="lightgray",
    )

    map_extent = [-130, 0, -20, 40]
    ax = plt.axes(projection=ccrs.PlateCarree())
    ax.add_feature(continents)
    ax.set_extent(map_extent, crs=ccrs.PlateCarree())

    # Specify contour levels (from -12 to 40 by 4).   The top range value of 44 is not included in the levels.
    levels = np.arange(-12, 44, 4)

    # Using a dictionary prevents repeating the same keyword arguments twice for the contours.
    kwargs = dict(
        levels=levels,  # contour levels specified outside this function
        xticks=[-120, -90, -60, -30, 0],  # nice x ticks
        yticks=[-20, 0, 20, 40],  # nice y ticks
        transform=ccrs.PlateCarree(),  # ds projection
        add_colorbar=False,  # don't add individual colorbars for each plot call
        add_labels=False,  # turn off xarray's automatic Lat, lon labels
        colors="gray",  # note plurals in this and following kwargs
        linestyles="-",
        linewidths=0.5,
    )

    # Create contour plot
    hdl = ds.U.plot.contour(
        x="lon",  # not strictly necessary but good to be explicit
        y="lat",
        ax=ax,    # this is the axes object we want to plot to
        **kwargs,
    )

    # Add contour labels.   Default contour labels are sparsely placed, so we specify label locations manually.
    # Label locations only need to be approximate; the nearest contour will be selected.
    label_locations = [(-123, 35), (-116, 17), (-94, 4), (-85, -6), (-95, -10),
                       (-85, -15), (-70, 35), (-42, 28), (-54, 7), (-53, -5),
                       (-39, -11), (-28, 11), (-16, -1), (-8, -9),             # Python allows trailing list separators.
                       ]
    ax.clabel(hdl,
              np.arange(-8, 24, 8),    # Only label these contour levels: [-8, 0, 8, 16]
              fontsize="small",
              colors="black",
              fmt="%.0f",              # Turn off decimal points
              manual=label_locations,  # Manual label locations
              inline=False)            # Don't remove the contour line where labels are located.

    return ax

# Draw the basic contour plot
ax = make_base_plot()
plt.show()
```

The only differences in the first few cells (suppressing the whitespace inherent in moving your main cell into a function):

```diff
$ git diff -w
diff --git a/Plots/Polygons/NCL_polyg_4.py b/Plots/Polygons/NCL_polyg_4.py
index d7758ec..36f26e7 100644
--- a/Plots/Polygons/NCL_polyg_4.py
+++ b/Plots/Polygons/NCL_polyg_4.py
@@ -28,8 +28,9 @@ import geocat.viz as gcv
 ds = xr.open_dataset("../../data/netcdf_files/uv300.nc").isel(time=1)

 ###############################################################################
-# Draw the basic contour plot
+# Define a function to create the basic contour plot

+def make_base_plot():
     # first add continents
     continents = cartopy.feature.NaturalEarthFeature(
         name="coastline",
@@ -82,9 +83,17 @@ ax.clabel(hdl,
               manual=label_locations,  # Manual label locations
               inline=False)            # Don't remove the contour line where labels are located.

+    return ax
+
+###############################################################################
+# Draw the basic contour plot
+ax = make_base_plot()
+plt.show()
+
 ###############################################################################
 # Create a rectangular polygon with hatch pattern.

+ax = make_base_plot()
 x_points = [-90.0, -45.0, -45.0, -90.0, -90.0]
 y_points = [ 30.0,  30.0,   0.0,   0.0,  30.0]

@@ -112,6 +121,8 @@ p = plt.Rectangle((left, bottom), width, height, fill=False,
                   alpha=0.5)         # Lower color intensity.
 ax.add_patch(p)

+plt.show()
```

bonnland:
> 3\. My understanding of the issue: I cannot make a comprehensive ruling to fully prevent this issue, but my understanding is that PyCharm-like IDEs generate the plot when they interpret `plt.show()` line. This is why we do not get any artifact when we are running code in our IDEs. On the other hand, Sphynx-generated Jupyter notebooks might be creating separate plots as soon as they interpret various lines of codes in changing orders. To come up with a complete ruling, the issue requires a bit deeper search.


So we might want to be clear about what we want for these examples.   I am a little uncomfortable with complicating the code structure for educational examples, when the problem is only going to occur for Sphinx.    Anyone trying to run the python code directly is going to see a complete plot.   It seems like trying to fix the document generation is the best way to go.

bonnland:
There is a possible bug with matplotlib savefig() that can clip the figure;  see here for a possible solution (a parameter must be added to the savefig() call):

https://stackoverflow.com/questions/44642082/text-or-legend-cut-from-matplotlib-figure-on-savefig

erogluorhan:
**All, I found more evidences about the main issue that leads to multiple plot generation at Sphinx level on Geocat-examples site for indeed a single plot:**

I noticed that separating the axes-related commands and plotting commands from each other **by creating new Jupyter cells (with 25 or more ####### characters)** is resulting in the generation of multiple plots (most of the time, one of the plots is empty, the other shows data) at Sphinx level. In other words, without modifying any source code and their orders, making all the plotting lines as a single Jupyter cell is working correctly for the cases I tested.

From this point, I believe that we should resolve this issue at Sphinx, if possible. We can further discuss with @khallock about it, and I can show a number of cases that lead to such issues.

@bonnland I agree about simplifying example-implementation process, and I believe our ongoing investigations will help us achieve this. 

@mnlevy1981 Thank you for your contribution; that makes a great option at coding level.

bonnland:
> The only difference is that it is missing the "CONTOUR FROM..." label in the bottom right. Moreover, once again only a single plot is generated.

Until now, I missed the fact that the NCL script generated two plots.   I will add the missing plot shortly.

bonnland:
> An alternative to @erogluorhan's suggestion is to create a function `make_base_plot()` (or similar) that returns `ax`. Then start each new cell with `ax = make_base_plot()` and the `plt.show()` command will work as expected.

@mnlevy1981 I see now how this helps when trying to create two plots.   Thanks for the suggestion.

bonnland:
@khallock @clyne @erogluorhan @mnlevy1981   Please review these changes.  You should see two plots produced by Sphinx.   However, I am still seeing only one thumbnail image on the main index.html page.    

If someone knows of a way to generate both thumbnail images, let me know.   Thank you!

khallock:
@bonnland Unfortunately, sphinx-gallery only allows a single thumbnail image to be used on the main gallery page.

However, it looks like it's possible to select which image (if multiple are generated) is used as the thumbnail:
https://sphinx-gallery.github.io/stable/configuration.html#choosing-thumbnail
https://sphinx-gallery.github.io/stable/auto_examples/plot_4_choose_thumbnail.html#sphx-glr-auto-examples-plot-4-choose-thumbnail-py

khallock:
This PR will close #9

Plots/Polygons/NCL_polyg_4.py
```diff
@@ -0,0 +1,230 @@
...
+    ax.text(1.0,  1.05, "m/s", fontsize=12, transform=ax.transAxes, horizontalalignment='right')
+    ax.text(1.0, -0.20, "CONTOUR FROM -12 TO 40 BY 4",
+            fontname='Helvetica',
+            horizontalalignment='right',
+            transform=ax.transAxes,
+            bbox=dict(boxstyle='square, pad=0.1', facecolor='white', edgecolor='black'))
+
+    gcv.util.nclize_axis(ax)
+    gcv.util.add_lat_lon_ticklabels(ax)
+
```
mnlevy1981: After these two lines, I would add

```python
ax.yaxis.set_minor_locator(AutoMinorLocator(n=4))
```

To get the right number of minor tick marks in the y axis (especially nice because your major ticks are 20 degrees apart, so this makes the minor ticks 5 degrees apart instead of 6 2/3 degrees). You'll also need

```
from matplotlib.ticker import AutoMinorLocator
```

in your `import` block

bonnland: @mnlevy1981  That's a good catch, thanks Mike!   I've pushed the change.

PR #49, opened 2020-01-25
Station 2

### LINKED ISSUE #21: station_2
Description: Station markers

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/station_2.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/station.shtml

mnlevy1981:
Randomly generates 100 lat, lon points and data at said points, then generates a scatter plot where the color of the scatter dot depends on the value at the point. Two plots, one with a legend and one with a color bar:

sphx_glr_station_2_001

sphx_glr_station_2_002

The dot colors don't exactly line up - it's probably possible to bring in `cmaps.ncl_default` and get the exact same colors, but I think showing the user a list named `colors` using string names is more informative.

Fixes #21

### DIFF
--- a/Plots/Station/README.rst
+++ b/Plots/Station/README.rst
@@ -1,0 +1,6 @@
+.. _station_examples:
+
+.. _station-examples-index:
+
+Station Examples
+================
--- a/Plots/Station/station_2.py
+++ b/Plots/Station/station_2.py
@@ -6,6 +6,8 @@ Plot random data on US, colored according to value
 
 https://www.ncl.ucar.edu/Applications/Scripts/station_2.ncl
 """
+###################################################
+# Import necessary packages
 import numpy as np
 import matplotlib as mpl
 from matplotlib import pyplot as plt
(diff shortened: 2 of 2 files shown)


jeff-willison:
@mnlevy1981: this script does not seem to work as expected. I believe norm (or vmin/vmax) need to be used in the scatter call or else the values don't correspond to the correct colors. Out of the box with hard coded values of 17.5:

Screenshot 2024-12-10 112542

Passing norm as an argument to the scatter call:

Screenshot 2024-12-10 112617

mnlevy1981:
> @mnlevy1981: this script does not seem to work as expected. I believe norm (or vmin/vmax) need to be used in the scatter call or else the values don't correspond to the correct colors. Out of the box with hard coded values of 17.5:
> 
> Screenshot 2024-12-10 112542
> 
> Passing norm as an argument to the scatter call:
> 
> Screenshot 2024-12-10 112617

Good catch! I've opened #629

PR #51, opened 2020-01-29
eof_1

bonnland:


khallock:
@bonnland Just a heads up that the NCL functions `eofunc_Wrap` and `eofunc_ts_Wrap` are available in `geocat.comp`, which should already be installed in your local conda environment.

Here are the documentation pages for these functions, please let me know if you have any questions:
https://geocat-comp.readthedocs.io/en/latest/user_api/generated/geocat.comp.eofunc.html
https://geocat-comp.readthedocs.io/en/latest/user_api/generated/geocat.comp.eofunc_ts.html

Note that `geocat.comp` does not use the `_Wrap` suffix on function names like NCL does, but instead uses an optional `meta=True` keyword argument:
```
geocat.comp.eofunc(..., meta=True) # good
geocat.comp.eofunc_Wrap(...)       # bad
```

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+###############################################################################
+# Flip longitude coordinates.
+print(f'Before flip, longitude range is [{ds["lon"].min().data}, {ds["lon"].max().data}].')
+
+ds["lon"] = ((ds["lon"] + 180) % 360) - 180
+
+# Sort longitudes, so that subset operations end up being simpler.
+ds["lon"] = np.sort(ds["lon"])
```
dcherian: This can't be right. I think you want `ds = ds.sortby("lon")`

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,67 @@
...
+
+print(f'After flip, longitude range is [{ds["lon"].min().data}, {ds["lon"].max().data}].')
+
+
+###############################################################################
+# Reverse latitude ordering.
+
+# Array indexing syntax is usually [start:end:stride], but here we leave off
+# start and end to indicate the full array.
+ds["lat"] = ds["lat"][::-1]
```
dcherian: similarly here, `ds = ds.sortby("lat", ascending=True/False)` or `ds = ds.reindex(lat=ds.lat[::=1])`

Plots/Contours/NCL_eof_1_1.py
bonnland: @dcherian Thanks for the advice.  But what I have seems to work; it produces the desired subset.   What is the reason to prefer one version over the other?   I ask as a naive user.  Thanks.

Plots/Contours/NCL_eof_1_1.py
bonnland: Similar comment to the above.

Plots/Contours/NCL_eof_1_1.py
dcherian: you're changing the order of the coordinate values (`lat`) without changing the order of the associated variables such as `slp`.

bonnland: I see; I haven't plotted the subset values yet, so I did not have a way of knowing this.   Thank you.

dcherian: :+1:

bonnland:
@khallock  This example computes seasonal means using month_to_season().   Is this function available?   If not, are months weighted by the number of days in the month?   I see an xarray example that does this, and I could borrow from it.

khallock:
@bonnland `month_to_season` is not yet available in `geocat.comp`, although we do have an issue for it on the GeoCAT-comp github repo: NCAR/geocat-comp#7

The NCL code for `month_to_season` can be found [here](https://github.com/NCAR/ncl/blob/develop/ni/src/examples/gsun/contributed.ncl#L8001). It looks like all months are weighted equally regardless of number of days.

If you do write an implementation of `month_to_season` using xarray, we'd be happy to consider accepting it as a pull request on geocat-comp.

Thanks for working on this!

bonnland:
@khallock  Thanks for the link; I realized shortly after posting my comment that I could have searched for it myself.   

I guess I would like to take a crack at an xarray implementation of month_to_season(), or at least see how much work it would be.   I wonder about whether the goal would be to preserve the exact behavior of the NCL version, or whether there is some room to either simplify or enhance the behavior.    An example would be whether to weight by the number of days in the month.  I guess this is a question with a subjective answer, and it's just simplest if the xarray version replicates behavior.

dcherian:
> If you do write an implementation of month_to_season using xarray, we'd be happy to consider accepting it as a pull request on geocat-comp.

AFAIK esmlab does this (cc @andersy005)

bonnland:
@khallock  I believe I am getting back something reasonable from geocat.comp.eofunc(), but I am getting an error when following up with a call to geocat.comp.eofunc_ts().   I may be missing something basic, but from my best reading of the geocat docs, the call I'm making seems correct. 

If you copy/paste the current NCL_eof_1_1.py example into a file and run it, you should see an error for the final line that says: 

`  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/pydevd.py", line 1434, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Applications/PyCharm CE.app/Contents/plugins/python-ce/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/bonnland/PycharmProjects/GeoCat-examples/Plots/Contours/NCL_eof_1_1.py", line 164, in <module>
    eof_ts = eofunc_ts(xw["slp"], eof, time_dim=1, meta=True)
  File "/Users/bonnland/miniconda3/envs/geocat-examples/lib/python3.7/site-packages/geocat/comp/__init__.py", line 780, in eofunc_ts
    response = _ncomp._eofunc_ts_n(np_data, np_evec, time_dim, options, missing_value=missing_value)
  File "src/geocat/comp/_ncomp.pyx", line 31, in geocat.comp._ncomp.carrayify.wrapper
  File "src/geocat/comp/_ncomp.pyx", line 589, in geocat.comp._ncomp._eofunc_ts_n
geocat.comp._ncomp.NcompError: An error occurred while calling libncomp.eofunc_ts_n with error code: 3
`

khallock:
@bonnland I was able to reproduce that error and trace it back to a bug in the NComp library (the C++ and Fortran code used by GeoCAT-comp). I created a pull request (NCAR/ncomp#23) with a fix and will try to get a new NComp conda package built as soon as I can. Thanks for the update and the bug report!

bonnland:
@khallock   Please review the plot created by this code.   It looks like the first plot is identical to the NCL version, but the other two plots have flipped signs.   

Because the EOF computation sounds similar to PCA, I'm wondering if the computed EOF directions come in (postive, negative) pairs, so that the sign is arbitrary (choosing the EOF direction or its negated version is arbitrary).    This is just a huge guess, having not looked at the computation myself.

khallock:
@bonnland It looks like the arrays being passed to `eofunc` in Python and in NCL are different and that might be affecting the plots. As a quick check, I printed some information about the first argument arrays being passed to `eofunc`.

If I add`print(xw["slp"])` at line 160 in `NCL_eof_1_1.py`, I get:
```
<xarray.DataArray 'slp' (lat: 23, time: 26, lon: 45)>
```
And if I add `printVarSummary(xw)` at line 98 of `eof_1.ncl`:
```
Variable: xw
Type: float
Total Size: 103500 bytes
            25875 values
Number of Dimensions: 3
Dimensions and sizes:	[lat | 23] x [lon | 45] x [time | 25]
```

There seems to be an extra timestep in the Python implementation, so the code that generates that array might be a good place to start poking around. I hope this helps!

bonnland:
Here is a diagnostic plot to verify correct values.    This plot, made by running NCL, shows the first three timesteps of the SLP variable.   
SLP

bonnland:
@khallock @dcherian   I fixed a bug in my month_to_season() implementation, and now the contour plots look basically correct.   The code can still use some simplifying and cleanup before publishing.

I am, however, getting some strange results when trying to add contour labels.  You can uncomment the "clabel" line to see.   It looks like some strange polygons get added to the plot, in addition to the labels.    I have not found any explanation online for this behavior.

bonnland:
@khallock  The bar plot is interesting.   If you run the NCL script, the bar plot values are scaled differently from this screenshot:   https://www.ncl.ucar.edu/Applications/Images/eof_1_2_lg.png

I think the plot generated in python has the same scaling as when the NCL script is run.

bonnland:
@erogluorhan  The code is basically ready for review.   There is some strange plotting issue when contour labels are added to the plot.   You will find that line commented out in the code.   It also appears that the second plot is basically correct, except that the plotted values are off by a factor of two or so.  

Please try running the script, or looking at the Sphinx output, to review.   Thank you!

bonnland:
@erogluorhan   I tried to take all your suggestions into account.   There are a few outstanding issues that I see:

-   If you look at the Sphinx output, you will see some warnings about constrained_layout not applied.   I have no idea what is going on; if you ever figure out how to address this, I would like to know how you did it.  
-  The diagnostic print statements might be considered by some people to be excessive.   Personally I feel it provides some understanding about data slicing issues.
- The EOF bar plot does not look like this screenshot:  https://www.ncl.ucar.edu/Applications/Images/eof_1_2_lg.png.  I believe this is because the screenshot is outdated.   If you run the NCL script as provided, it produces bar heights that are half as tall as the screenshot.   

I can make another pass if you like, but I am also happy with you taking the code and making it consistent with the other examples.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,290 @@
...
+    ax.coastlines(linewidth=0.5)
+    gvutil.add_lat_lon_ticklabels(ax)
+    return cplot, ax
+
+
+###############################################################################
+# Draw a contour plot for each EOF.
+
+fig, axs = plt.subplots(neof, 1,
+                        constrained_layout=True,  # "magic"
```
dcherian:
```suggestion
```

This should fix it.

bonnland: Hi Deepak, when I remove 'constrained_layout=True',  it shrinks the plot axes and gives too much of the figure space to the colorbar.    It seems wrong, like matplotlib could not possibly have these configuration settings as defaults.  Have you encountered this before?   I have looked at the other examples in the NCL gallery with multiple map subplots, and they all seem to have 'constrained_layout=True'.

dcherian: You could try `plt.tight_layout()` instead

Otherwise I would tweak figure size till constrained_layout does OK. `constrained_layout` uses a linear constraint solver; sometimes you don't find a solution and it issues that warning. https://matplotlib.org/tutorials/intermediate/constrainedlayout_guide.html

dcherian: https://matplotlib.org/tutorials/intermediate/constrainedlayout_guide.html#debugging

bonnland: I have tried debugging, and the warning about constrained_layout will not go away even if the plot consists of three simple boxes:
test

It looks to be an outstanding bug when using constrained_layout and cartopy together.

@erogluorhan   I have run out of time working on this example.   Please let me know if you are able to take over changes to this pull request.  Thank you!

bonnland: BTW I tried using both matplotlib 3.1.3 and 3.2.1.   Both versions had this problem with the warning message.   However, the plot (after a call to plt.savefig()) looks fine despite the warning message.

erogluorhan: @bonnland Thanks a lot for this great effort! Yes, let me take over this code and see if I can put anything beyond. I will get into every single comment of yours and let you know about my findings, if any.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,320 @@
...
+cbar = plt.colorbar(cplot, ax=axs, orientation='horizontal', shrink=0.6)
+cbar.ax.tick_params(labelsize=8)
+
+# Set a common title
+axs[0].set_title(f'SLP: DJF: {yearStart}-{yearEnd}', fontsize=14, y=1.12)
+
+# plt.savefig('test.png')
+plt.savefig('eof_1_1_lg.png')
+
+# Show the plot
```
bonnland, line 269: Remove?

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,320 @@
...
+
+    axs[i] = make_bar_plot(axs[i], eof_single)
+    pct = eof.pcvar[i]
+    gvutil.set_titles_and_labels(axs[i], lefttitle=f'EOF {i + 1}', lefttitlefontsize=10,
+                                 righttitle=f'{pct:.1f}%', righttitlefontsize=10)
+
+# Set a common title
+axs[0].set_title(f'SLP: DJF: {yearStart}-{yearEnd}', fontsize=14, y=1.12)
+
+# plt.savefig('test.png')
```
bonnland, line 315: ibid

Plots/Contours/NCL_eof_1_1.py
erogluorhan: Yes, thanks a lot for catching this! Removed it with a new commit.

Plots/Contours/NCL_eof_1_1.py
erogluorhan: Yes, thanks a lot for catching this! Removed it with a new commit.

erogluorhan:
Hi @khallock  and @dcherian, I have pushed a new commit [Constraint_layout warnings resolved](https://github.com/NCAR/GeoCAT-examples/pull/51/commits/9a4a965ee0f5c9cfc439c995434e5c130139d882) that resolves the **constraint_layout** warning that is generated within the notebook and attaches **diagnostic print statements** to a simple debug parameter (in order to avoid prints in examples website).

With this version, it should be good for you reviews.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+deg2rad = 4. * atan(1.) / 180.
+clat = SLP['lat'].astype(dtype=np.float64)
+clat = np.sqrt(np.cos(deg2rad * clat))
+print_debug('\n\nclat:\n')
+print_debug(clat)
+
+###############################################################################
+# Multiply SLP by weights:
+
+# Xarray uses the supplied coordinate information to apply latitude-based
```
dcherian, line 168:
```suggestion
# Xarray will apply latitude-based
```

"This is called broadcasting"

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+eof_ts = eofunc_ts(xw["slp"], eof, time_dim=1, meta=True)
+
+print_debug('\n\neof_ts:\n\n')
+print_debug(eof_ts)
+
+###############################################################################
+# Normalize time series:
+
+# Sum spatial weights over the area used.
+nLon = len(xw['lon'])
```
dcherian, line 203:
```suggestion
nLon = xw.sizes["lon"]
```

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+latS = 25.
+latN = 80.
+lonL = -70.
+lonR = 40.
+
+yearStart = 1979
+yearEnd = 2003
+
+neof = 3   # number of EOFs
+optETS = False
```
dcherian, line 56: what is optETS?

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+print_debug(eof_ts)
+
+###############################################################################
+# Normalize time series:
+
+# Sum spatial weights over the area used.
+nLon = len(xw['lon'])
+clat_subset = clat.where((clat.lat >= latS) & (clat.lat <= latN), drop=True)
+weightTotal = clat_subset.sum() * nLon
+eof_ts = eof_ts / weightTotal
```
dcherian, line 206:
```suggestion
eof_ts /= weightTotal
```

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+
+print_debug('\n\neof_ts:\n\n')
+print_debug(eof_ts)
+
+###############################################################################
+# Normalize time series:
+
+# Sum spatial weights over the area used.
+nLon = len(xw['lon'])
+clat_subset = clat.where((clat.lat >= latS) & (clat.lat <= latN), drop=True)
```
dcherian, line 204: Why not `lat.sel(lat=slice(latS, latN+0.01))`? it's a lot more efficient than where.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+
+    # Import an NCL colormap
+    cmap = gvcmaps.BlWhRe
+
+    # Specify contour levelstamam
+    v = np.linspace(-0.08, 0.08, 9, endpoint=True)
+
+    # The function contourf() produces fill colors, and contour() calculates contour label locations.
+    cplot = ax.contourf(lon, lat, values, levels=v, cmap=cmap, extend="both", transform=ccrs.PlateCarree())
+    p = ax.contour(lon, lat, values, levels=v, linewidths=0.0, transform=ccrs.PlateCarree())
```
dcherian, line 229: `0.0` linewidth seems funny; can we do `0.1`?

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+    except KeyError:
+        raise ValueError("contributed: month_to_season: bad season: SEASON = " + season)
+
+    # Compute the three-month means, moving time labels ahead to the middle month.
+    month_offset = 'MS'
+    xSeasons = xMon.resample(time=season_pd, loffset=month_offset).mean()
+
+    # Filter just the desired season, and trim to the desired time range.
+    xSea = xSeasons.sel(time=xSeasons.time.dt.month.isin(season_sel))
+    xSea = xSea.sel(time=slice(startDate, endDate))
```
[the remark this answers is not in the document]
bonnland: This decision was based on trying to duplicate output from the original NCL month_to_season() function.   If you provide 36 months of data, say from Jan-2000 to Dec-2002, then month_to_season() with season of 'DJF' should return three seasonal values, corresponding to the time points:  Jan-2000, Jan-2001, Jan-2002.    Without the trimming step, there would be a fourth point for Jan-2003.

Plots/Contours/NCL_eof_1_1.py
bonnland: I question whether this is simpler, given that the user needs to know about "sizes".   It seems likely that the user already knows how to calculate the length of an array.

bonnland: I guess the argument for using "sizes" is that length was calculated already, so the re-calculation is wasteful.   Probably worth doing for big arrays.

Plots/Contours/NCL_eof_1_1.py
bonnland: It should be removed; I think it may have been used at some point in the NCL script, but it is no longer being used.

Plots/Contours/NCL_eof_1_1.py
bonnland: This may decrease readability for people transitioning from NCL, not sure.

Plots/Contours/NCL_eof_1_1.py
bonnland: Agreed.  These differences in performance are not obvious without more experience, which I am in the process of getting.

Plots/Contours/NCL_eof_1_1.py
bonnland: The original plot has no contour lines.   We need to call the function to calculate contour label locations only.   This is the only way I know to get them.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,326 @@
...
+        (season_pd, season_sel) = seasons_pd[season]
+    except KeyError:
+        raise ValueError("contributed: month_to_season: bad season: SEASON = " + season)
+
+    # Compute the three-month means, moving time labels ahead to the middle month.
+    month_offset = 'MS'
+    xSeasons = xMon.resample(time=season_pd, loffset=month_offset).mean()
+
+    # Filter just the desired season, and trim to the desired time range.
+    xSea = xSeasons.sel(time=xSeasons.time.dt.month.isin(season_sel))
```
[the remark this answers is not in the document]
bonnland: Why is the suggested change better?  More general somehow?  Faster?  I'm not clear.

Plots/Contours/NCL_eof_1_1.py
dcherian: oh hmm... add a comment then?

dcherian: actually can't you pass `cplot` to the `clabel` function?

Plots/Contours/NCL_eof_1_1.py
dcherian: :+1: though thinking about it more; it might not be so bad. `sel` is a little more readable though you'll have to explain why you need the `+0.01`

Plots/Contours/NCL_eof_1_1.py
dcherian: agreed.

Plots/Contours/NCL_eof_1_1.py
dcherian: or the user could learn about `sizes` through this script...

Plots/Contours/NCL_eof_1_1.py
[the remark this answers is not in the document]
dcherian: Yeah, let's mention it so it's clear to someone who doesn't use NCL :)

Plots/Contours/NCL_eof_1_1.py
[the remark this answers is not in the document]
dcherian: isn't it clearer since `season_sel` is a single value? I think of `isin` as more appropriate for a list of possible values.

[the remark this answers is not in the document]
bonnland: I agree.  I misinterpreted which line was the original, and which was the change.   The equals sign is clearer.

Plots/Contours/NCL_eof_1_1.py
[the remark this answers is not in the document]
bonnland: OK, agreed.

Plots/Contours/NCL_eof_1_1.py
bonnland: OK, will do.

Plots/Contours/NCL_eof_1_1.py
bonnland: I see your point; done.

Plots/Contours/NCL_eof_1_1.py
bonnland: I tried doing this, and it did not behave well for me.    Strange triangular artifacts appear on the plot when I do this.    If you see a way around, or a different outcome than me, let me know.

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,328 @@
...
+# Diagnostic plot: show slice of SLP
+sliceSLP = SLP.sel(lat=slice(latS, latN), lon=slice(lonL, lonR))
+
+print_debug('\n\nsliceSLP:\n')
+print_debug(sliceSLP)
+
+###############################################################################
+# Create weights: sqrt(cos(lat))   [or sqrt(gw) ]
+
+deg2rad = 4. * atan(1.) / 180.
```
dcherian, line 163:
```suggestion
deg2rad = np.pi / 180.
```

Plots/Contours/NCL_eof_1_1.py
```diff
@@ -0,0 +1,328 @@
...
+clat = SLP['lat'].astype(dtype=np.float64)
```
dcherian, line 164: this may be easier to understand? no strong feelings either way

```suggestion
clat = SLP['lat'].astype('float64')
```

Plots/Contours/NCL_eof_1_1.py
erogluorhan: Good catch. Done.

Plots/Contours/NCL_eof_1_1.py
erogluorhan: Me, too, with any strong feeling between two, but made change as suggested.

PR #52, opened 2020-01-30
Remove /data submodule, use geocat.datafiles instead

khallock:
Replace relative paths to data files in scripts with `geocat.datafiles.get()` calls:
```
xr.open_dataset('../../data/netcdf_files/sst.nc')
```
becomes:
```
xr.open_dataset(geocat.datafiles.get('netcdf_files/sst.nc'))
```

This PR should stay open until all of the remaining PRs are merged because it removes the `data` directory and we don't want to break any new scripts that have not yet switched to using `geocat.datafiles`. We can add more commits to this PR to update new scripts as they get merged to master.


khallock:
For reference, the only remaining PRs with scripts that need to switch to `geocat.datafiles` are #46 and #51.

erogluorhan:
To me, we can merge this PR even in the cost of breaking those two new scripts as the recent geocat-viz PR (https://github.com/NCAR/geocat-viz/pull/5) will already break them in case it is merged.

mnlevy1981:
@erogluorhan a PR that breaks existing functionality elsewhere in the same repo (as merging this would do) is very different from an API-changing pull request in an upstream dependency (such as https://github.com/NCAR/geocat-viz/pull/5). For the latter, we can just pin the `geocat-viz` version number in the conda environment file to the old version and then update both the environment file and all dependent scripts in the same PR, so there is never a tag on master with broken scripts.

PR #53, opened 2020-02-03
Use geocat.viz.util for utility functions

khallock:
Remove locally defined utility functions that were duplicated throughout many example scripts.


clyne:
@khallock there are no reviewers assigned to this PR. I'd suggest adding anyone whose script is affected.

khallock:
@clyne Great idea! I'll add them all now.

PR #54, opened 2020-02-10
Cleanup Vector Plots

mnlevy1981:
The two main purposes of this PR are

1. Add "Concepts Illustrated" to the info in the header of the html pages
2. Use `from geocat.viz.util import ...` instead of `import geocat.viz as gcv`

I also made a few minor tweaks, like using the `xarray` wrapper to plot SST (meaning I don't need local copies of `lat` and `lon`) and adding the timestamp to the Vector_3 plot.


mnlevy1981:
Oh, I should add that I'm using `from geocat.viz import cmaps` instead of just `import cmaps`

erogluorhan:
Cleanup looks good to me.

PR #55, opened 2020-02-12
Fix xarray.DataArray keyword args in mask_1.py

khallock:
The xr_add_cyclic function defined in mask_1.py includes the 'encoding' keyword argument to the xarray.DataArray constructor that was deprecated as of Xarray 0.14 and was removed in 0.15.

I believe the xr_add_cyclic function will eventually go into geocat-viz, so we should make this change there as well.

### DIFF
--- a/Plots/Masking/mask_1.py
+++ b/Plots/Masking/mask_1.py
@@ -48,14 +48,14 @@ def xr_add_cyclic(da, coord):
 
     coords = da.coords.to_dataset()
     coords[coord] = cyclic_coord
-    return xr.DataArray(
+    da_out = xr.DataArray(
         cyclic_data,
         dims=da.dims,
         coords=coords.coords,
         attrs=da.attrs,
-        encoding=da.encoding,
     )
-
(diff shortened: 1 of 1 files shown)


erogluorhan:
Since `xr_add_cyclic` function has gone into geocat-viz (A new PR will introduce it when merged), this PR may be be closed without approval&merge. In either case, a refactor through geocat-examples soon will remove `xr_add_cyclic` function from `mask_1.py`.

khallock:
@erogluorhan I agree that this change will eventually be unnecessary once we move the `xr_add_cyclic` function to `geocat.viz.util.xr_add_cyclic_longitudes`.

However, I found that this example was broken when I helped a new user set up their geocat-examples conda environment (with Xarray v0.15) and wanted to add a quick fix to this repo in the meantime. NCAR/GeoCAT-viz#5 will likely be merged soon, but the refactoring PR for geocat-examples that you mentioned may still take some time before being merged.

Alternatively, we could just temporarily pin `xarray=0.14` in `conda_environment.yml`.

PR #67, opened 2020-05-20
NCL xy_7 example

kmpaul:
Here's what it should look like...


jukent:
Looks great! Maybe add the plot.

Issue #60, opened 2020-04-14
panel_35


khallock:
Description: This example shows how to draw three filled contour plots, attached along their Y axes, with a common labelbar and title

Script URL: http://www.ncl.ucar.edu/Applications/Scripts/panel_35.ncl

Plot URL: http://www.ncl.ucar.edu/Applications/Images/panel_35_lg.png

PR #71, opened 2020-05-21
Adding panel_35 example python file

hCraker:
References issue number #60.
Here is the plot in it's current state.
image

### DIFF
--- a/Plots/Panels/NCL_panel_35.py
+++ b/Plots/Panels/NCL_panel_35.py
@@ -3,10 +3,10 @@ NCL_panel_35.py
 ===============
 This script illustrates the following concepts:
    - Attaching three filled contour plots along Y axes
-   - Adding a common labelbar to attached plots
+   - Adding a common colorbar to attached plots
    - Adding a common title to attached plots
-   - Generating dummy data using "numpy.rand" and "signal.convolve"
-   - Drawing a custom labelbar
+   - Generating dummy data using "generate_2d_array"
+   - Drawing a custom colorbar
    - Drawing a custom title
    - Retrieving the bounding box of a plot
 
@@ -24,6 +24,179 @@ from geocat.viz import cmaps as gvcmaps
 import geocat.viz.util as gvutil
 
 from scipy import signal
+import math
+################################################################
+# Definition of generate_2d_array and helper functions from https://github.com/NCAR/pyngl/blob/develop/src/ngl/__init__.py
+#
+#  Random number generator for generate_2d_array.
+#
+#
+#  Globals for random number generator for generat_2d_array
+#
+dfran_iseq = 0
+dfran_rseq = [.749, .973, .666, .804, .081, .483, .919, .903,   \
+              .951, .960, .039, .269, .270, .756, .222, .478,   \
+              .621, .063, .550, .798, .027, .569, .149, .697,   \
+              .451, .738, .508, .041, .266, .249, .019, .191,   \
+              .266, .625, .492, .940, .508, .406, .972, .311,   \
+              .757, .378, .299, .536, .619, .844, .342, .295,   \
+              .447, .499, .688, .193, .225, .520, .954, .749,   \
+              .997, .693, .217, .273, .961, .948, .902, .104,   \
+              .495, .257, .524, .100, .492, .347, .981, .019,   \
+              .225, .806, .678, .710, .235, .600, .994, .758,   \
+              .682, .373, .009, .469, .203, .730, .588, .603,   \
+              .213, .495, .884, .032, .185, .127, .010, .180,   \
+              .689, .354, .372, .429                            \
+             ]
+
+#
+#  Random number generator for generate_2d_array.
+#
+def _dfran():
+  global dfran_iseq
+  global dfran_rseq
+  dfran_iseq = dfran_iseq % 100
+  r = dfran_rseq[dfran_iseq]
+  dfran_iseq = dfran_iseq + 1
+  return r
+
+def generate_2d_array(dims, num_low, num_high, minv, maxv, seed=0, \
+                      highs_at=None, lows_at=None):
+  """
+Generates smooth 2D arrays primarily for use in examples.
+array = generate_2d_array(dims, num_low, num_high, minv, maxv, seed=0,
+                          highs_at=None, lows_at=None)
+dims -- a list (or array) containing the dimensions of the
+        two-dimensional array to be returned.
+num_low, num_high -- Integers representing the approximate minimum 
+                     and maximum number of highs and lows that the 
+                     output array will have. They must be in the 
+                     range 1 to 25. If not, then they will be set to 
+                     either 1 or 25.
+minv, maxv -- The exact minimum and maximum values that the output array 
+              will have.
+iseed -- an optional argument specifying a seed for the random number
+         generator.  If iseed is outside the range 0 to 99, it will
+         be set to 0.
+lows_at -- an optional argument that is a list of coordinate  
+           pairs specifying where the lows will occur.  If this
+           argument appears, then its length must equal num_low and
+           the coordinates must be in the ranges specified in dims.
+highs_at -- an optional argument that is a list of coordinate  
+            pairs specifying where the highs will occur.  If this
+            argument appears, then its length must equal num_high and
+            the coordinates must be in the ranges specified in dims.
+  """  
+#
+#  Globals for random numbers.
+#
+  global dfran_iseq
+  dfran_iseq = seed
+#
+#  Check arguments.
+#
+  try:
+    alen = len(dims)
+  except:
+    print("generate_2d_array: first argument must be a list, tuple, or array having two elements specifying the dimensions of the output array.")
+    return None
+  if (alen != 2):
+    print("generate_2d_array: first argument must have two elements specifying the dimensions of the output array.")
+    return None
+  if (int(dims[0]) <=1 and int(dims[1]) <=1):
+    print("generate_2d_array: array must have at least two elements.")
+    return None
+  if (num_low < 1):
+    print("generate_2d_array: number of lows must be at least 1 - defaulting to 1.")
+    num_low = 1
+  if (num_low > 25):
+    print("generate_2d_array: number of lows must be at most 25 - defaulting to 25.")
+    num_high =25
+  if (num_high < 1):
+    print("generate_2d_array: number of highs must be at least 1 - defaulting to 1.")
+    num_high = 1
+  if (num_high > 25):
+    print("generate_2d_array: number of highs must be at most 25 - defaulting to 25.")
+    num_high =25
+  if (seed > 100 or seed < 0):
+    print("generate_2d_array: seed must be in the interval [0,100] - seed set to 0.")
+    seed = 0
+  if not lows_at is None:
+    if (len(lows_at) != num_low):
+      print("generate_2d_array: the list of positions for the lows must be the same size as num_low.")
+  if not highs_at is None:
+    if (len(highs_at) != num_high):
+      print("generate_2d_array: the list of positions for the highs must be the same size as num_high.")
+
+#
+#  Dims are reversed in order to get the same results as the NCL function.
+#
+  nx = int(dims[1])
+  ny = int(dims[0])
+  out_array = np.zeros([nx,ny],'f')
+  tmp_array = np.zeros([3,51],'f')
+  fovm = 9./float(nx)
+  fovn = 9./float(ny)
+  nlow = max(1,min(25,num_low))
+  nhgh = max(1,min(25,num_high))
+  ncnt = nlow + nhgh
+
+#
+  for k in range(num_low):
+    if not lows_at is None:
+      tmp_array[0,k] =  float(lows_at[k][1])   # lows at specified locations.
+      tmp_array[1,k] =  float(lows_at[k][0])
+      tmp_array[2,k] = -1.
+    else:
+      tmp_array[0,k] =  1.+(float(nx)-1.)*_dfran() # lows at random locations.
+      tmp_array[1,k] =  1.+(float(ny)-1.)*_dfran() # lows at random locations.
+      tmp_array[2,k] = -1.
+  for k in range(num_low,num_low+num_high):
+    if not highs_at is None:
+      tmp_array[0,k] =  float(highs_at[k-num_low][1])  # highs locations
+      tmp_array[1,k] =  float(highs_at[k-num_low][0])  # highs locations
+      tmp_array[2,k] =  1.
+    else:
+      tmp_array[0,k] =  1.+(float(nx)-1.)*_dfran() # highs at random locations.
+      tmp_array[1,k] =  1.+(float(ny)-1.)*_dfran() # highs at random locations.
+      tmp_array[2,k] =  1.
+  
+  dmin =  1.e+36
+  dmax = -1.e+36
+  midpt = 0.5*(minv + maxv)
+  for j in range(ny):
+    for i in range(nx):
+      out_array[i,j] = midpt
+      for k in range(ncnt):
+        tempi = fovm*(float(i+1)-tmp_array[0,k])
+        tempj = fovn*(float(j+1)-tmp_array[1,k])
+        temp  = -(tempi*tempi + tempj*tempj)
+        if (temp >= -20.):
+          out_array[i,j] = out_array[i,j] +    \
+             midpt*tmp_array[2,k]*math.exp(temp)
+      dmin = min(dmin,out_array[i,j])
+      dmax = max(dmax,out_array[i,j])
+                   
+  out_array = (((out_array-dmin)/(dmax-dmin))*(maxv-minv))+minv
+
+  del tmp_array
+
+  return np.transpose(out_array,[1,0])
+
+
+def _get_double(obj,name):
+  return(NhlGetDouble(_int_id(obj),name))
+
+def _get_double_array(obj,name):
+  return(NhlGetDoubleArray(_int_id(obj),name))
+
+###############################################################################
+# Create dummy data
+nx = 100
+ny = 100
+data1 = generate_2d_array((ny,nx), 10, 10, -19., 16., 0)
+data2 = generate_2d_array((ny,nx), 10, 10, -28., 15., 1)
+data3 = generate_2d_array((ny,nx), 10, 10, -25., 18., 2)
 
 ###############################################################################
 # Create figure and axes using gvutil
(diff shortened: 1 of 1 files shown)


hCraker:
Currently the plot looks like this. @NCAR/ncl-examples what would be a better way to generate random data that isn't so noisy? If I add in the contour lines, the plot becomes illegible.

image

kmpaul:
@NCAR/ncl-examples

jukent:
Creating random data that looks natural is a very difficult problem. Did you try convolving the random noise with a trigonometric function based on x and y locations?

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,61 @@
+"""
+NCL_panel_1.py
```
erogluorhan, line 2: Should be NCL_panel_35.py

hCraker: Good catch. I copied some of the formatting from other completed examples and I missed this.

jukent:
I noticed you are importing packages that you are not using (namely the geocat data package). It is best to delete that line.

hCraker:
Here's what it looks like with the synthetic data. I didn't know how to create data that is more similar to the original plot.
figure

jukent:
I think this looks great! If you want more than one maximum/minimum per subplot you could change the period of the trig function used.

Did you figure out how to maintain square subplots?

hCraker:
@jukent I'll try changing the periods on the trig functions. 

I called `axs[0].set_aspect(aspect='equal')` on each of the subplot axes and it maintains square plots even when the colorbar size is changed or if it's removed.

jukent:
I'm glad that worked! I think the dummy data is fine as is, but I would wait to hear the opinion of the @NCAR/geocat team.

mnlevy1981:
>what would be a better way to generate random data that isn't so noisy?

you can see documentation on the NCL function `generate_2d_array()` here: https://www.ncl.ucar.edu/Document/Functions/Built-in/generate_2d_array.shtml

>  This function generates a two-dimensional array (dimensioned `dsizes`) of pseudo random numbers using a sum of exponentials. This function is especially useful for generating random data for nice 2D graphics.
> 
> The values will fall exactly within the range [`dlow`,`dhigh`] and there will be approximately `mlow` lows and `mhigh` highs.

The function was ported to PyNGL at one point -- @khallock, does that mean it's likely available in `geocat.comp` as well? (Sorry @hCraker, I was hoping to be able to dig up an actual formula for you to use... though if there's already a python version of the function that's even better).

kmpaul:
(Hey, @mnlevy1981!  Thanks for the assistance!)

clyne:
> > what would be a better way to generate random data that isn't so noisy?
> 
> you can see documentation on the NCL function `generate_2d_array()` here: https://www.ncl.ucar.edu/Document/Functions/Built-in/generate_2d_array.shtml
> 
> > This function generates a two-dimensional array (dimensioned `dsizes`) of pseudo random numbers using a sum of exponentials. This function is especially useful for generating random data for nice 2D graphics.
> > The values will fall exactly within the range [`dlow`,`dhigh`] and there will be approximately `mlow` lows and `mhigh` highs.
> 
> The function was ported to PyNGL at one point -- @khallock, does that mean it's likely available in `geocat.comp` as well? (Sorry @hCraker, I was hoping to be able to dig up an actual formula for you to use... though if there's already a python version of the function that's even better).

Good catch @dcherian . There is a Python version of this function in PyNGL, but it has not yet been ported to GeoCAT-comp. @hCraker the Python version of generate_2d_array is in the file ./src/ngl/__init__.py found in the PyNGL repo: https://github.com/NCAR/pyngl

The code is pretty ugly, and would need to be cleaned up before we'd include it in GeoCAT, but it might worth the experiment to try it as a source for your data.

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,79 @@
...
+p = axs[2].contourf(x, y, c, cmap=newcmap, levels=12)
+axs[2].contour(p, colors='k', linestyles='solid', linewidths=0.5)
+
+cbar = fig.colorbar(p, orientation='horizontal', ax=axs, ticks=np.arange(20, 85, 5), shrink=0.75)
+
+# Add title
+fig.suptitle("Three dummy plots attached along Y axes", fontsize=18, fontweight='bold')
+
+# Get bounding box
+bbox = fig.get_window_extent()
```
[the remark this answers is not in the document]
hCraker: The NCL example shows how to retrieve the bounding box of the plot, but it doesn't manipulate it in anyway. This is the python equivalent

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,79 @@
...
+gvutil.add_major_minor_ticks(axs[1], x_minor_per_major=4, y_minor_per_major=4)
+axs[1].tick_params(axis='both', which='both', top=True, left=False, right=False)
+axs[1].set_aspect(aspect='equal')
+
+gvutil.set_axes_limits_and_ticks(axs[2], xticks=np.arange(0, 120, 20), yticks=np.arange(0, 120, 20), xticklabels=np.arange(0, 100, 20), yticklabels=np.arange(0, 100, 20))
+gvutil.add_major_minor_ticks(axs[2], x_minor_per_major=4, y_minor_per_major=4)
+axs[2].tick_params(axis='both', which='both', top=True, left=False, right=True)
+axs[2].set_aspect(aspect='equal')
+
+# Create dummy data
```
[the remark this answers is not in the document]
hCraker: Do you mean separate it by adding a line of hashes to make a new cell in a Jupyter notebook?

hCraker:
> One more thing, the Sphinx-generated RTD of this script has a huge vertical space between code and plot as follows (I could not figure it out because the image size looks good; does anybody have an idea where this much spacing is coming from?):
@erogluorhan I realized that I was calling `plt.figure` and `plt.subplots` so I was making a blank figure by accident. I believe that is where the extra white space was coming from.

erogluorhan:
> > One more thing, the Sphinx-generated RTD of this script has a huge vertical space between code and plot as follows (I could not figure it out because the image size looks good; does anybody have an idea where this much spacing is coming from?):
> 
> @erogluorhan I realized that I was calling `plt.figure` and `plt.subplots` so I was making a blank figure by accident. I believe that is where the extra white space was coming from.

That's great!

Plots/Panels/NCL_panel_35.py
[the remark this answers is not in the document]
erogluorhan: Yes, and move it to a place right after package import block, which gives us the following workflow of the code:

Import packages -> Generate or read data -> Plot

Plots/Panels/NCL_panel_35.py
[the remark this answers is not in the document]
erogluorhan: Got it, thanks for the calrification!

In fact, NCL script makes use of that bounding box through following codes:

```
;---Get bounding box that encloses plot, so we know where the top edge is.
  bb  = NhlGetBB(plot)
  top = bb(0)             ; top edge of plot
...
txres@txPosYF       = top + 0.01   ; Add a little bit of a gap between top of plot 
                                     ; and bottom of title.
...
;---Get bounding box that encloses plot, so we know where the bottom edge is.
  bb  = NhlGetBB(plot)
  bot = bb(1)             ; bottom edge of plot
...
lbres@vpXF               = 0.15         ; Position labelbar at         
lbres@vpYF               = bot          ; bottom edge of plot
```

However, I am not sure if we should deal with it the same way or just retrieve it bbox and leave it as is as you did. Hi @clyne @kmpaul @jukent , do you have any preference on this?

hCraker:
Here is what the plot looks like now. It seems that the colors are inverted. I think this is due to how the `generate_2d_array` function works rather than the colorbar, but I'm not certain.
image

hCraker:
The last commit ensured that the color map is the same for each subplot. This fixed the whitespace on the third subplot.
image

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+axs[0].contour(filled1, colors='k', linestyles='solid', linewidths=0.5)
+filled2 = axs[1].contourf(data2, cmap=newcmap, levels=contour_levels)
+axs[1].contour(filled2, colors='k', linestyles='solid', linewidths=0.5)
+filled3 = axs[2].contourf(data3, cmap=newcmap, levels=contour_levels)
+axs[2].contour(filled3, colors='k', linestyles='solid', linewidths=0.5)
+
+cbar = plt.colorbar(filled3, orientation='horizontal', ax=axs, ticks=np.arange(-28,20,4), shrink=0.75, drawedges=True, pad=0.1)
+
+# Add title
+title=fig.suptitle("Three dummy plots attached along Y axes", horizontalalignment='center', y=0.9, fontsize=18, fontweight='bold', fontfamily='sans-serif')
```
erogluorhan, line 235: Since `title` variable is not used later, we can directly call `fig.suptitle()` without assigning its output to `title`.

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+contour_levels = np.arange(-32,24,4) # levels=contour_levels ensures that each plot has the same scale
+
+filled1 = axs[0].contourf(data1, cmap=newcmap, levels=contour_levels)
+axs[0].contour(filled1, colors='k', linestyles='solid', linewidths=0.5)
+filled2 = axs[1].contourf(data2, cmap=newcmap, levels=contour_levels)
+axs[1].contour(filled2, colors='k', linestyles='solid', linewidths=0.5)
+filled3 = axs[2].contourf(data3, cmap=newcmap, levels=contour_levels)
+axs[2].contour(filled3, colors='k', linestyles='solid', linewidths=0.5)
+
+cbar = plt.colorbar(filled3, orientation='horizontal', ax=axs, ticks=np.arange(-28,20,4), shrink=0.75, drawedges=True, pad=0.1)
```
erogluorhan, line 232: Since `cbar` variable is not used later, we can directly call `plt.colorbar()` without assigning its output to `cbar`.

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+ny = 100
+data1 = generate_2d_array((ny,nx), 10, 10, -19., 16., 0)
+data2 = generate_2d_array((ny,nx), 10, 10, -28., 15., 1)
+data3 = generate_2d_array((ny,nx), 10, 10, -25., 18., 2)
+
+###############################################################################
+# Create figure and axes using gvutil
+fig, axs = plt.subplots(1, 3, figsize=(12, 6), sharex='all', sharey='all', gridspec_kw={'wspace': 0})
+
+gvutil.set_axes_limits_and_ticks(axs[0], xticks=np.arange(0, 120, 20), yticks=np.arange(0, 120, 20), xticklabels=np.arange(0, 100, 20), yticklabels=np.arange(0, 100, 20))
```
erogluorhan, line 205: Please provide some comments about what these convenience functions do like following (At least for this first occurence in the code):

```
# Use geocat.viz.util convenience function to set axes tick values

# Use geocat.viz.util convenience function to add minor and major tick lines
```

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+gvutil.add_major_minor_ticks(axs[0], x_minor_per_major=4, y_minor_per_major=4)
+axs[0].tick_params(axis='both', which='both', top=True, left=True, right=False)
```
erogluorhan, line 207: Fyi: Setting `top=True` might not be needed because it is already set `True` in `gvutil.add_major_minor_ticks()`

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+axs[0].set_aspect(aspect='equal')
+
+gvutil.set_axes_limits_and_ticks(axs[1], xticks=np.arange(0, 120, 20), yticks=np.arange(0, 120, 20), xticklabels=np.arange(0, 100, 20), yticklabels=np.arange(0, 100, 20))
+gvutil.add_major_minor_ticks(axs[1], x_minor_per_major=4, y_minor_per_major=4)
+axs[1].tick_params(axis='both', which='both', top=True, left=False, right=False)
```
erogluorhan, line 212: Same Fyi

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,240 @@
...
+axs[1].set_aspect(aspect='equal')
+
+gvutil.set_axes_limits_and_ticks(axs[2], xticks=np.arange(0, 120, 20), yticks=np.arange(0, 120, 20), xticklabels=np.arange(0, 100, 20), yticklabels=np.arange(0, 100, 20))
+gvutil.add_major_minor_ticks(axs[2], x_minor_per_major=4, y_minor_per_major=4)
+axs[2].tick_params(axis='both', which='both', top=True, left=False, right=True)
```
erogluorhan, line 217: Same Fyi

Plots/Panels/NCL_panel_35.py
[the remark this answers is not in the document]
clyne: It would seem that the only point of retrieving the bounding box in the NCL plot is to be able to adjust the position of the label bar. If the Python version of this example can position the annotation correctly without knowing the bounding box extent I would say leave out the bounding box fetching.

hCraker:
> Glad you were able to port the data generation function from PyNGL without any trouble. The plot now looks very much like its NCL counterpart. However, I'm wondering about the colors being flipped. It looks like the signs may be reversed on the data.

@clyne I also thought that the signs were simply flipped, but I tried negating the output and this is the result. It's not quite what we are looking for. I'll keep looking for what's causing this.
image

hCraker:
@clyne So I figured out the issue with the color flip. I swapped `minv` and `maxv` values and the plot looks like what we want know. The weird part is that the argument labels are inconsistent with my solution. Can you take a look at it and tell me what you think? The function definition says `minv` should come first in the arg list, but putting `maxv` first is what worked for this plot.

image

clyne:
> > Glad you were able to port the data generation function from PyNGL without any trouble. The plot now looks very much like its NCL counterpart. However, I'm wondering about the colors being flipped. It looks like the signs may be reversed on the data.
> 
> @clyne I also thought that the signs were simply flipped, but I tried negating the output and this is the result. It's not quite what we are looking for. I'll keep looking for what's causing this.
> image

It may simply be that the NCL and Python versions of the random data generation function produce different results. If they do, that is fine. We might simply include a comment in the script indicating so. If they do not produce different data we should figure out why we are getting different plots.

jukent:
> @clyne So I figured out the issue with the color flip. I swapped `minv` and `maxv` values and the plot looks like what we want know. The weird part is that the argument labels are inconsistent with my solution. Can you take a look at it and tell me what you think? The function definition says `minv` should come first in the arg list, but putting `maxv` first is what worked for this plot.
> 
> `out_array = (((out_array-dmin)/(dmax-dmin))*(maxv-minv))+minv` This is the only line where the order of `minv` and `maxv` seems to matter.
> 
> image

@clyne Should we change the vmin and vmax definitions in the function definition? We want to make sure we understand what is going on here before doing so.

mnlevy1981:
> > `out_array = (((out_array-dmin)/(dmax-dmin))*(maxv-minv))+minv` This is the only line where the order of `minv` and `maxv` seems to matter.

I think there's actually a bug in the pyNGL version... if you look at the [original fortran](https://github.com/NCAR/ncl/blob/ed6016bf579f8c8e8f77341503daef3c532f1069/ni/src/lib/nfpfort/gendat.f), there's one other place where the order matters: https://github.com/NCAR/ncl/blob/ed6016bf579f8c8e8f77341503daef3c532f1069/ni/src/lib/nfpfort/gendat.f#L61

```
                  IF (TEMP.GE.-20.D0) DATA(I,J) = DATA(I,J) +
     +                .5D0* (DHGH-DLOW)*CCNT(3,K)*EXP(TEMP)
```

but the [equivalent python line](https://github.com/NCAR/GeoCAT-examples/pull/71/files#diff-a50f70233829fc0c9da4c63e83097cf2R172) reuses `midpt`:

```
        if (temp >= -20.):
          out_array[i,j] = out_array[i,j] +    \
             midpt*tmp_array[2,k]*math.exp(temp)
```

I'll comment inline with a suggested fix

Plots/Panels/NCL_panel_35.py
```diff
@@ -0,0 +1,238 @@
...
+  for j in range(ny):
+    for i in range(nx):
+      out_array[i,j] = midpt
+      for k in range(ncnt):
+        tempi = fovm*(float(i+1)-tmp_array[0,k])
+        tempj = fovn*(float(j+1)-tmp_array[1,k])
+        temp  = -(tempi*tempi + tempj*tempj)
+        if (temp >= -20.):
+          out_array[i,j] = out_array[i,j] +    \
+             midpt*tmp_array[2,k]*math.exp(temp)
```
mnlevy1981, line 172: I think this line should be

```
0.5*(maxv - minv)*tmp_array[2,k]*math.exp(temp)
```

to match the [fortran code](https://github.com/NCAR/ncl/blob/ed6016bf579f8c8e8f77341503daef3c532f1069/ni/src/lib/nfpfort/gendat.f#L61)

mnlevy1981: (modulo proper whitespace usage, of course :)

jukent: We're going to test this now (with switching the minv and maxv back to the original order in the function call)

hCraker: Thank you that worked!

erogluorhan:
@clyne Are you getting some errors? If so, you may want ot share that with me.

clyne:
@erogluorhan I'm not getting any errors. So I assume it is working for you?

erogluorhan:
@clyne Yes, it is working. Could you clarify what version of **geocat-viz** is installed in you **geocat-examples** and could you share the terminal output when you run `make html` command?

clyne:
@hCraker Orhan and I confirmed that "make html" no longer produces an html file with a plot. We get a blank image (actually two blank images). Can you verify that the current snapshot works for you?

hCraker:
@clyne I have not used `make html` before. How do I call it and in which directory?

erogluorhan:
> @clyne I have not used `make html` before. How do I call it and in which directory?

From the root directory of your repository, you can call it. Please have a look at **Step 6** of [Contributor's Guide](https://github.com/NCAR/GeoCAT-examples/blob/master/CONTRIBUTING.md) for further information (e,g, where html files are generated etc.).

Let us know anything further is needed.

clyne:
> @clyne I have not used `make html` before. How do I call it and in which directory?

Good question. To verify that your script is playing nice with Sphinx, and producing a usable Jupyter notebook, you should run the command "make html" from a terminal with your current working directory set to the root of the local copy of your repo. Look for any error messages. Then open the file _build/html/index.html with a web browser, and examine the entry that was created for your plot.

clyne:
Or you can following the directions in the contributor's guide as Orhan more wisely than I pointed out :-)

hCraker:
@clyne @erogluorhan The script is also not producing plots on the html page when I run it. I will look into it.

erogluorhan:
> @clyne @erogluorhan The script is also not producing plots on the html page when I run it. I will look into it.

Hi @**Rose Craker**  I have found the cause: It is directly related to Sphinx's bad handling of notebook cells (which we create by putting a line with series of ###'s) throughout the plotting lines of codes. As a walk-around for the time, we have to keep all of plotting lines of codes in the same notebook cell, i.e. you should remove Line 232 (###'s that separate plotting block into two cells). I am putting this information to your corresponding PR as a comment as well.

We should look at solving this at Sphinx side, if possible, in the long run.

hCraker:
> > @clyne @erogluorhan The script is also not producing plots on the html page when I run it. I will look into it.
> 
> Hi @**Rose Craker** I have found the cause: It is directly related to Sphinx's bad handling of notebook cells (which we create by putting a line with series of ###'s) throughout the plotting lines of codes. As a walk-around for the time, we have to keep all of plotting lines of codes in the same notebook cell, i.e. you should remove Line 232 (###'s that separate plotting block into two cells). I am putting this information to your corresponding PR as a comment as well.
> 
> We should look at solving this at Sphinx side, if possible, in the long run.

I have just pushed that change. Please let me know if there are any other edits that need to be made to this PR.

Issue #22, opened 2019-12-20
unique_5


clyne:
Description: Bar charts

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/unique_5.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/unique_5_lg.png

jukent:
Missing 4 files:      
"hadisst.nino34.1900-2003.nc"
"b30004.nino34.0100-0799.nc"
"b30009.nino34.0100-0599.nc"
"b20007.nino34.nc"

khallock:
I searched for these files but couldn't find them anywhere. It looks like the original script was written by @phillips-ad, perhaps he can help track down the files.

However even if we can't find the original files, I think the concepts illustrated by this plot could still be demonstrated with synthetic data.

jukent:
Okay. You're right- it is more important to demonstrate how to make this type of plot than to produce the identical plot.

PR #72, opened 2020-05-22
created trajectory plot for example traj_1

clairefio:
trajectory plot

Issue #61. Plot created with randomly generated lat/long coordinates.

PR #74, opened 2020-05-22
Unique 5

clairefio:
Issue #22:

unique_5


clairefio:
unique_5

clairefio:
unique_5
Fixed the colors, y-axis label, and extra EOF space @jukent

Plots/Bar/NCL_unique_5.py
```diff
@@ -0,0 +1,91 @@
...
+w = 0.15
+
+# Create subplots for each category
+sub = plt.subplot(111)
+sub.bar(x+w, obs, width=0.15, color=color_list[0], edgecolor='black', linewidth=0.25, align='center')
+sub.bar((x+(2*w)), ccsm2_t42, width=0.15, color=color_list[1], edgecolor='black', linewidth=0.25, align='center')
+sub.bar(x+3*w, ccsm3_t42, width=0.15, color=color_list[2], edgecolor='black', linewidth=0.25, align='center')
+sub.bar(x+4*w, ccsm3_t85, width=0.15, color=color_list[3], edgecolor='black', linewidth=0.25, align='center')
+
+# Add label to each bar
```
erogluorhan, line 73: This `for-loop` seems to work well, but I would personally not use it. Instead I would remove it and modify **Line 81** and **Line 84** as follows to plot the same month names on **x-axis**:

`gvutil.set_axes_limits_and_ticks(ax, ylim=(0.4, plot_y_max), xticks=x, xticklabels=labels, yticks=np.linspace(0.4, plot_y_max, 5))`

`gvutil.add_major_minor_ticks(ax, x_minor_per_major=1, y_minor_per_major=4, labelsize=12)`

This way, it would also solve another issue: The current version of this plot does not have major ticks for months on top and bottom x-axis but **NCL**'s original plot has. The above two lines add those major ticks for months into the plot. However, there is also an issue with my solution here: Major ticks are displayed in the beginning of every 4-bar group, but NCL displays those major ticks in the middle of each 4-bar group. You may further play with it to walk around this issue.

Plots/Trajectories/NCL_traj_1.py
```diff
@@ -0,0 +1,86 @@
+"""
```
erogluorhan, line 1: As far as I understand, `Plots/Trajectories/NCL_traj_1.py` is already being implemented within another PR #72 by the branch **traj_1**. The reason for this file `Plots/Trajectories/NCL_traj_1.py` being also changed here in this PR  seems to me that it was first started being implemented by this branch (unique_5) but then left as is.

Correct me if I am wrong and please clarify the reason. Otherwise, could you please remove this file from this PR, then we can handle it in PR #72?

Plots/Bar/NCL_unique_5.py
```diff
@@ -0,0 +1,91 @@
...
+# Import packages:
+import numpy as np
+import matplotlib.pyplot as plt
+import random
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Generate labels:
+x = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12])
```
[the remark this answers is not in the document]
jukent: Seconding this!

Plots/Bar/NCL_unique_5.py
clairefio: It looks like the NCL graph displays them at the beginning as well, the only difference I can spot with the python version is that the last tick on the top and bottom axes is small. Do you know why this is?
Inkednew graph_LI

erogluorhan: 1. "the only difference I can spot with the python version is that the last tick on the top and bottom axes is small. Do you know why this is?"

You can play with the `xlim` argument of the following function to make disappear that final minor tick:

`gvutil.set_axes_limits_and_ticks(ax, ylim=(0.4, plot_y_max), xticks=x, xticklabels=labels, yticks=np.linspace(0.4, plot_y_max, 5))`

2. "It looks like the NCL graph displays them at the beginning as well"

The [NCL plot](https://www.ncl.ucar.edu/Applications/Images/unique_5_lg.png) given in the corresponding issue seems to be placing those tick in the middle of each 4-bar group (i.e. centers the bar groups to each month tick), Please see:

image

However, I am not sure if we should worry about it. Hi @jukent do you have a strong feeling about it?

jukent: I like the way they look in the middle better.

Here is an example of how to align them in the center:
https://stackoverflow.com/questions/31067389/matplotlib-barplot-generic-way-of-setting-the-xticklabels-in-the-middle-of-the

erogluorhan:
Hi @michaelavs , is it possible to put your review anytime soon for this?

Issue #58, opened 2020-04-14
polar_1


clyne:
Description: Drawing black-and-white contours over a polar stereographic map

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/polar_1.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/polar_1_lg.png

Category: Map projections

PR #70, opened 2020-05-21
Adding NCL_polar_1_lg example python file

michaelavs:
Closes issue #58
polar_1

PR #68, opened 2020-05-20
Adding NCL_polar_1_lg example python file

michaelavs:
polar_1

PR #75, opened 2020-05-26
Adding NCL_polar_1_lg.py example python file 

michaelavs:
Closes issue [#58](https://github.com/NCAR/GeoCAT-examples/issues/58)

image

### DIFF
--- a/Plots/Contours/NCL_polar_1_lg.py
+++ b/Plots/Contours/NCL_polar_1_lg.py
@@ -17,23 +17,23 @@ from geocat.viz import util as gvutil
 
 # Open a netCDF data file using xarray default engine and load the data into xarrays
 ds = xr.open_dataset(gdf.get("netcdf_files/uv300.nc"))
-U = ds.U[1, :, :]
+U=ds.U[1,:,:]
 
 # Fix the artifact of not-shown-data around 0 and 360-degree longitudes
 wrap_U = gvutil.xr_add_cyclic_longitudes(U, "lon")
 
 # Generate figure (set its size (width, height) in inches)
-fig = plt.figure(figsize=(10, 10))
+fig = plt.figure(figsize=(10,10))
 
 # Generate axes, using Cartopy, drawing coastlines, and adding features
-fig = plt.figure(figsize=(10, 10))
+fig = plt.figure(figsize=(10,10))
 projection = ccrs.NorthPolarStereo()
 ax = plt.axes(projection=projection)
 ax.coastlines(linewidths=0.5)
 ax.add_feature(cfeature.LAND, facecolor='lightgray')
 
-# OERO: Set extent to include latitudes between 0 and 40 and longitudes between -180 and 180 only
-ax.set_extent([-180, 180, 0, 40], ccrs.PlateCarree())
+# Set extent to include latitudes between 0 and 40 and longitudes between -180 and 180 only
+ax.set_extent([-180,180,0,40], ccrs.PlateCarree())
 
 # Set draw_labels to False so that you can manually manipulate it later
 gl = ax.gridlines(ccrs.PlateCarree(), draw_labels=False, linestyle="--", color='k')
@@ -55,17 +55,20 @@ etick = ['0'] + ['%d$^\circ$E' % tick for tick in ticks if (tick != 0) & (tick !
 wtick = ['%d$^\circ$W' % tick for tick in ticks if (tick != 0) & (tick != 180)]
 labels = etick + wtick
 xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -114, -140]
-yticks = [-3, -2, -1, -1, -1, -1, -3, -7, -7, -7, -7, -7]
+yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]
+
+
+
 
 for xtick, ytick, label in zip(xticks, yticks, labels):
     ax.text(xtick, ytick, label, transform=ccrs.Geodetic())
 
 
 # Contour-plot U-data
-p = wrap_U.plot.contour(ax=ax, vmin=-8, vmax=16, transform=ccrs.PlateCarree(),
-                    levels=np.arange(-12,44,4), linewidths=0.5, cmap='k', add_labels=False)
+p = wrap_U.plot.contour(ax=ax,vmin=-8,vmax=16, transform=ccrs.PlateCarree(),
+                    levels = np.arange(-12,44,4), linewidths=0.5, cmap='k', add_labels=False)
 
-ax.clabel(p, np.arange(-8,17,8), fmt='%d', inline=1, fontsize=14)
+ax.clabel(p, np.arange(-8,17,8),fmt='%d', inline=1, fontsize=14)
 
 # # Use geocat.viz.util convenience function to add titles to left and right of the plot axis.
 gvutil.set_titles_and_labels(ax, lefttitle="Zonal Wind", righttitle="m/s")
--- a/Plots/Contours/NCL_polar_1_lg.py
+++ b/Plots/Contours/NCL_polar_1_lg.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Tue May 12 09:38:46 2020
+
+@author: misi1684
+"""
+import numpy as np
+import xarray as xr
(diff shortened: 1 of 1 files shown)


michaelavs:
This pull request has been re-named from a now closed "NCL_proj_2_lh.py" PR. The comments from the previous PR are still available in the closed PR.

Plots/Contours/NCL_polar_1_lg.py
```diff
@@ -0,0 +1,115 @@
...
+#Plot:
+    
+# Generate figure (set its size (width, height) in inches)
+fig = plt.figure(figsize=(10, 10))
+
+# Generate axes, using Cartopy, drawing coastlines, and adding features
+fig = plt.figure(figsize=(10, 10))
+projection = ccrs.NorthPolarStereo()
+ax = plt.axes(projection=projection)
+ax.coastlines(linewidths=0.5)
```
clairefio, line 47: I'm not sure if this matters, but in the original NCL plot the land is not outlined.

erogluorhan: Good catch, it matters.

jukent: You could turn coastlines off. I recommend doing this so they don't distract from the contour lines.

michaelavs: Switching them off now. Do you mind also posting the code that fixes the latitude line drop out at 150E? I will fix both issues and then upload the new file

Plots/Contours/NCL_polar_1_lg.py
```diff
@@ -0,0 +1,115 @@
...
+wrap_U = gvutil.xr_add_cyclic_longitudes(U, "lon")
+
+###############################################################################
+#Plot:
+    
+# Generate figure (set its size (width, height) in inches)
+fig = plt.figure(figsize=(10, 10))
+
+# Generate axes, using Cartopy, drawing coastlines, and adding features
+fig = plt.figure(figsize=(10, 10))
```
erogluorhan, line 44: Should remove this as it is a duplicate of **Line 41** and leads to a large vertical spacing in Sphinx-generated html file.

Plots/Contours/NCL_polar_1_lg.py
jukent: I don't think I caught what that fix was. @erogluorhan Did you know how to fix the latitude line drop out at 150E?

erogluorhan: I think that both of latitudinal and longitudinal line drop out issues were resolved during the previously closed ancestor PR of this PR (PR #68)? So, the latest version of the plot looks like it is already resolved:

image

If you are explicitly looking for the exact codes that were the solution of this issue, I can try to grab them from commits of PR #68. Please let me know.

pilotchute: The Longitude labels on the left hand side of the graph seem to be out of position when compared to the right hand side. It's just a kerning issue, so we may not need to address it. 30˚ and 150˚ W seem to be the worst, while 90˚ W is fine.

erogluorhan: Thanks for bringing this up, @pilotchute! Indeed, we are manually adjusting those labels and their positions as a walk-around to the labeling issue with circularly-bounded `ccrs.NorthPolarStereo() `projections. Therefore, I believe that this issue of label positioning could be solved by playing with the following codes:

```
# Manipulate longitude labels (0, 30 E, 60 E, ..., 30 W, etc.)
ticks = np.arange(0, 210, 30)
etick = ['0'] + [r'%d$^\circ$E' %
                 tick for tick in ticks if (tick != 0) & (tick != 180)] + ['180']
wtick = [
    r'%d$^\circ$W' %
    tick for tick in ticks if (
        tick != 0) & (
            tick != 180)]
labels = etick + wtick
xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -114, -140]
yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]


for xtick, ytick, label in zip(xticks, yticks, labels):
    ax.text(xtick, ytick, label, transform=ccrs.Geodetic())
```

@michaelavs , does that make sense?

michaelavs: @erogluorhan I will look into the position they should be in, I had tweaked it a bit previously, so I think it will be a fairly simple fix. Also, I was looking through the commits from the previous version of this PR and couldn't find where the code was updated, do you mind looking into where it may be? I can then add it for when I commit the longitude label changes as well

erogluorhan: Hi @michaelavs I walked through each of those commits in PR #70 and #68, but there is nothing related to this issue. Indeed, I now believe that there should not be any manipulation of our own to overcome such an issue because the code that manages those latitude and longitude lines is the following:

```
# Manipulate latitude and longitude gridline numbers and spacing
gl.ylocator = mticker.FixedLocator(np.arange(0, 90, 15))
gl.xlocator = mticker.FixedLocator(np.arange(-180, 180, 30))
```

So, it should all be managed by `matplotlib.ticker`.

I doubt here whether the very first image with those line drop out issue was taken by a screenshot application, and those "drop-outs" were only a resolution artifact. Can you confirm whether you have observed those artifacts on the originally generated plot?

michaelavs: @erogluorhan I just tried to run the original code you had sent me when we first went over the lat/lon issue to see if I had made an edit that would have caused this issue, and the line drop out is occurring at 150E in the original as well. You're able to run the code and there isn't a line dropout correct? If so, there may just be something wrong on my side that isn't necessarily code related that is causing the output to drop the line at 150E

erogluorhan: @michaela Yes, I tried the original code I had sent you initially, and it generates the following plot without any lat/lon drop-out:

image

Could you confirm that you are running `make html` command and checking the plot from where it generates (_build/html/gallery/Contours)?

Plots/Contours/NCL_polar_1_lg.py
```diff
@@ -0,0 +1,115 @@
...
+center, radius = [0.5, 0.5], 0.5
+verts = np.vstack([np.sin(theta), np.cos(theta)]).T
+circle = mpath.Path(verts * radius + center)
+ax.set_boundary(circle, transform=ax.transAxes)
+
+# Manipulate longitude labels (0, 30 E, 60 E, ..., 30 W, etc.)
+ticks = np.arange(0, 210, 30)
+etick = ['0'] + [r'%d$^\circ$E' %
+                 tick for tick in ticks if (tick != 0) & (tick != 180)] + ['180']
+wtick = [
```
clyne, line 76: @erogluorhan do you think we should have @michaelavs turn this code into a geocat-viz convenience function for circular plots? Thanks!

kmpaul: I think that sounds reasonable to me.  Seems like it will be pretty standard, though maybe it makes sense to wait on that decision until we actually see duplication?

erogluorhan: @clyne @kmpalul we should definitely think of adding such a solution for circular plots in **geocat-viz**; however, this code includes pretty much custom (i.e. hard-coded) numeric values in its current version, which might need to be changed for other cases or be calculated in a generic way with respect to plot size etc. For example, two arrays for tick label locations:

```
xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -114, -140]
yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]
```

So, my recommendation is that @michaelavs could try to come up with a generic calculation for `xtick, ytick` locations in this case. If that's not easy on a single case, we may want to study on a few of additional similar plotting examples that need this exact solution then come up with a generic calculation. 

Does that make sense?

michaelavs:
image

Just an update on the line dropout at 150E. This is the graph after running it through Sphinx

Plots/Contours/NCL_polar_1_lg.py
clyne: Sounds good to me.

michaelavs:
@clyne I just created an issue [(#91) ](https://github.com/NCAR/GeoCAT-examples/issues/91). @hCraker I think you will need to assign yourself to it or the function can be pulled over from your code later on, either or.

Issue #64, opened 2020-04-15
Proj_2 (Mercator map projection)


clyne:
Description: An example of a Mercator projection.

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/proj_2.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/proj_2_lg.png

Category: Map projections

PR #76, opened 2020-05-26
Adding NCL_proj_2_lg.py example python file

michaelavs:
Closes issue [#64](https://github.com/NCAR/GeoCAT-examples/issues/64)

image


Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
erogluorhan, line 1: Please revise and format the docstrings accordingly as in the other functions

Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,63 @@
...
+# -*- coding: utf-8 -*-
+"""
+Created on Wed May 13 10:21:41 2020
+
+@author: misi1684
+"""
+
+
+import numpy as np
```
erogluorhan, line 10: Could you please consider separating significant code blocks (e.g. import packages, read data or generate data, plot, etc.) from each other by using special strings such as following:

```
# Import packages:
```

FYI: Whenever you use the above exact amount of number sign (#) in a line, the following code block is separated as a notebook cell by Sphinx in RTD.

Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,63 @@
...
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mticker
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
+from geocat.viz import util as gvutil
+
+ds = xr.open_dataset(gdf.get("netcdf_files/atmos.nc"), decode_times=False)
```
erogluorhan, line 19: Same as above, you can start like this:

```
# Read in data:

# Open a netCDF data file using xarray default engine and load the data into xarrays
```

Plots/Contours/NCL_polar_1_lg.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
```
erogluorhan, line 1: As far as I understand, `Plots/Contours/NCL_polar_1_lg.py` is already being implemented within another PR #75  by the branch **NCL_polar_1_lg.py**. 

Shoul we keep this file in this PR, or otherwise, should you remove it from this PR?

michaelavs: I just saw I have the wrong file uploaded here. Let me check into resolving by uploading the correct file here instead

michaelavs: Actually I see what the problem is, how would I go about removing the other file? This also happens in the proj_3 PR as well but it includes polar_1 and proj_2 so I will have to do the same thing there

erogluorhan: It looks like you were able to delete the wrong file. Great!

Plots/MapProjections/NCL_proj_2_lg.py
michaelavs: Went ahead and based revisions on what we have on the geocat-examples page and have uploaded new version

jukent:
Perhaps this should be inferno as well.

erogluorhan:
Since this PR contains a new folder added to the source code (i.e. **MapProjections**), please include the file **Plots/MapProjections/README.rst** that you included in PR #69 for us to run `make html` command and generate plots appropriately.

Plots/MapProjections/NCL_proj_2_lg.py
erogluorhan: Good!

michaelavs:
@jukent Just updated to the inferno scheme here is the output:
image

michaelavs:
@erogluorhan Just added README.rst to this PR and the proj_3 PR

Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,82 @@
...
+NCL_proj_2_lg.py
+================
+
+This script illustrates the following concepts:
+   - Drawing filled contours over a Mercator map
+   - Setting the spacing for latitude/longitude grid lines
+   - Turning off the map perimeter (boundary)
+   - Making the plot larger using viewport resources
+   - Turning off map fill
+   - Spanning part of a color map for contour fill
```
erogluorhan, line 12: You may want to add a new concept here that is specific to our Python implementation about our effort to introduce "inferno" as a better substitute to "rainbow"

Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,82 @@
...
+
+# Set extent to include latitudes from -90 to 89 and longitudes from -180
+# to 180
+ax.set_extent([180, -180, -90, 89], ccrs.PlateCarree())
+
+# Draw gridlines
+gl = ax.gridlines(crs=ccrs.PlateCarree(), linewidth=1, color='k', alpha=0.5)
+
+# Import an NCL colormap
+newcmp = gvcmaps.gui_default
```
erogluorhan, line 59: This NCL colormap is no longer used in this script; can be removed if we are determined on using "inferno"

michaelavs:
@erogluorhan This is something I had noticed as well, but from what I can tell the entirety of the globe is encompassed. I would assume this indicates I either didn't select the data in its entirety when reading it in or that the data is missing from the set. But if this was true, I figure there would also be the same issue on proj_3 which uses the same data just with a different visualization. I'll do a bit more research on what may be the issue, but if you have any other ideas on the cause of this issue let me know!

michaelavs:
@erogluorhan I figured out the reason the visualization was cutting off part of the globe. The Mercator projection defaults the minimum latitude to -80. The true minimum value for this data is -87.8638 so by bypassing the default, I now have the full projection as is seen in the original. I need to update the lat/lon line spacings and will upload the new projection and code shortly.

erogluorhan:
> @erogluorhan I figured out the reason the visualization was cutting off part of the globe. The Mercator projection defaults the minimum latitude to -80. The true minimum value for this data is -87.8638 so by bypassing the default, I now have the full projection as is seen in the original. I need to update the lat/lon line spacings and will upload the new projection and code shortly.

This sounds great, thanks!

Issue #65, opened 2020-04-15
proj_3 (Orthographic map projection)


clyne:
Description:  An example of a orthographic projection.

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/proj_3.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/proj_3_lg.png

Category: Map projections

PR #78, opened 2020-05-26
Adding NCL_proj_3_lg.py example python file

michaelavs:
Closes issue [#65](https://github.com/NCAR/GeoCAT-examples/issues/65) 

image


michaelavs:
image

Here is the colorblind friendly (viridis scheme) version of this projection

jukent:
Typically we use red-blues for temperature. We want to use a colormap that matches what some quality of the data (viridis is great for ocean data). magma, plasma, or inferno might be more appropriate?

michaelavs:
Here is magma:
image

plasma:
image

and inferno:
image

I feel magma or inferno may be the better of the three in regards to how it may look as a black and white copy, but I am completely open to whatever everyone else thinks would look best!

jukent:
I agree. I think inferno is best!

clyne:
Since this is temperature data, how about trying the built in "coolwarm" colormap?

michaelavs:
@clyne here is how it would look with coolwarm:
image
My one concern would be if the graph were to be black and white as shading may cause the blues to look like the reds. But if we aren't too worried about the black and white aspect of the visualization, I overall like the idea of this color scheme instead.

jukent:
I agree @michaelavs coolwarm has fallen out of favor because it doesn't translate into grayscale. 

Screen Shot 2020-05-27 at 4 48 58 PM

jukent:
Also when using a diverging colormap, such as coolwarm, it is important for the midpoint to be something meaningful, otherwise it lends to misinterpretations of the data. This value could be set at average temperature or the degree of a phase change, for example. So if we are going to use a diverging map we should demonstrate how to set the colorbar midpoint as well using matplotlib.colors.DivergingNorm

michaelavs:
@clyne @jukent I've been looking into using the diverging norm function and the output says that this class has been depreciated with MPL version 3.2 and will be removed in two releases. It recommended using TwoSlopeNorm instead and I've been playing around with that function instead. From what I've seen it looks like it masks the data set and you can plot from there, but the values in the masked array don't make logical sense. All of the values are decimals so I'm going to keep investigating this but if either of you have heard anything about this function/why the array may be decimals, let me know.

jukent:
@michaelavs @clyne That's interesting, especially since the diverging norm wasn't added until 3.1. Before it was added I used my own class, that was something like this:
https://gist.github.com/bdjackson/2215109d2eaa8c68919c

Perhaps this is a good functionality to add to GeoCAT if matplotlib doesn't present a clean and easy solution.

clyne:
A few comments on colormaps:

1. Yes, diverging colormaps don't translate to gray scale well, and they work best when there is some special value to center the "double-ended" color map on (e.g. 0, the mean value, or some other significant value). The gray scale problem is not the problem that is used to be, as most publications now use color figures. You don't see too many grayscale plots in scientific journals these days. For a temperature field, 0 degrees C might be a reasonable value to center the color map over, as would be the mean value. 

2. I'm not sure that for temperature scaling the range of values differently on either side of some point as DivergingNorm or TwoSlopeNorm do is what most people would want, but I really don't know for certain. It's a question to put to the users. You can, however, center the map with uniform scaling by simply establishing the bounds as needed.

3. The MPL "coolwarm" map is pretty widely used by the NASA Scientific Visualization Studio for temperature, as is the "blue orange" diverging map: https://sciviscolor.org/home/colormaps/contrasting-divergent-colormaps/

Several examples of both can be found here:

https://svs.gsfc.nasa.gov/Gallery/WhatsNewwithEarthToday.html

I'm not suggesting that NASA SVS is in anyway an authority, but they produce a lot of high profile material for the public and stakeholders.

4. There is no right answer to the question "what is the best color map to use"?  There are best practices to follow from the field of perceptual psychology and sci vis. There are accepted conventions used by scientists. And there are personal preferences. Some blend of all three may come into play when choosing the "best" palette. Earlier, I suggested a "bake off". We have a number of CGD scientists on zulip, and we have a very active user community on the NCL mailing list. We should take advantage of this to get their input after putting together a few options for them to look at. That said, we'd also like to get this PR accepted, and this particular plot may not be the best choice for getting input from the science community. I would be fine with approving one of the warm=red/cool=blue diverging color maps, or one of the perceptually uniform sequential color maps (e.g. inferno, plasma), and taking up the issue of scientific input as a separate topic outside of this PR. Michaela, as far as I'm concerned you should feel free to keep exploring this issue in the context of this PR if you'd like, or settle on one of the maps you've already presented. In the latter case we may revisit our color map choices later after we have some input from our community.

jukent:
These are all good points @clyne And I agree, we can have this conversation outside of this PR.

Plots/MapProjections/NCL_proj_3_lg.py
```diff
@@ -0,0 +1,79 @@
...
+        "orientation": "horizontal",
+        "ticks": np.linspace(
+            210,
+            310,
+            11),
+        "label": '',
+        "shrink": 0.9})
+
+# Contour-plot data (for borderlines)
+wrap_t.plot.contour(ax=ax, transform=ccrs.PlateCarree(),
```
erogluorhan, line 70: The original NCL plot does not have contour borderlines shown. Is this intentional to show them here?

michaelavs: I hadn't even noticed that! I went ahead and turned them off so to best match the original. Thanks for pointing this out!

Plots/MapProjections/NCL_proj_3_lg.py
```diff
@@ -0,0 +1,80 @@
...
+ax = plt.axes(projection=ccrs.Orthographic(-120, 50))
+
+# Set extent to include latitudes between 0 and 90, and longitude between
+# 0 and -180 only
+ax.set_extent([0, -180, 0, 90], ccrs.PlateCarree())
+ax.set_global()
+ax.coastlines(linewidths=0.5)
+
+# Import an NCL colormap
+newcmp = gvcmaps.gui_default
```
erogluorhan, line 53: This is not used anymore, so it should be removed.

Plots/MapProjections/NCL_proj_3_lg.py
```diff
@@ -0,0 +1,80 @@
...
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/proj_3_lg.png
+"""
+###############################################################################
+# Import packages:
+import numpy as np
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
```
erogluorhan, line 22: I suggested to remove `newcmp = gvcmaps.gui_default` below, so this should be removed as well.

Plots/MapProjections/NCL_proj_3_lg.py
michaelavs: Just updated code to reflect these changes!

Plots/MapProjections/NCL_proj_3_lg.py
```diff
@@ -0,0 +1,71 @@
...
+# 0 and -180 only
+ax.set_extent([0, -180, 0, 90], ccrs.PlateCarree())
+ax.set_global()
+ax.coastlines(linewidths=0.5)
+
+# Contourf-plot data (for filled contours)
+wrap_t.plot.contourf(
+    ax=ax,
+    transform=ccrs.PlateCarree(),
+     levels=11,
```
erogluorhan, line 54: Please check the indentation consistency of these lines (I haven't checked from the source code, but it looks inconsistent here).

Plots/MapProjections/NCL_proj_3_lg.py
```diff
@@ -0,0 +1,71 @@
...
+wrap_t = gvutil.xr_add_cyclic_longitudes(t, "lon")
+
+###############################################################################
+#Plot:
+
+# Generate figure (set its size (width, height) in inches)
+fig = plt.figure(figsize=(10, 10))
+
+# Generate axes using Cartopy and draw coastlines
+ax = plt.axes(projection=ccrs.Orthographic(-120, 50))
```
erogluorhan, line 42: Please add as either argument name or comment that -120 and 50 are `central_longitude` and `central_latitude` values, respectively.

Issue #63, opened 2020-04-15
lcnative_1 (Lambert Conformal projection)


clyne:
Description: Drawing filled contours over a Lambert Conformal map

Script URL:https://www.ncl.ucar.edu/Applications/Scripts/lcnative_1.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/lcnative_1_lg.png

Category: Map projections

PR #81, opened 2020-05-27
Adding NCL_lcnative_1_lg.py example python file

michaelavs:
Closes issue [#63 ](https://github.com/NCAR/GeoCAT-examples/issues/63)

image


michaelavs:
Currently having an issue with using the Lambert conformal projection for this visualization. When projection=ccrs.LambertConformal() in line 25, the code on line 46 (gvutil.add_lat_lon_ticklabels(ax)) throws an error that "This formatter cannot be used with non-rectangular projection". But when switching to Lambert Conformal and removing this line of code, the projection does not look the same as the example, there is a triangular chunk missing from the upper right corner of the visualization. You will also probably notice that the contours are not the same for this projection as in the example. I am not sure if this comes from the data that is being read in, or if there is a discrepancy between the original example and the one I created/used.

Plots/MapProjections/NCL_proj_2_lg.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
erogluorhan, line 1: Please remove this file form this PR since it has its own PR already.

Plots/Contours/NCL_polar_1_lg.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
```
erogluorhan, line 1: Please remove this file form this PR since it has its own PR already.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
erogluorhan, line 1: Please revise and format the docstrings accordingly as in the other functions

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,58 @@
...
+# -*- coding: utf-8 -*-
+"""
+Created on Tue May 19 11:17:13 2020
+
+@author: misi1684
+"""
+
+import numpy as np
```
erogluorhan, line 9: Pease consider separating significant code blocks (e.g. import packages, read data or generate data, plot, etc.) from each other by using special strings as done before.

clyne:
> Currently having an issue with using the Lambert conformal projection for this visualization. When projection=ccrs.LambertConformal() in line 25, the code on line 46 (gvutil.add_lat_lon_ticklabels(ax)) throws an error that "This formatter cannot be used with non-rectangular projection". But when switching to Lambert Conformal and removing this line of code, the projection does not look the same as the example, there is a triangular chunk missing from the upper right corner of the visualization. You will also probably notice that the contours are not the same for this projection as in the example. I am not sure if this comes from the data that is being read in, or if there is a discrepancy between the original example and the one I created/used.

Have you configured the Cartopy parameters for the Lambert Conformal projection to match those used by NCL's LC projection (i.e. the Meridian and parallels)? See 

https://www.ncl.ucar.edu/Document/Graphics/Resources/mp.shtml

https://scitools.org.uk/cartopy/docs/latest/crs/projections.html#lambertconformal

michaelavs:
@clyne just looked into it and added the central_longitude and standard_parallels that are listed in the ncl file but with no success. I then tried to again remove line 48 which allowed this visualization to be produced:
image
and I am not sure why it is one color and not reporting the data correctly

clyne:
Looks like it's going to take some more digging :-). I'd suggest ignoring the data for now and just trying to get the map projection correct (or vise versa)

michaelavs:
here is the conformal projection with gridlines turned on (so we can see where the lat/lon lines are located) 
image
I will also be pushing the code for this shortly

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,80 @@
...
+ax.coastlines(linewidth=0.5)
+
+ax.gridlines()
+
+ax.plot(np.arange(30,55,5))
+
+
+# Plot data and create colorbar
+newcmp = gvcmaps.BlueYellowRed
+t.plot.contourf(ax=ax, cmap=newcmp, transform=ccrs.PlateCarree(), levels = 14, cbar_kwargs={"orientation":"horizontal",  "ticks":np.arange(0, 240, 20),  "label":'', "shrink":0.9})
```
clairefio, line 57: I would change the "shrink" argument to make the colorbar line up with the map.
```suggestion
t.plot.contourf(ax=ax, cmap=newcmp, transform=ccrs.PlateCarree(), levels = 14, cbar_kwargs={"orientation":"horizontal",  "ticks":np.arange(0, 240, 20),  "label":'', "shrink":0.7})
```

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,80 @@
...
+
+# Use geocat.viz.util convenience function to add minor and major tick lines
+#Use geocat.viz.util convenience function to set axes tick values
+#gvutil.set_axes_limits_and_ticks(ax, xticks=np.linspace(30, 55, 6), yticks=np.linspace(20, 45, 6))
```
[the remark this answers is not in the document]
michaelavs: When we use the Lambert conformal projection, the convenience tools don't like that the ticks are non-rectangular so if they are active in the code, we get an error in the code. This is most likely going to have to be addressed in the convenience functions to allow for the projection to have ticks added to non-rectangular projections. The best way to visualize the problem the function is running into is in the last projection I uploaded vs the original NCL projection. In the NCL projection, the image is "flat" so the ticks going around the outside border make sense to be read in as such, however, in the python version the gridlines show where the lat/lon lines should be and they are not "flat" which I think is why the convenience function is getting angry.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+import numpy as np
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
+from geocat.viz import util as gvutil
+import matplotlib.ticker as ticker
```
hCraker, line 23: I would move the ticker import next to the plt import. We should group import statements from the same package together and put the geocat imports at the end for clarity. That way, the imports are organized by common packages (matplotlib.pyploy, cartopy) at the top and less common (geocat.util) at the bottom.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,80 @@
...
+
+###############################################################################
+# Read in data:
+
+# Open a netCDF data file using xarray default engine and load the data into xarrays
+ds = xr.open_dataset(gdf.get("netcdf_files/pre.8912.mon.nc"))
+
+# Extract a slice of the data
+t = ds.pre[0,:]
+#print(t)
```
hCraker, line 33:
```suggestion

```
Unnecessary print statement

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,80 @@
...
+#gvutil.xr_add_cyclic_longitudes(t,'lon')
```
hCraker, line 34: Is this still needed or can it be deleted?

Plots/MapProjections/NCL_lcnative_1_lg.py
hCraker, line 57: Here's a comment from the NCL file:
>; Usually, when data is placed onto a map, it is TRANSFORMED to the specified
; projection. Since this model is already on a native lambert conformal grid,
; we want to turn OFF the transformation.

I think the reason why your image isn't "flat" is because you are transforming the data from one projection to another. This data is native to a Lambert Conformal grid so `transform=ccrs.PlateCarree()` isn't needed. Give it a try with out the transformation and you may be able to add the lat/lon ticks.

hCraker:
> image
> I was able to find this method to add the lat/lon lines. I haven't quite figured out how to get the the longitude lines to go from being in the middle of the projection to being on the x axis instead. Also not too sure why it presents two sets of numbers on the projection. @clairefio @hCraker do either of you have any experience with the xlines, xlocator, or xlabel functions?

I have only used those functions on rectangular maps. I suggest looking at the `Gridliner` function in `cartopy`. It has options to draw labels along with the latitude and longitude lines.
Here's the link https://scitools.org.uk/cartopy/docs/v0.13/matplotlib/gridliner.html

michaelavs:
I've been doing some more trial and error attempts with this projection and just got this error: "TypeError: Cannot label gridlines on a LambertConformal plot.  Only PlateCarree and Mercator plots are currently supported." With a quick google search, it appears this was questioned in a [SciTools Github ](https://github.com/SciTools/cartopy/issues/1323) and may have been resolved in the most [recent update](https://github.com/SciTools/cartopy/releases/tag/v0.18.0) of cartopy which was release on May 3, 2020. It may not be an immediate thing to address, but I would be curious if updating our Cartopy in the geocat examples environment may be able to address the problem with adding lat/lon lines to Lambert conformal projections. That being said, will I need to do anything in particular to proceed with this version update on my side @erogluorhan? I just tried conda update cartopy and it shows as up to date, but conda list shows as version 0.17. Additionally, if this does end up solving the issue, how would we go about updating the environment for all users?

erogluorhan:
> I've been doing some more trial and error attempts with this projection and just got this error: "TypeError: Cannot label gridlines on a LambertConformal plot. Only PlateCarree and Mercator plots are currently supported." With a quick google search, it appears this was questioned in a [SciTools Github ](https://github.com/SciTools/cartopy/issues/1323) and may have been resolved in the most [recent update](https://github.com/SciTools/cartopy/releases/tag/v0.18.0) of cartopy which was release on May 3, 2020. 

I agree that version 0.18.0 would likely have resolved this issue, reading major changes under release notes of this version. It'd be good to give this version a try then.

> I just tried conda update cartopy and it shows as up to date, but conda list shows as version 0.17. 

Please try `conda update --all` as well. I checked my cartopy version and it is 0.18.0, and I think it was handled by recently calling `conda update --all`. FYI: The possible reasons of why conda cannot update to the actual latest version of a package by simply calling `conda update <package_name>`  but `conda update --all `can simply work is discussed here: https://github.com/conda/conda/issues/786

> Additionally, if this does end up solving the issue, how would we go about updating the environment for all users?

There is nothing to do on your side to update others' environments. Other users of the **geocat-examples** will get errors with the cartopy v0.18.0 function you would use, then they will handle their own environment update.

michaelavs:
> > I've been doing some more trial and error attempts with this projection and just got this error: "TypeError: Cannot label gridlines on a LambertConformal plot. Only PlateCarree and Mercator plots are currently supported." With a quick google search, it appears this was questioned in a [SciTools Github ](https://github.com/SciTools/cartopy/issues/1323) and may have been resolved in the most [recent update](https://github.com/SciTools/cartopy/releases/tag/v0.18.0) of cartopy which was release on May 3, 2020.
> 
> I agree that version 0.18.0 would likely have resolved this issue, reading major changes under release notes of this version. It'd be good to give this version a try then.
> 
> > I just tried conda update cartopy and it shows as up to date, but conda list shows as version 0.17.
> 
> Please try `conda update --all` as well. I checked my cartopy version and it is 0.18.0, and I think it was handled by recently calling `conda update --all`. FYI: The possible reasons of why conda cannot update to the actual latest version of a package by simply calling `conda update <package_name>` but `conda update --all `can simply work is discussed here: [conda/conda#786](https://github.com/conda/conda/issues/786)
> 
> > Additionally, if this does end up solving the issue, how would we go about updating the environment for all users?
> 
> There is nothing to do on your side to update others' environments. Other users of the **geocat-examples** will get errors with the cartopy v0.18.0 function you would use, then they will handle their own environment update.

I just tried doing  `conda update --all` in both my base environment and geocat-examples and cartopy is still at version 0.17. I have also tried using `conda update --all -c ncar` just from the update we did for GeoCAT packages today and it also kept cartopy at 0.17.

erogluorhan:
> I just tried doing `conda update --all` in both my base environment and geocat-examples and cartopy is still at version 0.17. I have also tried using `conda update --all -c ncar` just from the update we did for GeoCAT packages today and it also kept cartopy at 0.17.

Try this: `conda install cartopy=0.18.0`

michaelavs:
> > I just tried doing `conda update --all` in both my base environment and geocat-examples and cartopy is still at version 0.17. I have also tried using `conda update --all -c ncar` just from the update we did for GeoCAT packages today and it also kept cartopy at 0.17.
> 
> Try this: `conda install cartopy=0.18.0`

I got this error: 
`PackagesNotFoundError: The following packages are not available from current channels:

  - cartopy=0.18.0

Current channels:

  - https://repo.anaconda.com/pkgs/main/osx-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/osx-64
  - https://repo.anaconda.com/pkgs/r/noarch`

clyne:
FWIW I just checked and I have 0.18.0 on my system.

erogluorhan:
> FWIW I just checked and I have 0.18.0 on my system.

Thanks! We later figured out that conda-forge was not added into conda channels, so we managed to upgrade to 0.18.0 by enforcing conda-forge.

@michaelavs you may want to add conda-forge to your conda channels for future convenince as follows:

`conda config --add channels conda-forge`

michaelavs:
Was able to turn on gridlines using cartopy 0.18, here is the new version: 
image
I am now looking into if there is a way to be more specific about which labels are present because there is overlap between come of the lat/lon lines and their location on the visualization

michaelavs:
I've been looking into this issue some more and talked with John about this idea yesterday, but wanted some more input from everyone else before making a decisions on direction for this projection. When using NCL, the Lambert conformal projection has no curvature to the latitude lines, which it should have because a conformal projection is conic in shape. From looking more into the documentation of the code, it appears there is a function that essentially pins the projection to 4 corners on lat/lon lines and will project as a rectangle instead of cone. The line that does this is `res@mpLimitMode = "Corners" ` which is not turned on in the lcmask projection which Rose has worked on, which is why that projection is rounded while this one is rectangular (in NCL). All of that being said, Cartopy has three separate projection options for Lambert: Conformal, Cylindrical (rectangular output), and AzimuthalEqualArea (rounded with equal area). The cylindrical projection allows to almost perfectly recreate the NCL projection and has no issues with adding gridlines or using convenience functions, but conformal is technically the version we should be using to remain true to NCL. However, John and I discussed the potential of including all three of these Lambert projections to showcase the true functionality of Cartopy and even discuss how there are differences between NCL and Python when it comes to certain projections and therefore the importance of choosing the correct projection for the data. What is everyones thoughts on slightly straying from the norm and including extra info/plots for this example? I am also including the three separate outputs so you can all see them as well @clairefio @hCraker @jukent @kmpaul 
image

image

image

kmpaul:
@michaelavs: It seems to me that you might have answered your own question.  You said that a "Conformal" plot should be conical, but that is *not* what NCL is displaying.  So, by that definition, the NCL plot is either *wrong* or, at least, *misleading*.  If you decide that the NCL plot is actually *not* displaying what it suggests it is displaying, then I think the best coarse of action is to display the correct thing...  and perhaps showcase what all Cartopy can do.  So, I think the "all Lambert options" solution is the way to go.

michaelavs:
Closing this PR to switch to [newer](https://github.com/NCAR/GeoCAT-examples/pull/141) version to fix issues with pushing changes

Issue #79, opened 2020-05-26
conwomap_3


erogluorhan:
Description: Contouring data without a map

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/conwomap_3.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/conwomap_3_lg.png

PR #84, opened 2020-05-28
Conwomap 3

clairefio:
Issue #79 
Most recent plot:
conwomap_3 plot


clairefio:
@NCAR/geocat is it okay if I use the data file used in the other NCL conwomap plots instead of using random data like they used in the NCL version of this specific plot?

erogluorhan:
> @NCAR/geocat is it okay if I use the data file used in the other NCL conwomap plots instead of using random data like they used in the NCL version of this specific plot?

I don't have strong preference but would be nice to see a `random_normal()` implementation. Is there a specific reason for using netCDF data instead?

erogluorhan:
I would like to discuss an observation on this and some other PRs, too:

Newer PRs (such as this one) seem to include "**Files Changed**" from previously created PRs as well, which should not be the case. For example, this PR also includes the files **"Plots/Bar/NCL_unique_5.py"** and **"Plots/Trajectories/NCL_traj_1.py"** as "**Files Changed**" (i.e. added to repository), which indeed have been dealt with by their own PRs #74 and #82, respectively.

Best practice should be that every single PR should only contain the target files of its own, not other files from other PRs. 

If this makes sense, we can also discuss what might be the cause and how we can avoid.

cc: @michaelavs @hCraker

clairefio:
> I would like to discuss an observation on this and some other PRs, too:
> 
> Newer PRs (such as this one) seem to include "**Files Changed**" from previously created PRs as well, which should not be the case. For example, this PR also includes the files **"Plots/Bar/NCL_unique_5.py"** and **"Plots/Trajectories/NCL_traj_1.py"** as "**Files Changed**" (i.e. added to repository), which indeed have been dealt with by their own PRs #74 and #82, respectively.
> 
> Best practice should be that every single PR should only contain the target files of its own, not other files from other PRs.
> 
> If this makes sense, we can also discuss what might be the cause and how we can avoid.
> 
> cc: @michaelavs @hCraker

I think this might be happening because I'm not branching from the master branch (ex. creating branch unique_5 while in traj_1 instead of master). Does that sound like it would be causing this problem?

clyne:
> @NCAR/geocat is it okay if I use the data file used in the other NCL conwomap plots instead of using random data like they used in the NCL version of this specific plot?

Does numpy.random.normal() not produce a comparable data set to NCL's random_normal?

erogluorhan:
> I think this might be happening because I'm not branching from the master branch (ex. creating branch unique_5 while in traj_1 instead of master). Does that sound like it would be causing this problem?

Yes, it seems to be the case. Indeed, it is not a problem, and even it might be useful for some specific cases, but for PRs here, it could be better to branch from master branch and so have each PR have its own files changed only.

cc: @michaelavs @hCraker

clairefio:
Updated plot:
conwomap_3 graph

clairefio:
Updated plot:
conwomap_3 graph

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,92 @@
...
+# Import packages:
+import numpy as np
+import matplotlib.pyplot as plt
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Generate random data:
+
+xlist = np.linspace(0, 30.0, 30)
```
clyne, line 31: The NCL version uses a 31x31 array. Here we're using 30x30. We should be consistent to avoid confusion.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,92 @@
...
+ylist = np.linspace(0, 30.0, 30)
+xdata, ydata = np.meshgrid(xlist, ylist)
+
+zdata = np.random.normal(0, 2, size=(30, 30))
```
clyne, line 35: Ditto, the second parameter (variance?)  for random.normal is 2, in the NCL version it is 3.0. They should be consistent. If there is a reason that they need to be different it should be documented.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,92 @@
...
+gvutil.add_major_minor_ticks(ax, x_minor_per_major=4, y_minor_per_major=4)
+
+# Use geocat.viz.util convenience function to add titles to left and right of the plot axis.
+gvutil.set_titles_and_labels(ax, ylabel="wave number", labelfontsize=24)
+
+# Set ticks and labels only on left and top of plot
+ax.xaxis.tick_top()
+ax.yaxis.tick_left()
+
+# Make tick font size bigger
```
erogluorhan, line 57: Instead of these **Lines 57** through **59**, I would add an argument `labelsize=18` to the above function call `gvutil.add_major_minor_ticks()`, which handles the same task.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,92 @@
...
+plt.xticks(fontsize=18)
+plt.yticks(fontsize=18)
```
erogluorhan, line 59: Instead of these **Lines 57** through **59**, I would add an argument `labelsize=18` to the above function call `gvutil.add_major_minor_ticks()`, which handles the same task.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,92 @@
...
+plt.figure(figsize=(10, 10))
+
+# Create axes
+ax = plt.axes()
+
+# Use geocat.viz.util convenience function to set axes limits & tick values without calling several matplotlib functions
+gvutil.set_axes_limits_and_ticks(ax, xlim=(0, 30), ylim=(0, 30), xticks=None, yticks=None, xticklabels=None, yticklabels=None)
+
+# Use geocat.viz.util to add major and minor tics
+gvutil.add_major_minor_ticks(ax, x_minor_per_major=4, y_minor_per_major=4)
```
erogluorhan, line 48: Sorry, I noticed after my approval of the PR that `x_minor_per_major` and `y_minor_per_major` should be set equal to **5** instead of **4** as in the NCL case. Please apply this with a commit.

clairefio:
Final plot:

conwomap final

erogluorhan:
@clairefio , I now notice that there is a Github warning at the final line of the code, which is "No newline at end of file". It's not an error, but for the best style, you may want to add a newline after your actual final line of code.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,88 @@
+"""
+NCL_conwomap_3.py
+=================
+Concepts illustrated:
+  - Drawing a simple contour plot
+  - Generating dummy data using "random_normal"
+  - Masking mirrored contour data
+  - Drawing a perimeter around areas on a contour plot with missing data
+  - Turning off the bottom and right borders of a contour plot
+  - Using "getvalues" to retrieve resource values
```
clyne, line 10: It looks like some of these comments apply to the original NCL script, but not the Python version. For example, I think "getvalues" is an NCL construct. Could you review the comments and remove the ones that aren't relevant?

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Set ticks and labels only on left and top of plot
+ax.xaxis.tick_top()
+ax.yaxis.tick_left()
+
+# Set color of right and bottom axes to make them invisible
+ax.spines['right'].set_color('white')
+ax.spines['bottom'].set_color('white')
+
+# Create a numpy array of 30
```
clyne, line 61: This comment doesn't agree with what the code actually does (array length is 31, not 30). Also, rather than replicating the literal 31, it would be better practice to size the array 'x' using  the size of xlist.

Plots/Contours/NCL_conwomap_3.py
```diff
@@ -0,0 +1,88 @@
...
+plt.step(x, x, color='black', zorder=7)
+
+# Plot contour data
+cp = ax.contour(xdata, ydata, zdata, colors='k', linewidths=1.0)
+
+# Label contours
+ax.clabel(cp, inline=True, fontsize=10, colors='k', fmt="%.0f")
+
+# Ignore second half of the graph
+y1 = np.full(shape=31, fill_value=0, dtype=np.int)
```
clyne, line 74: Ditto here for array size.

clyne:
Can we merge this at your convenience, @erogluorhan

Issue #77, opened 2020-05-26
polyg_2 (Color coded Lambert Conformal map)


clairefio:
Description: An example of a color coded lambert conformal map

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/polyg_2.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/polyg_2_lg.png

Category: Polygons

PR #85, opened 2020-05-29
Polyg 2

clairefio:
Issue #77:

Current plot:

polyg_2 graph
color bar


clairefio:
Updated plot:
polyg_2 graph

clairefio:
The data file I used for this (climdiv_polygons.nc) does not contain actual data on precipitation levels for each climate division, only the climate division locations/borders. I just set each division to a random color from the and tried to emulate the colors from the NCL graph colorbar. @NCAR/geocat

erogluorhan:
> The data file I used for this (climdiv_polygons.nc) does not contain actual data on precipitation levels for each climate division, only the climate division locations/borders. I just set each division to a random color from the and tried to emulate the colors from the NCL graph colorbar. @NCAR/geocat

To understand correctly, was there a specific reason for not using "climdiv_prcp_1899-1999.nc", which is the actual data file the NCL original plot uses?

clairefio:
> > The data file I used for this (climdiv_polygons.nc) does not contain actual data on precipitation levels for each climate division, only the climate division locations/borders. I just set each division to a random color from the and tried to emulate the colors from the NCL graph colorbar. @NCAR/geocat
> 
> To understand correctly, was there a specific reason for not using "climdiv_prcp_1899-1999.nc", which is the actual data file the NCL original plot uses (May be it was too large)?

I couldn't find it in GeoCAT-datafiles, do you know where it might be?

erogluorhan:
> > > The data file I used for this (climdiv_polygons.nc) does not contain actual data on precipitation levels for each climate division, only the climate division locations/borders. I just set each division to a random color from the and tried to emulate the colors from the NCL graph colorbar. @NCAR/geocat
> > 
> > 
> > To understand correctly, was there a specific reason for not using "climdiv_prcp_1899-1999.nc", which is the actual data file the NCL original plot uses (May be it was too large)?
> 
> I couldn't find it in GeoCAT-datafiles, do you know where it might be?

I've found it and will make it available at geocat-datafiles soon.

erogluorhan:
@clairefio , you should be now able to access the actual data file on **geocat-datafiles**.

clairefio:
> @clairefio , you should be now able to access the actual data file on **geocat-datafiles**.

Thank you!

clairefio:
Updated plot with real precipitation data:

polyg_2 graph

erogluorhan:
This looks awesome!

The only thing I can notice is that NCL plot has no colorbar levels for 40-45 and 45- 40; instead, it has 40-50 as combined. This makes a slight shift of colorscale on locations where precipitation is higher than 45.  Once this is resolved, I will review line-by-line.

clairefio:
Updated plot:
polyg_2 graph

clairefio:
> This looks awesome!
> 
> The only thing I can notice is that NCL plot has no colorbar levels for 40-45 and 45- 40; instead, it has 40-50 as combined. This makes a slight shift of colorscale on locations where precipitation is higher than 45. Once this is resolved, I will review line-by-line.

Thanks! Fixed the color scale increments.

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,155 @@
...
+        first = ds.get(varname)
+        precipitationdata = sum(first.values)/100
+
+        # Get borders of each climate division
+        lat = first.lat
+        lon = first.lon
+
+        # Determine which color to fill the divisions based on precipitation data
+        colorindex = 0
+        if (precipitationdata<5):
```
clyne, line 91: Rather than having a big, hard-coded if/else statement, how about writing a small helper function that take a 'bounds' array and a value (precipitation data), and returns a color index. It will make the code more maintainable and extendable if the mapping ever needs to be changed. Also you won't have all of the special bounds values duplicated (once in the loop, and once below in the 'bounds' array)

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,155 @@
...
+    facecolor = 'white'
+    edgecolor = 'black'
+
+    ax.add_geometries([state], ccrs.PlateCarree(), facecolor=facecolor, edgecolor=edgecolor)
+
+# Open climate division datafile and add to xarray
+ds = xr.open_dataset(gdf.get("netcdf_files/climdiv_prcp_1899-1999.nc"), decode_times=False)
+
+# For each variable (climate division) in data set, create outline on map and fill with random color
+for varname, da in ds.data_vars.items():
```
clyne, line 78: This loop doesn't use 'da'. It looks like you're using the 'first' variable below instead of 'da'

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,155 @@
...
+        elif (precipitationdata>=60 and precipitationdata<70):
+            colorindex = 11
+        elif (precipitationdata>=70 and precipitationdata<80):
+            colorindex = 12
+        elif (precipitationdata>=80 and precipitationdata<90):
+            colorindex = 13
+        elif (precipitationdata>=90 and precipitationdata<100):
+            colorindex = 14           
+
+        track = sgeom.LineString(zip(lon, lat))
```
clyne, line 120: What does this statement do? Can you document it?

Plots/Trajectories/NCL_traj_1.py
```diff
@@ -0,0 +1,86 @@
+"""
```
clyne, line 1: Why is this file, NCL_traj_1.py, being included in this PR? Are you branching from master, or from another branch? Should be the former.

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Open climate division datafile and add to xarray
+ds = xr.open_dataset(gdf.get("netcdf_files/climdiv_prcp_1899-1999.nc"), decode_times=False)
+
+# For each variable (climate division) in data set, create outline on map and fill with random color
+for varname, da in ds.data_vars.items():
+
+    try:
+        # Get precipitation data for each climate division
+        first = da
```
clyne, line 85: Why assign 'da' to 'first'? It's not incorrect code, but it makes the reader wonder why you simply didn't use 'first' in the for loop assignment above, or just us 'da' in the loop body?

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,140 @@
...
+        precipitationdata = sum(first.values)/100
+
+        # Get borders of each climate division
+        lat = first.lat
+        lon = first.lon
+
+        # Set color map with bounds
+        colormap = [(5, 'mediumpurple'), (10, 'mediumblue'), (15, 'royalblue'),
```
clyne, line 93: Now we have the list of colors duplicated (here and after the for loop when you generate your ListedColormap), and we have the mapping interval  (5, 10, 15, etc.) duplicated here and when you create the 'bounds' loop. How can you restructure this code so that you can still have a help function for mapping a color to an index, but avoid the duplication of values/colors? The problem with duplication like this is that it leads to less maintainable code; someone, for example, will change the 10 in colormap, but forget to change it in bounds.

Plots/Polygons/NCL_polyg_2.py
clairefio: you're right! Not sure why I did that... but it's fixed.

Plots/Polygons/NCL_polyg_2.py
clairefio: Fixed it! Now there is only one colormap dictionary defined, and the other two temporary arrays draw their color/bound values from it.

clyne: Hmm. Moving in the right direction, but...

Now we have two colormaps, and two sets of bounds, where one of each would easily do. Rather than creating a custom, dictionary colormap that will be unfamiliar to readers we can make do with the single matplotlib.colors class to store only our colors, and a single array to store our bounds. All we need to do is generate the matplotlib.colors instance from the list of colors, create a bounds array, and take advantage of the fact that ListedColormaps store their colors - the color names that you've provided - in the .colors attribute. Hence, you can access the color names in the 'cmap' variable with:

`cmap.colors[i]`

where 'i' is the color index. Then you can make a simple change to findDivColor so that it operates on the 'bounds' array and returns an index that you can then use to retrieve the desired color. Make sense? The point of all this is to avoid having to create separate copies of the colormap and bounds information in different representations when one will easily do

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,132 @@
...
+colormap = colors.ListedColormap(['mediumpurple', 'mediumblue', 'royalblue', 'cornflowerblue', 'lightblue', 'teal', 'yellowgreen',
+                                  'green', 'wheat', 'tan', 'gold', 'orange', 'red', 'firebrick'])
+
+colorbounds = [0, 5, 10, 15, 20, 25, 30, 35, 40, 50, 60, 70, 80, 90, 100]
+
+###############################################################################
+# Define helper function to determine which color to fill the divisions based on precipitation data
+
+
+def findDivColor(colorbounds, pdata):
```
clyne, line 49: What gets returned by this function if the 'else' clause is never executed (if pdata is outside of range of colorbounds)?

Plots/Polygons/NCL_polyg_2.py
clyne: Nice!

Plots/Polygons/NCL_polyg_2.py
clyne: Looks good!

clyne:
Just need to try to figure out why the two regions in the north west (washington and oregon) are getting mapped to different colors than the NCL version.

Plots/Polygons/NCL_polyg_2.py
clairefio: @clyne I added a case where if it's bigger than the last color bound listed, it still returns the last color in the colorbar. Is this ok?

clairefio: > Just need to try to figure out why the two regions in the north west (washington and oregon) are getting mapped to different colors than the NCL version.

I am still confused about the colors of the original NCL plot... I thought I was making a rounding error, but now I'm not sure that is the case. 

In the original plot the 4th climate division in Oregon (the uppermost leftmost division) is pink even though the precipitation data is 93.402, which falls within the 90-100 range (which is supposed to be dark red). 

Additionally, the rounding for the NCL bounds doesn't seem to be consistent. For example, for the 10th division in Texas, the precipitation value is 25.3749 and the color shows it is in the 25-30 range. However, for the 4th division in Oregon, the precipitation value is 70.389 and the color shows it is in the 60-70 range. Not sure how they decided which divisions to round up/round down.

clyne: Are you sure your script is calculating the precipitation data the same way the NCL script is?  They look very different to me.

clairefio: > Are you sure your script is calculating the precipitation data the same way the NCL script is? They look very different to me.

Yes I was! I was getting the average annual precipitation by dividing the data value by 100 instead of 101. I think it's fixed now. 

Here's how the plot looks now:

polyg_2 graph

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,136 @@
...
+import shapely.geometry as sgeom
+from mpl_toolkits.axes_grid1.inset_locator import inset_axes
+
+###############################################################################
+# Initialize color map and bounds for each color
+
+colormap = colors.ListedColormap(['mediumpurple', 'mediumblue', 'royalblue', 'cornflowerblue', 'lightblue', 'lightseagreen', 'yellowgreen',
+                                  'green', 'wheat', 'tan', 'gold', 'orange', 'red', 'firebrick'])
+
+colorbounds = [0, 5, 10, 15, 20, 25, 30, 35, 40, 50, 60, 70, 80, 90, 100]
```
clyne, line 43: Please comment that this represents inches of rain (or whatever)

Plots/Polygons/NCL_polyg_2.py
clyne: Nice!

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,136 @@
...
+
+# Open climate division datafile and add to xarray
+ds = xr.open_dataset(gdf.get("netcdf_files/climdiv_prcp_1899-1999.nc"), decode_times=False)
+
+# For each variable (climate division) in data set, create outline on map and fill with random color
+for varname, da in ds.data_vars.items():
+
+    try:
+        # Get precipitation data for each climate division
+        precipitationdata = sum(da.values)/101
```
clyne, line 98: Can we get the number of years from the data? Hardcoding 101 is probably not what we want

clyne, line 98: The calculation of precipitation data *looks* very different than what is done in the NCL script (it's much simpler :-). Can you add a comment explaining why it's so much simpler in Python?

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,141 @@
...
+
+    ax.add_geometries([state], ccrs.PlateCarree(), facecolor=facecolor, edgecolor=edgecolor)
+
+# Open climate division datafile and add to xarray
+ds = xr.open_dataset(gdf.get("netcdf_files/climdiv_prcp_1899-1999.nc"), decode_times=False)
+
+# For each variable (climate division) in data set, create outline on map and fill with random color
+for varname, da in ds.data_vars.items():
+
+    try:
```
clyne, line 97: Can you comment why this code should go inside a try-catch block?

clairefio: I took out the try-catch and just added an if statement instead- it was only there so the loop would ignore the first xarray item, which doesn't contain any climate division data.

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,138 @@
...
+colormap = colors.ListedColormap(['mediumpurple', 'mediumblue', 'royalblue', 'cornflowerblue', 'lightblue', 'lightseagreen', 'yellowgreen',
+                                  'green', 'wheat', 'tan', 'gold', 'orange', 'red', 'firebrick'])
+
+# Values represent average number of inches of rain
+colorbounds = [0, 5, 10, 15, 20, 25, 30, 35, 40, 50, 60, 70, 80, 90, 100]
+
+###############################################################################
+# Define helper function to determine which color to fill the divisions based on precipitation data
+
+
```
erogluorhan, line 49: May remove one of these two line breaks

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,138 @@
...
+# Add outlines of each state within the United States
+for state in shpreader.Reader(states_shp).geometries():
+
+    facecolor = 'white'
+    edgecolor = 'black'
+
+    ax.add_geometries([state], ccrs.PlateCarree(), facecolor=facecolor, edgecolor=edgecolor)
+
+# Open climate division datafile and add to xarray
+ds = xr.open_dataset(gdf.get("netcdf_files/climdiv_prcp_1899-1999.nc"), decode_times=False)
```
erogluorhan, line 92: How about moving this data reading block to a place just after import packages block so that we can have a code flow like:

import packages -> read data -> helper functions -> plot

Plots/Polygons/NCL_polyg_2.py
```diff
@@ -0,0 +1,140 @@
...
+  - Creating a red-yellow-blue color map
+
+See following URLs to see the reproduced NCL plot & script:
+    - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/polyg_2.ncl
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/polyg_2_lg.png
+"""
+
+###############################################################################
+# Import packages:
+# ----------------
```
erogluorhan, line 24: Please remove this line to keep this heading consistent with other cell headings throughout the html page.

PR #96, opened 2020-06-04
Created xy_3 example

### LINKED ISSUE #95: xy_3
Description: An XY plot with the y-axis inverted and dashed lines

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/xy_3.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/xy_3_1_lg.png
https://www.ncl.ucar.edu/Applications/Images/xy_3_2_lg.png

Category: XY Plots

hCraker:
Closes #95 
image
image

### DIFF
--- a/Plots/XY/NCL_xy_3.py
+++ b/Plots/XY/NCL_xy_3.py
@@ -51,3 +51,28 @@ gvutil.set_axes_limits_and_ticks(ax, ylim=(1000, 0),
 gvutil.set_titles_and_labels(ax, maintitle="Profile Plot", xlabel=ds.long_name, ylabel=ds['lev'].long_name)
 
 plt.show()
+
+###############################################################################
+# Plot:
+
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+ax = plt.gca()
+
+# Plot data with custom line characterisitcs
+# Use keyword `color` to change the line color
+# Use keyword `linewidth` to change the line thickness
+# Use keyword `dashes` to create a custom dash pattern
+plt.plot(ds.data, ds.lev, color='red', linewidth=3, dashes=[3, 1, 1, 1, 1, 1, 1, 1])
+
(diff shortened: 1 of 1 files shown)


Plots/XY/NCL_xy_3.py
```diff
@@ -0,0 +1,78 @@
...
+
+###############################################################################
+# Plot:
+
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+ax = plt.gca()
+
+# Plot data
+plt.plot(ds.data, ds.lev)
```
michaelavs, line 41: Can you define line color here as you did in line 66?

hCraker: I refrained from doing that because the NCL example first plots a line with the default characteristics. NCL and matplotlib have different defaults (color and linestyle being two of them) so I wanted to showcase matplotlib's defaults of a solid, blue line here. Do you think that's appropriate for this example?

michaelavs: I think that's fair. I wanted to make sure this was addressed in case it was an issue within any of the functions and would need to be fixed

jukent:
I think this looks good. And I agree it makes sense to have plot 1 match the default settings rather than the NCL appearance. Are there any other keywords that can be used to manipulate a line plot in Python that we could also demonstrate?

hCraker:
> I think this looks good. And I agree it makes sense to have plot 1 match the default settings rather than the NCL appearance. Are there any other keywords that can be used to manipulate a line plot in Python that we could also demonstrate?
@jukent 
I added dash_capstyle to make the dashes rounded. I could add point markers, but with the dashes things get too cluttered. Point markers come up in a few NCL examples we haven't made in python yet (i.e. xy_37) so I think it might be best to leave the markers out of this example. Is there anything else you think I should add?

jukent:
I think that's good! Nice work.

PR #98, opened 2020-06-05
Comparisons

michaelavs:
Creating comparison plots to show how to implement color blind appropriate color maps


michaelavs:
Color map comparisons with rain data
image

Plots/ColormapComparisons/CB_Rain.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Thu Jun  4 14:12:34 2020
+
```
hCraker, line 5: Don't forget to add a description of what this file shows like you did for CB temperature.

Plots/ColormapComparisons/CB_Rain.py
```diff
@@ -0,0 +1,136 @@
...
+
+import numpy as np
+import xarray as xr
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
+from geocat.viz import util as gvutil
+import cartopy.feature as cfeature
```
hCraker, line 17: I would group the imports from cartopy together. It's a small thing, but it a bit more organized that way so people reading the example know what all they need to import from cartopy.

Plots/ColormapComparisons/CB_Rain.py
```diff
@@ -0,0 +1,136 @@
...
+###############################################################################
+# Read in data:
+
+# Open a netCDF data file using xarray default engine and load the data into xarrays
+ds = xr.open_dataset(gdf.get("netcdf_files/pre.8912.mon.nc"))
+
+# Extract a slice of the data
+t = ds.pre[0,:]
+
+#gvutil.xr_add_cyclic_longitudes(t,'lon')
```
hCraker, line 27: You commented this out. Does it still need to be here, or can it be removed? If it needs to remain, add a comment explaining what happens with and without this line.

Plots/ColormapComparisons/CB_Temperature.py
```diff
@@ -0,0 +1,145 @@
...
+
+import numpy as np
+import xarray as xr
+import cartopy.feature as cfeature
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
+from geocat.viz import util as gvutil
+
```
hCraker, line 30: I think this file could benefit from being broken up into chucks using that lines of hashes. That way Sphinx will show code blocks versus one long file.

Plots/ColormapComparisons/CB_Rain.py
michaelavs: Just added updates you had mentioned, let me know what you think. Definitely open to feedback/revisions on the description of the plots, it was definitely hard to summarize the goal of creating the plots in only a couple of lines so if any info seems redundant or could be updated let me know!

Plots/ColormapComparisons/CB_Rain.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+CB_Rain.py
+================
+
+This script illustrates multiple color schemes for color maps which will allow for those
+impacted by color blindness to see visualizations. Using rainbow color schemes not only
+cause confusion for those impacted by color blindness, but is a poor choice in color scheme
```
hCraker, line 9:
```suggestion
 can cause confusion for those impacted by color blindness, but is a poor choice 
```
I think this sounds a bit better. "in color sheme" is implied and not needed if you're more focused on brevity.

Plots/ColormapComparisons/CB_Rain.py
```diff
@@ -0,0 +1,233 @@
...
+This script illustrates multiple color schemes for color maps which will allow for those
+impacted by color blindness to see visualizations. Using rainbow color schemes not only
+cause confusion for those impacted by color blindness, but is a poor choice in color scheme 
+for images that may be transferred to a black and white scale for printing.
+This code addresses a handful of options to use in place of rainbow color schemes for 
+use in the matplotlib.pyplot library.
+
+More information on this subject can be found here:
+        -https://agilescientific.com/blog/2017/12/14/no-more-rainbows
+        -https://www.researchgate.net/publication/328361220_The_Effect_of_Color_Scales_on_Climate_Scientists'_Objective_and_Subjective_Performance_in_Spatial_Data_Analysis_Tasks
```
hCraker, line 16: Wonderful idea including the links! This can help further educate scientists on proper color maps.

hCraker: Also, this made me wonder if we should add a page to the gallery specifically mentioning the various resources we have looked at as a group about color maps. @clyne @kmpaul What do you two think?

michaelavs:
Added another example using Planetary Boundary Layer Height data to show use of mono-color maps
image

Plots/CB_Height.py
```diff
@@ -0,0 +1,240 @@
...
+ax4 = plt.subplot(2, 2, 4, projection=projection)
+ax4.coastlines(linewidths=0.5)
+# Change land colors from gray to green to avoid confusion
+ax4.add_feature(cfeature.LAND, facecolor="lightgray")
+plt.subplots_adjust(wspace=0.5)
+# Import an NCL colormap
+newcmp = "Reds"
+
+# Contourf-plot data
+heatmap = t.plot.contourf(
```
clairefio, line 202: There seems to be a lot of function calls with the same/similar arguments that you reuse again many times like contourf and set_axes_limits_and_ticks. Not sure if there would be a way to store these functions and arguments in a variable or helper function, but that would probably reduce program length.

michaelavs:
@clyne @jukent I added you both as reviewers because of past conversations about this topic. Let me know how you feel about the plots and any changes you would like to see with them. I am working on creating a hyper link for all three for this link: https://www.researchgate.net/publication/328361220_The_Effect_of_Color_Scales_on_Climate_Scientists'_Objective_and_Subjective_Performance_in_Spatial_Data_Analysis_Tasks
because it is super long and throws off formatting for the Sphinx html output.

PR #99, opened 2020-06-05
Adding overlay_6 python example

### LINKED ISSUE #97: overlay_6
Description: A color filled contour field, streamlines, and color filled vectors all on one map plot

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/overlay_6.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/overlay_6_lg.png

Category: Overlays

hCraker:
Closes #97

### DIFF
--- a/Plots/Overlays/NCL_overlay_6.py
+++ b/Plots/Overlays/NCL_overlay_6.py
@@ -37,8 +37,8 @@ v500f = xr.open_dataset(gdf.get("netcdf_files/V500storm.cdf"))
 
 p = pf.p.isel(timestep=0).drop('timestep')
 t = tf.t
-u = uf.u
-v = vf.v
+u = uf.u.isel(timestep=0).drop('timestep')
+v = vf.v.isel(timestep=0).drop('timestep')
 u500 = u500f.u.isel(timestep=0).drop('timestep')
 v500 = v500f.v.isel(timestep=0).drop('timestep')
 time = vf.timestep
@@ -70,4 +70,16 @@ cbar = plt.colorbar(pressure, ticks=np.arange(980, 1045, 5), label = 'Sea Level 
 # Add streamline overlay
 plt.streamplot(u500.lon, u500.lat, u500.data, v500.data, transform=ccrs.PlateCarree(), color='black', arrowstyle='->', linewidth=0.5, density=2)
 
+# Add vector grid
+# First thin the data the the vector grid is less cluttered
+x = u['lon'].data[0:36:2]
+y = u['lat'].data[0:33:2]
+u = u.data[0:33:2,0:36:2]
+v = v.data[0:33:2,0:36:2]
+
+# Import color map for vectors
+#vect_cmap = gvcmaps.amwg_blueyellowred
+Q = plt.quiver(x, y, u, v, transform=ccrs.PlateCarree())
+plt.quiverkey(Q, 0.95, 0.05, 20, label="20")
 plt.show()
+
--- a/Plots/Overlays/NCL_overlay_6.py
+++ b/Plots/Overlays/NCL_overlay_6.py
@@ -69,7 +69,7 @@ p = p/100
 t = (t - 273.15) * 9/5 + 32
 
 ###############################################################################
-# Create map:
+# Create plot:
 fig = plt.figure(figsize=(8, 7))
 proj = ccrs.LambertAzimuthalEqualArea(central_longitude=-100,
                                       central_latitude=40)
@@ -90,13 +90,21 @@ cax2 = inset_axes(ax, width='100%', height='7%', loc='lower left',
 # Set extent to include roughly the United States
 ax.set_extent((-128, -58, 18, 65), crs=ccrs.PlateCarree())
 
+#
 # Add map features
+#
+# Using the zorder keyword, we can specify the order of the layering. Lower
+# numbers are plotted before higher ones. For example, the coastlines have
+# zorder=2 while the filled contours have zorder=1. This will draw the
+# coastlines on top of the filled contours.
 transparent = (0, 0, 0, 0)  # RGBA value for a transparent color for lakes
-ax.add_feature(cfeature.OCEAN, color='lightskyblue')
-ax.add_feature(cfeature.LAND, color='silver')
+ax.add_feature(cfeature.OCEAN, color='lightskyblue', zorder=0)
+ax.add_feature(cfeature.LAND, color='silver', zorder=0)
+ax.add_feature(cfeature.LAKES, linewidth=0.5, edgecolor=transparent,
+               facecolor='white', zorder=0)
 ax.add_feature(cfeature.LAKES, linewidth=0.5, edgecolor='black',
-               facecolor=transparent)
-ax.add_feature(cfeature.COASTLINE, linewidth=0.5)
+               facecolor=transparent, zorder=2)       
+ax.add_feature(cfeature.COASTLINE, linewidth=0.5, zorder=2)
 
 #
 # Plot pressure level contour
@@ -104,7 +112,7 @@ ax.add_feature(cfeature.COASTLINE, linewidth=0.5)
 p_cmap = gvcmaps.StepSeq25
 pressure = p.plot.contourf(ax=ax, transform=ccrs.PlateCarree(), cmap=p_cmap,
                            levels=np.arange(975, 1050, 5), add_colorbar=False,
-                           add_labels=False)
+                           add_labels=False, zorder=1)
 plt.colorbar(pressure, cax=cax1, ticks=np.arange(980, 1045, 5))
 # Format colorbar label
 cax1.yaxis.set_label_text(label='\n'.join('Sea Level Pressure'), fontsize=14,
@@ -117,7 +125,7 @@ cax1.tick_params(size=0)
 #
 streams = ax.streamplot(u500.lon, u500.lat, u500.data, v500.data,
               transform=ccrs.PlateCarree(), color='black', arrowstyle='-',
-              linewidth=0.5, density=2)
+              linewidth=0.5, density=2, zorder=5)
 # Divide streamlines into segments
 seg = streams.lines.get_segments()
 # Determine how many arrows on each streamline, the placement, and angles of the arrows
@@ -129,7 +137,7 @@ arrow_dy = np.array([seg[i][1, 1] - seg[i][0, 1] for i in range(0, len(seg), per
 # Add arrows to streamlines
 q = ax.quiver(arrow_x, arrow_y, arrow_dx, arrow_dy, color='black', angles='xy',
               scale=1, units='y', minshaft=3, headwidth=4, headlength=2,
-              headaxislength=2)
+              headaxislength=2, zorder=5)
 
 
 # First thin the data so the vector grid is less cluttered
@@ -151,7 +159,7 @@ norm = mcolors.BoundaryNorm(bounds, wind_cmap.N)  # Assigns colors to values
 # Plot wind vectors
 #
 Q = ax.quiver(x, y, u, v, t, transform=ccrs.PlateCarree(),
-              headwidth=5, cmap=wind_cmap, norm=norm)
+              headwidth=5, cmap=wind_cmap, norm=norm, zorder=4)
 plt.colorbar(Q, cax=cax2, ticks=np.arange(-20, 110, 10), norm=norm,
              orientation='horizontal')
 # Format colorbar label
@@ -160,9 +168,10 @@ cax2.xaxis.set_label_position('top')
 cax2.tick_params(size=0) 
 
 # Add quiverkey and white patch behind it
-ax.quiverkey(Q, 0.925, 0.025, 20, label='20')
 ax.add_patch(mpatches.Rectangle(xy=[0.85, 0], width=0.15, height=0.0925,
-             facecolor='white', transform=ax.transAxes))
+             facecolor='white', transform=ax.transAxes, zorder=3))
+ax.quiverkey(Q, 0.925, 0.025, 20, label='20', zorder=4)
+
 
 # Add title
 ax.set_title('January 1996 Snow Storm\n1996 01 05 00:00 + 0',
--- a/Plots/Overlays/NCL_overlay_6.py
+++ b/Plots/Overlays/NCL_overlay_6.py
@@ -1,0 +1,23 @@
+"""
+NCL_overlay_6.py
+===============
+This script illustrates the following concepts:
+   - Overlaying shaded contours on filled contours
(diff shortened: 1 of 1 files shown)


hCraker:
@clairefio @michaelavs 
So I'm trying to plot filled contours on top of this map. I've tried using xarray and matplotlib to do it, but to no avail. Can either of you help? This is what I currently have. Thanks!

image

hCraker:
image

hCraker:
image

hCraker:
I'm trying to change the scale of the vector grid manually so it will resemble the NCL plot, and I'm having some issues. Below is the default scaling decided by the `plt.quiver()` function. With the default scale applied, small vectors are almost invisible. But, when I increase the scale the large vectors are too long. I've tweaked the kwargs `scale` `scale_units` `minshaft` and `minlength` to no avail, and it seems that the scale of lengths for the vectors covers too wide of a range.  Does anyone know of a way to set both the upper and lower limits of the lengths? @clairefio @michaelavs 

image

clairefio:
> I'm trying to change the scale of the vector grid manually so it will resemble the NCL plot, and I'm having some issues. Below is the default scaling decided by the `plt.quiver()` function. With the default scale applied, small vectors are almost invisible. But, when I increase the scale the large vectors are too long. I've tweaked the kwargs `scale` `scale_units` `minshaft` and `minlength` to no avail, and it seems that the scale of lengths for the vectors covers too wide of a range. Does anyone know of a way to set both the upper and lower limits of the lengths? @clairefio @michaelavs
> 
> image

I haven't tried plotting vector fields yet, but I found an example of someone trying to normalize the arrow sizes here: 

https://stackoverflow.com/questions/44780933/matplotlib-quiver-plotting-with-constant-arrow-size

hCraker:
> I haven't tried plotting vector fields yet, but I found an example of someone trying to normalize the arrow sizes here:
> 
> https://stackoverflow.com/questions/44780933/matplotlib-quiver-plotting-with-constant-arrow-size

Thanks @clairefio! I still want to show the magnitudes, but I'll see if I can modify this approach.

hCraker:
In NCL there is a method that sets the length of the reference vector `vcRefLengthF` and the magnitude of that reference vector 'vcRefMagnitudeF`. There are similar methods that then set the length and magnitude of the smallest vector. NCL then scales every other vector based on the differences in length and magnitude of the minimum and reference vectors. Does anyone know of an equivalent in matplotlib? I can't find one. @jukent @clyne @erogluorhan

jukent:
I think `plt.quiver` is what you want. Perhaps you need to adjust `u` and `v` by some normalization factor (does not need to be linear if the small arrows are too small or the big arrows too big) as in this example: 
https://matplotlib.org/3.2.1/gallery/images_contours_and_fields/trigradient_demo.html#sphx-glr-gallery-images-contours-and-fields-trigradient-demo-py

jukent:
Have you adjusted `u` and `v` by some normalization factor as @clairefio suggested? I think this is the method. Your normalization factor can be nonlinear if you need to make the smallest arrows larger and the largest arrows smaller.

hCraker:
> Have you adjusted `u` and `v` by some normalization factor as @clairefio suggested? I think this is the method. Your normalization factor can be nonlinear if you need to make the smallest arrows larger and the largest arrows smaller.

I'm working on that now, although I'm not sure what nonlinear function I should use. I'm doing some research on normalization at the moment.

jukent:
I would try a logarithmic normalization first.

hCraker:
Range of u: [-8.639618, 17.110382]
Range of v: [-12.022264, 10.227736]

jukent:
What is the smallest magnitude of u and v (positive or negative)?

hCraker:
Smallest magnitude
u: 0.01461792
v: 0.022263527

jukent:
Here is my rough idea of how the symmetrical logarithmic normalization would work:

First want to shift or scale the values to be between 1 and 10 (so that the logarithm output is between 0 and 1). This is easiest if we scale the values to be between 0 and 9, and then add 1.

This would look something like:
new_u = log({9*(abs(u) - min) / (max + min)} + 1)

u and v would use the same min and max values to maintain the same relation between the two. So to get the min do a min of the 2 mins, ie. `min = min([min(abs(u)), min(abs(v)])`, and same for max.

And you'll want the function to treat the negative and positive values differently. If negative, you want the output to be negated again after the normalization.

I can't promise that this will give you exactly what you are looking for, but it is a sound normalization method worth trying.

clyne:
> In NCL there is a method that sets the length of the reference vector `vcRefLengthF` and the magnitude of that reference vector 'vcRefMagnitudeF`. There are similar methods that then set the length and magnitude of the smallest vector. NCL then scales every other vector based on the differences in length and magnitude of the minimum and reference vectors. Does anyone know of an equivalent in matplotlib? I can't find one, and I'm struggling to find a work around. @jukent @clyne @erogluorhan

Hmm. The NCL and Python vector fields look quite different. For example, the arrows along the  california coast are much smaller in the Python plot. I'm actually surprised at the uniformity of the magnitude of the vectors in the NCL plot. Compare the arrows along the coast of california in both plots. It makes me wonder if the NCL plot is really giving an accurate representation of the vector field. Have you tried plotting a histogram of the vector magnitudes and given it an eyeball test to see if arrow lengths in the Python plot look plausible for the distribution? Getting the NCL and Python plots to match may not be as important as presenting the data in a way that accurately represents them.

hCraker:
> Here is my rough idea of how the symmetrical logarithmic normalization would work:
> 
> First we want to shift or scale the values to be between 1 and 10 (so that the logarithm output is between 0 and 1). This is easiest if we scale the values to be between 0 and 9, and then add 1.
> 
> This would look something like:
> `new_u = log({9*(abs(u) - min) / (max + min)} + 1)`

@jukent Here is the plot after using the normailzation technique you suggested first.
image

> Afterthought:
> Perhaps we don't want to scale the min to be 0, but we do want to scale the max to be 1. Then the > function is easier:
> `new_u = log({9* abs(u) / max} + 1)`

Here is the plot with the simplified version where the min isn't scaled but the max is..
image

There doesn't seem to be a visible difference between theses two techniques, but please let me know if I missed something. Between the normalized python maps and the NCL maps, the Python vectors are rotated ever so slightly clockwise. I'm not sure if this is caused by the normalization, the difference in Cartopy and NCL map projections, or something else entirely.

hCraker:
> It makes me wonder if the NCL plot is really giving an accurate representation of the vector field. Have you tried plotting a histogram of the vector magnitudes and given it an eyeball test to see if arrow lengths in the Python plot look plausible for the distribution? Getting the NCL and Python plots to match may not be as important as presenting the data in a way that accurately represents them.

@clyne Here is the histogram of the magnitudes. It looks like the majority of wind speeds are around 3-6 units. Given the spread, I'm also beginning to doubt that the NCL plot is accurately showing the magnitude of the wind.

image

clyne:
@hCraker thanks for producing the histogram. It suggests that the plot of the non-transformed UV data is at lease plausible. 

After reviewing the NCL script and the numerous vector plot "resources" used, I'm not convinced that the NCL plot is wrong. I just find it pretty misleading, particularly given that the intent of this example is to show how to overlay various plotting techniques. Through the - complex, or at least not well documented, IMHO - interactions between the vcMinFracLengthF, vcMinMagnitudeF, vcRefLengthF, and vcRefMagnitudeF resources, the NCL plot has managed to make all of the glyphs more or less uniform in length. Moreover, the reference annotation in the bottom right corner suggests that all of the wind speeds are close to 20, which they are not. Given the misleading NCL plot, and since the intent of this example is to show how to overlay plots, not illustrate different ways to control vector glyphs, I would be inclined to NOT transform the data solely to achieve visual agreement with the NCL version, and to simply add a comment to the script wrt how this example differs from the NCL version and why we chose to do it this way.


cc: @jukent

jukent:
Sounds reasonable @clyne. I hope there is another example in the gallery on manipulating vector glyphs.

clyne:
> Sounds reasonable @clyne. I hope there is another example in the gallery on manipulating vector glyphs.

I looked, but didn't see anything that was comprehensive; different NCL glyph "resources" are spread out over a number of examples. There is no reason, however, that an example could not be created from scratch that illustrates glyph/vector control.

jukent:
@hCraker Have you looked at this yet? https://matplotlib.org/3.1.0/gallery/axes_grid1/demo_colorbar_with_inset_locator.html

hCraker:
> @hCraker Have you looked at this yet? https://matplotlib.org/3.1.0/gallery/axes_grid1/demo_colorbar_with_inset_locator.html

I have not yet. I'll see if I can adapt this example. Thanks!

hCraker:
@clairefio You were right that I needed to use `inset_axes()`. The issue was I wasn't using some of the kwargs I should have been. Thank you @jukent for suggesting the example that helped me figure that out.

image

hCraker:
Let me know if anyone has any suggestions.
image

Plots/Overlays/NCL_overlay_6.py
```diff
@@ -0,0 +1,151 @@
...
+    In the NCL version of this plot the vectors for the winds are nearly
+    uniform in length. Given the reference vector in that figure, the wind
+    speeds appear to be near 20 units. A historgram reveals that this is not a
+    true representation of the data, as the magnitude of the majority of wind
+    vectors are between 3 and 6 units with only a handful being greater than 13
+    and only one near 20. Because of this, we have chosen not to manipulate the
+    vector glyphs to appear more uniform as this would poorly represent the
+    data and be misleading. This is why the reference vector in this Python
+    example is longer than the NCL example and why the lengths vary more
+    between the minimum and maximum wind speeds.
```
clyne, line 26: How about adding that in the Python version the glyph (arrow) lengths are proportional to the velocity magnitude, and in the NCL version they are not?

Plots/Overlays/NCL_overlay_6.py
```diff
@@ -0,0 +1,151 @@
...
+                  bbox_to_anchor=(0, -0.15, 1, 1),
+                  bbox_transform=ax.transAxes,
+                  borderpad=0)
+
+# Set extent to include roughly the United States
+ax.set_extent((-128, -58, 18, 65), crs=ccrs.PlateCarree())
+
+# Add map features
+transparent = (0, 0, 0, 0)  # RGBA value for a transparent color for lakes
+ax.add_feature(cfeature.OCEAN, color='lightskyblue', zorder=0)
```
clyne, line 92: Can you add a comment about what the zorder keyword does?

Plots/Overlays/NCL_overlay_6.py
```diff
@@ -0,0 +1,151 @@
...
+#
+# Plot streamline overlay
+#
+ax.streamplot(u500.lon, u500.lat, u500.data, v500.data,
+              transform=ccrs.PlateCarree(), color='black', arrowstyle='->',
+              linewidth=0.5, density=2, zorder=4)
+
+
+# First thin the data so the vector grid is less cluttered
+x = u['lon'].data[0:36:2]
```
clyne, line 120: Can you stride the arrays without using the hard-coded dimension lengths?

Plots/Overlays/NCL_overlay_6.py
```diff
@@ -0,0 +1,151 @@
...
+p_cmap = gvcmaps.StepSeq25
+pressure = p.plot.contourf(ax=ax, transform=ccrs.PlateCarree(), cmap=p_cmap,
+                           levels=np.arange(975, 1050, 5), add_colorbar=False,
+                           add_labels=False, zorder=2)
+plt.colorbar(pressure, cax=cax1, ticks=np.arange(980, 1045, 5))
+# Format colorbar label
+cax1.yaxis.set_label_text(label='\n'.join('Sea Level Pressure'), fontsize=14,
+                          rotation=0)
+cax1.yaxis.set_label_coords(-0.5, 0.9)
+
```
michaelavs, line 110: Overall, looks great! I noticed in the original NCL plot the color bars do not have any ticks, if you want to also do this on your plot, I found using/adding cax1.tick_params(size=0) here (line 110) and cax2.tick_params(size=0) after line 140 would remove the tick marks. Up to you if you want to make this change or not!

jukent:
I will review after @clyne's comments have been addressed.

Plots/Overlays/NCL_overlay_6.py
clairefio, line 92: I'm pretty sure you don't need a zorder argument here because default drawing order is already assumed to be 0. It does make sense to keep it in if you just want to be consistent for the reader though.

jukent: Consistent with what? I think you need to keep this if there was another zorder specification somewhere in the plot, but otherwise it should be removed.

hCraker: I had some issues with the layering of the various plots and features so I had to use zorder. I'll go back and add a comment about it.

hCraker: While piecing this example together, I found that I needed to use `zorder` to correctly layer the different types of plots. Now it is no longer needed, so I have removed all uses of it.

hCraker: While piecing this example together, I found that I needed to use `zorder` to correctly layer the different types of plots. Now it is no longer needed, so I have removed all uses of it.

hCraker:
@jukent I have made edits based on everyone's suggestions. Please let me know if any of you have more.

hCraker:
I believe I'm about done. Thank you to @clairefio for figuring how to add more arrows to streamlines.

image

Plots/Overlays/NCL_overlay_6.py
```diff
@@ -0,0 +1,171 @@
...
+   - Add arrows to streamlines
+   
+See following URLs to see the reproduced NCL plot & script:
+    - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/overlay_6.ncl
+    - Original NCL plots: https://www.ncl.ucar.edu/Applications/Images/overlay_6_lg.png
+
+Differences between NCL example and this one:
+    In the NCL version of this plot the vectors for the winds are nearly
+    uniform in length. Given the reference vector in that figure, the wind
+    speeds appear to be near 20 units. A historgram reveals that this is not a
```
erogluorhan, line 20: Typo should be corrected: "historgram"

erogluorhan:
Code and plot looks great! However, there is an output after the plot, which you can see by observing your generated html file by calling `make html`:

image

Our approach is to get rid of all of such warning or error outputs (unless they are generated by inevitable 3rd party behaviors such as deprecation etc.). So, can you please check whether your several RuntimeWarnings could be eliminated?

Issue #100, opened 2020-06-05
stream_9


clairefio:
Description: This example shows how to select a color palette for colored streamlines (first frame) and to make streamlines partially transparent (second frame)

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/stream_9.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/stream_9_1_lg.png

Category: Streamlines

PR #101, opened 2020-06-06
Stream_9

clairefio:
Issue #100 

Updated plot:

stream_9


clairefio:
@michaelavs @hCraker Do either of you know a way to plot multiple arrows on each streamline instead of just one?

hCraker:
> @michaelavs @hCraker Do either of you know a way to plot multiple arrows on each streamline instead of just one?

Not at the moment. I'm trying to figure that out for PR #99. I'll let you know if/when I learn how to do it.

michaelavs:
@hCraker @clairefio These two resources may point you both in the right direction: https://problemsolvingwithpython.com/06-Plotting-with-Matplotlib/06.15-Quiver-and-Stream-Plots/
and the second response to this question is very long code wise but the output does have multiple arrows per line:
https://stackoverflow.com/questions/43150872/number-of-arrowheads-on-matplotlib-streamplot

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,80 @@
...
+from geocat.viz import util as gvutil
+import matplotlib.colors as colors
+from mpl_toolkits.axes_grid1.inset_locator import inset_axes
+################################################################################
+# Make color map 
+
+colormap = colors.ListedColormap(['darkblue', 'mediumblue', 'blue', 'cornflowerblue', 'skyblue', 'aquamarine',
+        'lime', 'greenyellow', 'gold', 'orange', 'orangered', 'red', 'maroon'])
+
+colorbounds = [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52]
```
michaelavs, line 33: Is it possible to do np.arange with these numbers? It may solve the color bar labels using 0 and 52 (number wise) but not being that way on the original projection. If not, I don't think it would be a problem to leave it as is.

hCraker:
> @hCraker @clairefio These two resources may point you both in the right direction: https://problemsolvingwithpython.com/06-Plotting-with-Matplotlib/06.15-Quiver-and-Stream-Plots/
> and the second response to this question is very long code wise but the output does have multiple arrows per line:
> https://stackoverflow.com/questions/43150872/number-of-arrowheads-on-matplotlib-streamplot

From that stackoverflow post, it looks like there are a few different functions that people wrote as a solution and there is not shorter, more elegant solution. Perhaps if we write a function for this, we should add it as a utility function geocat-util.

clyne:
The MPL flow visualization functions are, IMHO, a weak spot compared to NCL. There are definitely opportunities for improvement, some of which can be handled through convenience functions, and others that may require upstream development.

clairefio:
@hCraker I figured out how to plot multiple arrows by dividing the streamlines into segments and then plotting the arrows individually (I think it was like the third solution on this link: https://stackoverflow.com/questions/43150872/number-of-arrowheads-on-matplotlib-streamplot), but I am still having trouble scaling them and applying the color map to them. Do you have any ideas on how to accomplish this?

arrows

hCraker:
> @hCraker I figured out how to plot multiple arrows by dividing the streamlines into segments and then plotting the arrows individually (I think it was like the third solution on this link: https://stackoverflow.com/questions/43150872/number-of-arrowheads-on-matplotlib-streamplot), but I am still having trouble scaling them and applying the color map to them. Do you have any ideas on how to accomplish this?
> 
> arrows

To color the arrows, use the 5th positional argument for `plt.quiver()`. The argument takes an array with the numeric values that the colors should be based on (i.e. temperature, wind speed). You can also specify the color scheme by using the kwarg `cmap`.

About the scaling, note that `scale_units` affects the units for length while `units` specifies the units for every other dimension of the arrows. You may have to play with `headwidth`, `headlength`, and the other kwargs. I hope this helps!

clairefio:
Updated plot w/o colorbar: 

stream_9 updated

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,166 @@
...
+seg = streams.lines.get_segments()
+
+# Determine how many arrows on each streamline, the placement, and angles of the arrows
+period = 7
+arrow_x = np.array([seg[i][0, 0] for i in range(0, len(seg), period)])
+arrow_y = np.array([seg[i][0, 1] for i in range(0, len(seg), period)])
+arrow_dx = np.array([seg[i][1, 0] - seg[i][0, 0] for i in range(0, len(seg), period)])
+arrow_dy = np.array([seg[i][1, 1] - seg[i][0, 1] for i in range(0, len(seg), period)])
+
+#plt.scatter(arrow_x, arrow_y, zorder=12, color='pink')
```
hCraker, line 91: Is this line still needed?

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,166 @@
...
+plotsize = fig.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
+plotwidth = (plotsize.width*fig.dpi)/2
+plotheight = (plotsize.height*fig.dpi)/2
+
+axsize = ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
+axwidth = (axsize.width*fig.dpi)/2
+axheight = (axsize.height*fig.dpi)/2
+
+# Get x and y data, transform it into pixels, return RGB value of the pixels
+def getPixelVals(x, y):
```
hCraker, line 106: There are a lot of hard coded values here. Could you generalize this function or add comments about what these values represent?

clairefio: The hard coded values are the transform needed to find the location of the pixel. This solution is very roundabout! Because I am dividing each streamline into separate segments, and then plotting the arrows on those segments, I can't use the original 'U.data' argument used in the full streamline to determine the color. If you have a solution that doesn't involve manually finding the color of the pixels, that would be much appreciated haha.

hCraker: The color of the streamline is the magnitude of the wind right? If so, you can calculate the magnitude using the U and V components then use the 5th positional argument of `quiver` to specify that the magnitude is the value that will determine the color. Assign `cmap` to the same color map as the streamline and it should work.

clairefio:
Updated plot:

stream_9 updated

clairefio:
@NCAR/geocat I was able to change the matplotlib streamplot.py code to allow multiple colored arrows to be plotted on each streamline. I added the edited streamplot.py file to this directory, but I'm not sure where I should go from here.

stream_9 updated

clairefio:
Updated plot:

stream_9 updated

clyne:
Is there a question?

clairefio:
> Is there a question?

Yeah, should I just use the original matplotlib streamplot.py file, push the code that doesn't have the arrows on the streamlines, and then make a PR in the matplotlib repository?

clyne:
Gotcha. So that we don't hold up your PR, let's add the streaplot.py file to the GeoCAT-viz repo. 

Once it has been reviewed internally, lets's see if we can submit your changes to streamplot.py to Matplotlib.

clairefio:
Updated plot (without one arrow per streamline):

stream_9

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,95 @@
+"""
+NCL_stream_9.py
+===============
+This script illustrates the following concepts:
+   - Showing features of the new color display model
```
clyne, line 5: Is this comment relevant for the Python version?

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,95 @@
...
+   - Using opacity to emphasize or subdue overlain features
```
clyne, line 6: Is this comment true? I.e. does the plot use opacity?

Plots/Streamlines/NCL_stream_9.py
```diff
@@ -0,0 +1,95 @@
...
+   - Using stLevelPalette resource to assign a color palette
```
clyne, line 7: Relevant comment?

Plots/Streamlines/NCL_stream_9.py
clairefio: I just added a set_alpha line that changes the opacity of the plot, so this comment is applicable now. I took out the other comments.

PR #103, opened 2020-06-08
Add GitHub action and ReadTheDocs badges

andersy005:
This makes it easy to monitor the status of readthedocs builds without needing to  check the main readthedocs webpage:

Screen Shot 2020-06-08 at 1 12 04 PM


andersy005:
You are welcome ;)

Issue #92, opened 2020-06-04
colormap comparison plot


clyne:
Produce one or more montage (multiple subpanels) plots as necessary that illustrate different colormaps applied to the same data. One of the colormaps should be the rainbow as an illustration of what NOT to use. 

Choose data sets that are appropriate for different color map types (e.g. temperature or velocity components for divergent maps, a field such with no "special values" such as rain fall for a "color scale" map).

The choice of maps should be based on current best practices. See

http://www.kennethmoreland.com/color-advice/

and references therein

michaelavs:
@clyne should this be a part of the Plots sub directory, gallery subdirectory, or a new subdirectory completely? I am about to push the first version of a temperature projection but want to make sure we're all on the same page where this category of plots will go

clyne:
How about we just create a new sub directory under Plots called "Color"?

PR #106, opened 2020-06-10
Comparisons

michaelavs:
Resolves issue [#92](https://github.com/NCAR/GeoCAT-examples/issues/92)
Updated from earlier version to branch from local master instead of other local branches
image
image
image


michaelavs:
@clyne I've been looking into the changes you suggested with the text descriptions and potentially removing the lat/lon ticks for more room for the text. From what I've been seeing, the ticks will need to be present in order to do a below the projection text description, but I think this may end up looking/working better for what we want to accomplish. Here is an example of what I currently have for the text descriptions on the projections:
image

I also found the Kindlmann (and kindlmann extended) color scheme should be a part of the mpl color scheme gallery (if not, it is incredibly close to the color scheme of viridis, so that may be a viable back up should there be problems using it). Based on one article I was reading about types of color blindness, a blue/green scheme could be a poor choice for color blind individuals because the contrasts are too similar, but, I think it would be valuable to include this information because it may also prove valuable with the discussion of projections being black and white print friendly.

In regards to revisions on the color schemes, it may be best to do two "bad" choices and why and then followed by two "good" choices and why. With having 3 separate types of data set in this part of the gallery, each could be almost assigned to a type of color scheme. For example, the temperature projection would be a great way to showcase gradient type color schemes because of how the projection is laid out. The rain projections look better using the viridis and diverging schemes. The PBL Height looks best with mono-color schemes. 

I'm going to definitely continue tinkering around with layouts and implementing the different schemes and uploading the projections for critique, but wanted to see if the direction I'm heading in seems logical to everyone else!

clyne:
That all sounds good, Michaela. There is no deadline on this, so take whatever time you need to fully explore the issue and consider the options so that you can come up with something that you are satisfied with. Re the text description: I was thinking of changing the subplot titles from the current "Rainbow color projection of ...", etc. to simply the name of the name of the colormap. Or maybe the name of the colormap and a brief (fits on one line) description. And then having a single overall plot title at the top that lists the variable being displayed. E.g. "Temperature"

michaelavs:
How do we feel about this version?
image
I am having some difficulty with formatting the descriptions and figuring out what to put in each one, so if there are any changes to the wording/what to put there, definitely let me know!

hCraker:
I think the descriptions of the pros and cons of the color schemes are great. If you can, I recommend adding some space between each description so it's not a wall of text. If you separate the figure captions just a bit, it will become easier to read and to distinguish one caption from another. You've done a fantastic job!

michaelavs:
I've been playing around with the Sphinx out put of the files and currently I have this:
image
I would ideally like to have the four figure descriptions at the bottom of the page but can't figure out a way to do so. I tried replicating the three quote mark format from the top of the page, but at the bottom of the page, this is just read in as a normal comment from a python notebook and not formatted as it is at the top of the webpage. Is there a way to work around this/make the bottom format the same as the top?

hCraker:
I tried running `make html` locally and it does run; however, gallery pages for these plots are not created. @michaelavs does `make html` work as expected for you?

michaelavs:
> I tried running `make html` locally and it does run; however, gallery pages for these plots are not created. @michaelavs does `make html` work as expected for you?

I haven't run make html recently (just getting touch ups and merges done first then going back for make html), but in the past I have been able to open the pages from command line and see them. When I get the last push merged with this PR, I will run make html and see what happens

michaelavs:
I am going to be closing this PR because my local branch will not point to this PR. New PR can be found [here](https://github.com/NCAR/GeoCAT-examples/pull/147)

Issue #17, opened 2019-12-20
tm_2


clyne:
Description: Tic mark manipulation

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/tm_2.ncl

Plot URL: https://www.ncl.ucar.edu/Applications/Images/tm_2_lg.png

khallock:
Closed by #41

PR #108, opened 2020-06-11
Sat 1

clairefio:
Updated plot:

sat_1


Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,82 @@
...
+# Add land, coastlines, and ocean features
+ax.add_feature(cfeature.LAND, facecolor='lightgray')
+ax.add_feature(cfeature.COASTLINE, linewidth=.5)
+ax.add_feature(cfeature.OCEAN, facecolor='lightcyan')
+
+# Plot contour data
+p = wrap_U.plot.contour(ax=ax,
+                        transform=ccrs.PlateCarree(),
+                        linewidths=0.5,
+                        levels=15,
```
hCraker, line 62: In the NCL script, it specifies the range and spacing of the levels which results in 30 contour lines. Could you do the same for your plot either by explicitly setting the max and min values or `levels=30` so that the contours are denser?

clairefio:
Updated plot:
sat_1

clairefio:
> It looks great so far! I would just draw more contour lines to more closely resemble the NCL example. Also, see if you can find a way to indicate the area of low pressure denoted by "L" in the NCL plot.

I set the central latitude/longitude and the number of contour levels to be the same as the NCL version, but the orientation of the globe and the contour lines still seem to be just a little bit off... Can you see anything that I might be missing?

hCraker:
> > It looks great so far! I would just draw more contour lines to more closely resemble the NCL example. Also, see if you can find a way to indicate the area of low pressure denoted by "L" in the NCL plot.
> 
> I set the central latitude/longitude and the number of contour levels to be the same as the NCL version, but the orientation of the globe and the contour lines still seem to be just a little bit off... Can you see anything that I might be missing?

It looks like the NCL plot is zoomed in on the US ever so slightly. I don't think it is that big of a discrepancy so I wouldn't worry about it.
For the pressure labels, they are a bit cluttered in places. You can manually set where you want them via `ax.clabel` using the kwarg `manual` which accepts an array of coordinates in the format (lon, lat). The easiest way to find the coordinates is to hover over the plot when it pops up in its own window, and the coordinates for your cursor are listed in the lower left.

clairefio:
Updated plot:

sat_1

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,109 @@
...
+###############################################################################
+# Create plot
+
+# Set figure size
+fig = plt.figure(figsize=(8, 8))
+
+# Set global axes with an orthographic projection
+proj = ccrs.Orthographic(central_longitude=270, central_latitude=45)
+ax = plt.axes(projection=proj)
+ax.set_global()
```
hCraker, line 51: The NCL plot looks slightly zoomed in. Maybe using `ax.set_extent()` could make your plot look the same.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,109 @@
...
+                        levels=30,
+                        cmap='black',
+                        add_labels=False)
+
+# Specify array of contour levels to be labeled- These values were found by setting 'manual'
+# argument in ax.clabel call to 'True' and then hovering mouse over desired location of
+# countour label to find coordinate (which can be found in bottom left of figure window)
+
+# low pressure contour levels- these will be plotted as a subscript to an 'L' symbol
+lowClevels = [(-3.825, 4.063), (1.725, 2.026), (1.663, 4.479)]
```
hCraker, line 73: I'd be hesitant to manually mark the coordinates of the low pressure areas. There should be a local minimum pressure in the data, so use that to find the exact latitude and longitude of the lowest pressure. You can then use that coordinate to draw the L.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,109 @@
...
+           (4.157, 0.7792), (3.118, 0.3219), (1.164, -.2601),
+           (0.4781, 0.8831), (-2.972, 0.1348), (-4.344, -0.177),
+           (-1.995, -2.713), (-0.1247, -2.048), (-0.4781, 4.313),
+           (-0.3741, 3.128), (-1.538, -0.1354), (-0.5196, -0.9461),
+           (2.661, 1.735), (4.22, -1.694), (1.455, 2.754), (1.164, -1.632),
+           (1.642, -2.983), (2.827, 4.25), (4.303, 2.255),
+           (0.8107, 4.271), (-0.4157, 1.756)]
+
+# multiply each value by 10^6 to get correct window coordinates
+lowClevels = [(x[0]*1000000, x[1]*1000000) for x in lowClevels]
```
hCraker, line 87: I'm not sure why you'd have to scale the values. When I manually selected where contour labels would be drawn in #110, I only needed to specify the latitude and longitude values of where the label should go. Do you think this has to do with your projection not being a grid?

clairefio: I think I am unable to use lat/lon values because my projection is orthographic. Do you know if there is a transform or something I can use on the lat/lon values to get it into a useable coordinate for the manual label plotting?

hCraker: You could try using `transform_points()` in cartopy. You could specify your label coordinates using regular latitude and longitude coords and then transform it into orthographic coordinates.

Plots/Satellites/NCL_sat_1.py
clairefio: I tried doing this by using the "levels" argument in ax.clabel and setting it to the local minimum pressures ([972, 976, 960]). However, only one of the L labels showed up and it was off to the side of where it showed up in the NCL graph. Do you know why this might be?

hCraker: Are you sure those contour lines were drawn by `contour`? You can force those lines to be drawn by using the kwarg `levels` for `contour`. You can give it an array of the levels you want.

Plots/Satellites/NCL_sat_1.py
hCraker: Ah okay. I'm not sure how to fix that then.

Plots/Satellites/NCL_sat_1.py
clairefio: Even when I set the contour "levels" kwarg to levels=[960, 972, 976], I still get this:

polyg_19_limited

(This is an example of only plotting those levels)

hCraker: If that's the case then I think the solution you have is the best you can do.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,114 @@
...
+                        levels=30,
+                        cmap='black',
+                        add_labels=False)
+
+# Specify array of contour levels to be labeled- These values were found by setting 'manual'
+# argument in ax.clabel call to 'True' and then hovering mouse over desired location of
+# countour label to find coordinate (which can be found in bottom left of figure window)
+
+# low pressure contour levels- these will be plotted as a subscript to an 'L' symbol
+lowClevels = [(51.54, 169.59), (74.78, 4.54), (60.12, -57.0)]
```
hCraker, line 74: The manual placement of contour labels is much easier to follow now. Good job!

I did notice that the subscripts for the low pressure labels are different from the NCL plot. I recommend now that you can specify the coordinates for the labels in lat and lon, find the coordinates for the local pressure minimums and use that to plot the labels. The other labels are drawn where they aesthetically make sense and in a way that makes them legible, but the low pressure labels are drawn over a low and need to be more accurate than estimating. This should help with that.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,114 @@
+"""
+NCL_sat_1.py
+===============
+This script illustrates the following concepts:
+   - Drawing line contours over a satellite map
+   - Manually labeling contours
+   - Transforming coordinates
```
hCraker, line 7: I would add that this example illustrates how to draw an orthographic map. If someone is trying to google for and example like this, adding that comment should make it easier for them to find it.

Plots/Satellites/NCL_sat_1.py
clairefio: One thing I don't understand about the original NCL plot is how they got the L975 pressure marking. From my understanding, they have a list of contours to be plotted that ranges from 948 to 1064 with a step of 4. So if the list of levels looks like [948, 952, 956... 1064], how could 975 be plotted if it's not in the list? 

sat 1 ncl script

clairefio: I wrote a helper function to try to look through the xarray data, find the low pressure values, and then use the indexes of those to find the corresponding lat/lon values. This was the result of my most recent commit:

polyg_19_limited

If you have any ideas on what I'm doing wrong/how to get the low pressure coordinates without hard coding in the estimates it would be appreciated.

hCraker: These are the lines that draw the Low Pressure labels, and it appears to be a functionality build into NCL.
image

Plots/Satellites/NCL_sat_1.py
```diff
@@ -32,11 +33,31 @@
...
+
+    coordlist = []
+
+    for x in range(len(U.data)):
+        for y in range(len(U.data[x])):
+            if (int)(U.data[x][y]) == level:
+                latval = -lat[x]
+                lonval = -lon[y]+270
+                # If in scope of area of map depicted on plot
+                if (0<latval<90) and (-180<lonval<0):
```
hCraker, line 55: Instead of limiting this function by the region you are trying to draw a low pressure marker, why not define a low pressure point by how different it is from the pressure at a point nearby? For example, say a local minimum has been found. Check if that minimum has a pressure that is lower than it's surroundings by at least 30 hPa. If so, draw the label. Now the pressure threshold and the distance away for the reference pressure seem arbitrary, but it's a start and could be applied globally.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -58,11 +79,16 @@
 ax.add_feature(cfeature.BORDERS, linewidth=.5)
 ax.add_feature(cfeature.LAKES, facecolor='lightcyan', edgecolor='k', linewidth=.5)
 
+# Make array of the contour levels that will be plotted
+contours = np.arange(948, 1060, 4)
+contours = np.append(contours, 975)
```
hCraker, line 84: Why are you appending this contour level? You can add a label for the 975 hPa low using a function similar to the one you just defined.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -71,7 +97,15 @@
 # countour label to find coordinate (which can be found in bottom left of figure window)
 
 # low pressure contour levels- these will be plotted as a subscript to an 'L' symbol
-lowClevels = [(51.54, 169.59), (74.78, 4.54), (60.12, -57.0)]
+
+#lowClevels = [(51.54, 169.59), (74.78, 4.54), (60.12, -57.0)]
+
+coord956 = findLowPressureCoords(956)
```
hCraker, line 103: By specifying the low pressure value as an argument, it assumes that the user knows where the low pressure areas are. Try reworking this function to find low pressure areas for the user instead.

Plots/Satellites/NCL_sat_1.py
clairefio: I was trying to use levels before and it would throw an error every time I tried to clabel a contour that wasn't in the levels list.

hCraker: Ah understandable. I didn't consider that.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,162 @@
...
+                        cmap='black',
+                        add_labels=False)
+
+# Specify array of contour levels to be labeled- These values were found by setting 'manual'
+# argument in ax.clabel call to 'True' and then hovering mouse over desired location of
+# countour label to find coordinate (which can be found in bottom left of figure window)
+
+# low pressure contour levels- these will be plotted as a subscript to an 'L' symbol
+#lowClevels = [(51.54, 169.59), (74.78, 4.54), (60.12, -57.0)]
+lowClevels = findLowPressureCoords()
```
michaelavs, line 118: For this line, I would define your xradius, yradius, and pressureThreshold here instead of in the helper function. This way the function can be used in a more dynamic fashion for people that may use this code as a template for how to use/call functions in python, or want to look at different x/yradius or pressureThreshold values.

clairefio:
@hCraker I did what you suggested and used the gradient rather than finding the derivative of x and y individually. My function is still giving me the wrong values... I'm not sure if I'm mapping the U.data (pressure) value to the correct coordinates. Can you take a look at my helper function findLocalMinima() and tell me if you see anything wrong with the way I tried to accomplish this.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+import geocat.viz.util as gvutil
+
+###############################################################################
+# Read in data:
+
+# Open a netCDF data file using xarray default engine and load the data into xarrays
+ds = xr.open_dataset(gdf.get("netcdf_files/slp.1963.nc"), decode_times=False)
+
+# Get data from the 24th timestep
+U = ds.slp[24, :, :]
```
hCraker, line 33: Refactor `U` to something more representative of the data like `pressure`

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+
+# Translate short values to float values
+U = U.astype('float64')
+
+# Convert Pa to hPa data
+U = U*0.01
+
+# Fix the artifact of not-shown-data around 0 and 360-degree longitudes
+wrap_U = gvutil.xr_add_cyclic_longitudes(U, "lon")
```
hCraker, line 42: Similar to above. Refactor this to reflect the data

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+            if coordarr[x][y][0] == lat and coordarr[x][y][1] == lon:
+                return U.data[x][y]
+
+def makeCoordArr():
+    # Returns an array of (lat, lon) coord tuples with the same dimensions as the pressure data
+    coordarr = []
+    for x in np.array(U.lat):
+        temparr = []
+        for y in np.array(U.lon):
+            temparr.append((x,y))
```
hCraker, line 61: Longitudes are the x coordinates and latitudes are the y coordinates for cartopy. I suggest changing this to reflect that fact.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+           (42.19, -108.73), (49.66, -111.25), (41.93, -127.83),
+           (25.64, -92.49), (29.08, -77.29), (16.42, -77.04),
+           (57.59, -95.93), (84.47, -156.05), (82.52, -17.83),
+           (41.99, -76.3), (41.45, -48.89), (37.55, -33.43),
+           (17.17, -46.98), (63.67, 1.79), (67.05, -58.78),
+           (53.68, -44.78), (53.71, -69.69), (52.22, -78.02),
+           (44.33, -16.91), (35.17, -95.72), (73.62, -102.69)]
+
+# Transform the low pressure contour coordinates from geographic to projected
+lowclevelpoints = proj.transform_points(ccrs.Geodetic(), np.array([x[1] for x in lowClevels]), np.array([x[0] for x in lowClevels]))
```
hCraker, line 160: You may have fixed the flipped lon/ lat coordinates here by calling `x[1]` as the x coordinates and `x[0]` as the y coordinates. You will need to change this line if you fix `makeCoordsArr`

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+lowClevels = [(x[0], x[1]) for x in lowclevelpoints]
+
+# Transform the regular pressure contour coordinates from geographic to projected
+clevelpoints = proj.transform_points(ccrs.Geodetic(), np.array([x[1] for x in clevels]), np.array([x[0] for x in clevels]))
```
hCraker, line 164: Same as above

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+    minimacoords = []
+
+    # For every common zero in both gradient arrays 
+    for x in commonzeroes:
+
+        xval = x[0]
+        yval = x[1]-1
+
+        # If the gradient value is a "zero", and if the U.data value is less than minPressure
+        if -minPressure <= wrap_U.data[xval][yval] < minPressure:
```
hCraker, line 97: Why are you comparing the pressure at a location of a potential low against the positive *and* negative value of `minPressure`? There shouldn't be any negative pressures.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Set number that a derivative must be less than in order to classify as a "zero"
+    bound = 0.00001
+
+    # Get global gradient of U.data
+    grad = np.gradient(wrap_U.data)
+    arr1 = grad[0]
+    arr2 = grad[1]
+
+    firstzeroes = np.argwhere(arr1<=bound)
```
hCraker, line 78: Gradients can be positive or negative. I recommend checking to see if `arr1` is between plus or minus `bound` so you get more accurate results.

Plots/Satellites/NCL_sat_1.py
```diff
@@ -0,0 +1,191 @@
...
+    secondzeroes = np.argwhere(arr2<=bound)
```
hCraker, line 79: Same as above but for `arr2`

clairefio:
@hCraker I think I fixed it! I used scikit learn to K-means cluster the local minima we found, and then found the minimums of each cluster. This is what the plot looks like now:

sat_1

Plots/Satellites/NCL_sat_1.py
clairefio: This is a mistake, I meant to add this condition to when I was checking to see if arr1 gradient values were within positive and negative bounds, not for checking pressure values.

hCraker:
> @hCraker I think I fixed it! I used scikit learn to K-means cluster the local minima we found, and then found the minimums of each cluster. This is what the plot looks like now:
> 
> sat_1

Fantastic! It looks good to me! I would just clean up your code and helper functions from here. Once that's done, maybe some of the helper functions you made to find where to put the low pressure labels should be added to geocat-viz. I suggest opening an issue in geocat viz to add that functionality.

clairefio:
> > @hCraker I think I fixed it! I used scikit learn to K-means cluster the local minima we found, and then found the minimums of each cluster. This is what the plot looks like now:
> > sat_1
> 
> Fantastic! It looks good to me! I would just clean up your code and helper functions from here. Once that's done, maybe some of the helper functions you made to find where to put the low pressure labels should be added to geocat-viz. I suggest opening an issue in geocat viz to add that functionality.

Will do! Thanks for your help.

clyne:
Just adding my comments from our conversation so we don't loose track of them, mostly around generalizing the min finding function so we can include it in geocat-viz

1. Generalize the min finding function(s) so that it works with min or max extrema
2. Ditto for working with any variable (not just pressure)
3. K-means clustering requires a priori knowledge of the number of clusters. Explore (Google) other extrema finding approaches for 2D data that might be parameterless. Alternatively, consider logic that would help (perhaps optionally)  select the right number of clusters, possibly by comparing distances between them. 
4. I believe the current implementation only works for 1D lat and lon coordinates. Can we generalize to work with curvilinear grids (i.e. 2D lat and lon)? 

This will be really awesome to have in geocat-viz!!

clairefio:
> Just adding my comments from our conversation so we don't loose track of them, mostly around generalizing the min finding function so we can include it in geocat-viz
> 
> 1. Generalize the min finding function(s) so that it works with min or max extrema
> 2. Ditto for working with any variable (not just pressure)
> 3. K-means clustering requires a priori knowledge of the number of clusters. Explore (Google) other extrema finding approaches for 2D data that might be parameterless. Alternatively, consider logic that would help (perhaps optionally)  select the right number of clusters, possibly by comparing distances between them.
> 4. I believe the current implementation only works for 1D lat and lon coordinates. Can we generalize to work with curvilinear grids (i.e. 2D lat and lon)?
> 
> This will be really awesome to have in geocat-viz!!

I looked for different methods of clustering data, and I found DBSCAN (Density-based spatial clustering of applications with noise), which is also included in the scipy module. It is bettern than K-means for coordinate data because it clusters data based on distance and density, so you don't need to specify the number of clusters. It seems to identify all the clusters better than using k-means did! I am working on generalizing my function now.

erogluorhan:
@clairefio in regards to the build failure: This PR introduces `NCL_sat_1.py `under a new folder "**MapProjections**" but that new folder does not have a **"README.rst"** file in it, which would help **Sphinx** generate the apropriate **HTML** file for `NCL_sat_1` plot. In addition, even though this PR tries to create "**MapProjections**", it has been already existing in **master** branch. So, I would suggest to sync your branch (**sat_1**) to upstream's (i.e. **NCAR:geocat-examples**) `master` branch so that you could have **"README.rst"** and other files (i.e. `NCL_proj_2_lg.py` and `NCL_proj_3_lg.py`) from it.

clyne:
> @clairefio in regards to the build failure: This PR introduces `NCL_sat_1.py `under a new folder "**MapProjections**" but that new folder does not have a **"README.rst"** file in it, which would help **Sphinx** generate the apropriate **HTML** file for `NCL_sat_1` plot. In addition, even though this PR tries to create "**MapProjections**", it has been already existing in the upstream's (i.e. **NCAR:geocat-examples**) `master` branch. So, I would suggest to sync your branch (**clairefio:sat_1**) to upstream's `master` branch so that you could get up-to-date **"README.rst"** and other files (i.e. `NCL_proj_2_lg.py` and `NCL_proj_3_lg.py`) from it. This way, we could also avoid conflicts when merging your branch to upstream's `master`. Hope it helps; let me know if any point not clear.

FYI there is an existing MapProjections directory, but it was created fairly recently. @clairefio you may just need to merge GeoCat-examples master into your feature branch.

erogluorhan:
This PR cannot pass the build checks likely because **scikit-learn** dependency should have been added to `conda_environment.yml` as `scikit-learn `instead of `sklearn`.

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+def findLocalExtrema(da, highVal=1040, lowVal=975, eType='Low'):
+    """
+    Utility function to find local low/high field variable coordinates on a contour map. To classify as a local high, the data
+    point must be greater than highVal, and to classify as a local low, the data point must be less than lowVal.
+    Args:
+        da: (:class:`xarray.DataArray`):
+            Xarray data array containing the lat, lon, and field variable (ex. pressure) data values
+        highVal (:class:`int`):
+            Data value that the local high must be greater than to qualify as a "local high" location.
+            Default highVal is a pressure value of 1040 hectopascals.
```
clyne, line 59: Should we be including units here? Does the implementation make use of them in any way? For example, if the data units were in pascals would they be converted to hectopascals for comparison against highVal?

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+            Determines which extrema are being found- minimum or maximum, respectively.
+            Default eType is 'Low'.
+    Returns:
+        clusterExtremas (:class:`list`):
+            List of coordinate tuples in GPS form (lon in degrees, lat in degrees)
+            that specify local low/high locations
+    """
+
+    # Create a 2D array of coordinates in the same shape as the field variable data
+    # so each coordinate is easily mappable to a data value
```
clyne, line 74: I think what you are doing here is creating a 1D array of X,Y tuples, such that you have a coordinate pair for each grid point. Is that correct? Can you claify?

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+    coordarr = []
+    for y in np.array(da.lat):
+        temparr = []
+        for x in np.array(da.lon):
+            temparr.append((x, y))
+        coordarr.append(temparr)
+    coordarr = np.array(coordarr)
+
+    # Set number that a derivative must be less than in order to
```
clyne, line 83: Strictly less than or less than or equal?

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+    # Get global gradient of contour data
+    grad = np.gradient(da.data)
+
+    # Gradient in the x direction
+    arr1 = grad[0]
+
+    # Gradient in the y direction
+    arr2 = grad[1]
+
+    # Get all array 1 indexes where gradient value is between -bound and +bound
```
clyne, line 96: Since bound is equal 0.0, do you need to be checking the interval [-bound .. bound]?

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+                                             np.array([x[0] for x in Clevels]),
+                                             np.array([x[1] for x in Clevels]))
+        transformedClevels = [(x[0], x[1]) for x in clevelpoints]
+        ax.clabel(contours, manual=transformedClevels, inline=True, fontsize=rfs, colors='k', fmt="%.0f")
+        [allLabels.append(txt) for txt in contours.labelTexts]
+        if rHorizontal is True:
+            [txt.set_rotation('horizontal') for txt in contours.labelTexts]
+
+    # Plot any low contour levels
+    if lowClevels != []:
```
clyne, line 257: Could this block of code and the next be made into a function that plots either highs or lows? It seems like code is duplicated, the only real difference being the label that is used.

Plots/MapProjections/NCL_sat_1.py
```diff
@@ -0,0 +1,380 @@
...
+            Contour set that is being labeled.
+        transform (:class:`cartopy._crs`):
+            Instance of CRS that represents the source coordinate system of coordinates.
+            (ex. ccrs.Geodetic()).
+        ax (:class:`matplotlib.pyplot.axis`):
+            Axis containing the contour set.
+        proj (:class:`cartopy.crs`):
+            Projection 'ax' is defined by.
+            This is the instaance of CRS that the coordinates will be transformed to.
+        Clevels (:class:`list`):
```
clyne, line 204: It looks like rather than selecting a contour value and plotting the contour associated with that value, you are picking a location and plotting a contour for the value at that location. Is that correct? The original NCL example looks like it simply plots contours levels from 948 to 1064 in increments of 4.

clairefio: Yes- I did this because for the high and low clabels, the high and low contour values will change depending on how many levels are in the contour set. For example, if the contour set only has levels 948 to 1064 in increments of 4, if you have a high pressure value of 1075 in the data, it will probably print as 1064 because python just chooses the contour closest to the coordinate you are plotting at.

Plots/MapProjections/NCL_sat_1.py
clairefio: Not when the bound is 0... I kept that in there just in case the user wanted to change the bound, but I will take it out.

Plots/MapProjections/NCL_sat_1.py
clairefio: It was suggested to me that I include them just so people would understand the default values, but I will take them out.

Plots/MapProjections/NCL_sat_1.py
clyne: Sorry, I didn't quite follow that explanation. But I think we're going to have to figure out how to make this work using a conventional way of specifying the contours. I.e. specifying the levels, not locations that they pass through. Perhaps you can show me an example of what happens if you try to plot the contours by level so I can better understand what the problem is?

clairefio: Ok actually I think I was misunderstanding- the contour levels are plotted by specifying levels, just like in the NCL script:

using clevels

 I think this variable name is confusing because it actually just means a list of coordinates, it has nothing to do with the contour levels/lines that are being plotted. I will change it! Does this answer your question.

clyne: Changing the name would help for sure. But I think the larger question is what are all of the hard-coded values in regularCLabels? Are you simply attempting to place the contour labels at the exact same locations as in the NCL plot? Is there some reason the automatic selection of labels by MPL is not adequate (other than they will lie in different locations than the original NCL plot)?

clairefio: Yes, I was just trying to place the labels in the same locations as in the NCL plot. I don't think it would make a big difference if I just used automatic MPL labeling instead, do you want me to take the manually placed labels out? And if so, do you also want me to delete the helper function that manually plots non-high/low contour labels using coordinates?

clyne: Now I understand. In general, users are not going to want to have to manually select label locations. However, I can imagine situations when someone wants complete control over where the labels are placed. My suggestion: leave things as they are, but clearly document that this example demonstrates how to manually place contour labels. I realize the description of the plot says something like "manual placement of contour labels", but there is a lot going on in this example. I'd suggest expanding the comment for "regularCLabels" to reenforce that we're manually placing the label locations instead of letting MPL choose the locations for us. Sound good?

clairefio: Yes!

clairefio:
@clyne The helper functions in this PR are now the same as PR #16 and #17 in GeoCAT-viz

PR #112, opened 2020-06-16
xy_16 example

### LINKED ISSUE #111: xy_16
Description: Plotting a vertical profile with an inverted logarithmic y axis

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/xy_16.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/xy_16_1_lg.png
https://www.ncl.ucar.edu/Applications/Images/xy_16_2_lg.png

Category: XY Plots

hCraker:
Closes #111 
image
image

### DIFF
--- a/Plots/XY/NCL_xy_16.py
+++ b/Plots/XY/NCL_xy_16.py
@@ -20,6 +20,7 @@ See following URLs to see the reproduced NCL plot & script:
 import numpy as np
 import xarray as xr
 import matplotlib.pyplot as plt
+from matplotlib.ticker import (ScalarFormatter, NullFormatter)
 
 import geocat.datafiles as gdf
 from geocat.viz import util as gvutil
@@ -42,10 +43,19 @@ U50 = U.isel(lat=49).drop('lat')
 
 # Generate figure (set its size (width, height) in inches) and axes
 plt.figure(figsize=(8, 8))
+ax = plt.axes()
+
+# Format axes
+gvutil.set_axes_limits_and_ticks(ax, xlim=(-20, 40), ylim=(1000, 0), xticks=np.arange(-20,60,10), yticks=np.arange(0,1200,200))
+gvutil.add_major_minor_ticks(ax, x_minor_per_major=5, y_minor_per_major=4, labelsize=14)
+gvutil.set_titles_and_labels(ax, maintitle='Profile Plot', xlabel=U.long_name, ylabel=U['lev'].long_name, labelfontsize=16)
+
+
 
 plt.plot(U20.data, U20.lev, color='black', linestyle='-')
 plt.plot(U30.data, U30.lev, color='black', linestyle='--')
 plt.plot(U40.data, U40.lev, color='black', linestyle=':')
 plt.plot(U50.data, U50.lev, color='black', linestyle='-.')
+
 
 plt.show()
--- a/Plots/XY/NCL_xy_16.py
+++ b/Plots/XY/NCL_xy_16.py
@@ -1,0 +1,38 @@
+"""
+NCL_xy_16.py
+===============
+This script illustrates the following concepts:
+   - Drawing a legend inside an XY plot
+   - Drawing an X reference line in an XY plot
+   - Reversing the Y axis
+   - Using log scaling and explicit labeling
+   - Changing the labels in a legend
+   - Creating a vertical profile plot
+
+See following URLs to see the reproduced NCL plot & script:
+    - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/xy_16.ncl
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/xy_16_1_lg.png
+                         https://www.ncl.ucar.edu/Applications/Images/xy_16_2_lg.png
+"""
+
+###############################################################################
+# Import packages:
+import numpy as np
+import xarray as xr
+import matplotlib.pyplot as plt
+
+import geocat.datafiles as gdf
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Read in data:
+
+# Open a netCDF data file using xarray default engine and load the data into xarrays
+ds = xr.open_dataset(gdf.get("netcdf_files/atmos.nc"), decode_times=False)
+U = ds.U.isel(time=0).drop('time').isel(lon=0).drop('lon')
+print(U)
+# Extract slices of the data at different latitudes
+U20 = U.where(U.lat==20)
+U30 = U.where(U.lat==30)
+U40 = U.where(U.lat==40)
+U50 = U.where(U.lat==50)
--- a/Plots/XY/NCL_xy_16.py
+++ b/Plots/XY/NCL_xy_16.py
@@ -30,9 +30,22 @@ from geocat.viz import util as gvutil
 # Open a netCDF data file using xarray default engine and load the data into xarrays
 ds = xr.open_dataset(gdf.get("netcdf_files/atmos.nc"), decode_times=False)
 U = ds.U.isel(time=0).drop('time').isel(lon=0).drop('lon')
-print(U)
-# Extract slices of the data at different latitudes
-U20 = U.where(U.lat==20)
-U30 = U.where(U.lat==30)
-U40 = U.where(U.lat==40)
-U50 = U.where(U.lat==50)
+
+# Extract slices of the data at different latitudes using the index of the desired value
+U20 = U.isel(lat=39).drop('lat')
+U30 = U.isel(lat=42).drop('lat')
+U40 = U.isel(lat=46).drop('lat')
+U50 = U.isel(lat=49).drop('lat')
+
+###############################################################################
+# Plot:
+
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+
+plt.plot(U20.data, U20.lev, color='black', linestyle='-')
+plt.plot(U30.data, U30.lev, color='black', linestyle='--')
+plt.plot(U40.data, U40.lev, color='black', linestyle=':')
+plt.plot(U50.data, U50.lev, color='black', linestyle='-.')
+
+plt.show()
--- a/Plots/XY/NCL_xy_16.py
+++ b/Plots/XY/NCL_xy_16.py
@@ -56,10 +56,10 @@ gvutil.set_titles_and_labels(ax, maintitle='Profile Plot', xlabel=U.long_name, y
 ax.axvline(x=0, color='black', linewidth=0.5)
 
 # Plot data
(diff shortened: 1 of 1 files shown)


hCraker:
The following warning is thrown when I try to use `gvutil.add_major_minor_ticks()`. 
`UserWarning: AutoMinorLocator does not work with logarithmic scale`

@clairefio @michaelavs Do either of you know how to avoid this warning but still be able to use the utility function? I can simply copy the source code for the function and remove the line that sets the minor tick locator for the y axis, but I'm hoping there is a better solution. If there isn't one, should we consider modifying the function to handle axes with scales other than linear?

cc: @clyne

michaelavs:
> The following warning is thrown when I try to use `gvutil.add_major_minor_ticks()`.
> `UserWarning: AutoMinorLocator does not work with logarithmic scale`
> 
> @clairefio @michaelavs Do either of you know how to avoid this warning but still be able to use the utility function? I can simply copy the source code for the function and remove the line that sets the minor tick locator for the y axis, but I'm hoping there is a better solution. If there isn't one, should we consider modifying the function to handle axes with scales other than linear?
> 
> cc: @clyne

If I am understanding correctly, you have explored just using "regular" code (not a convenience function), correct? From a quick search of the UserWarning I found on [Stack Overflow](https://stackoverflow.com/questions/33126126/matplotlib-minor-ticks) that the normal minor tick locator is only designed for linear scales which is why it would work for the first projection, but doesn't agree with the log projection. It recommends using Log Locator. On first glance, this looks like a similar function but will recognize the log scale, but I haven't had the chance to fully dive into it. See if you can find a "normal" version to use, and if it works out, we may need to think about adding it to the convenience functions.

hCraker:
> If I am understanding correctly, you have explored just using "regular" code (not a convenience function), correct? From a quick search of the UserWarning I found on [Stack Overflow](https://stackoverflow.com/questions/33126126/matplotlib-minor-ticks) that the normal minor tick locator is only designed for linear scales which is why it would work for the first projection, but doesn't agree with the log projection. It recommends using Log Locator. On first glance, this looks like a similar function but will recognize the log scale, but I haven't had the chance to fully dive into it. See if you can find a "normal" version to use, and if it works out, we may need to think about adding it to the convenience functions.

I didn't think of that thanks. I will use matplotlib functions instead of the geocat convenience ones for now. We should certainly make a utility function for log axes in the future.

hCraker:
Updated logarithmic plot:
image

michaelavs:
Looks great so far! Minor changes I found: in line 92 `ax.yaxis.set_minor_locator(tic.LogitLocator(minor=True, nbins=1))` creates an error for the 'nbins' argument of an unexpected keyword argument, by removing 'nbins=1' completely, error is resolved and projection looks the same as yours does here. There is also a RuntimeWarning on my side, but I believe this is from the how vmax is being set because at the top pressure level, you will divide by 0 but I think this is a warning because it doesn't actually happen with this visualization. But as long as the graph is still running correctly, I don't think this is a big problem and probably doesn't need to be addressed past noting it exists.

clyne:
> > If I am understanding correctly, you have explored just using "regular" code (not a convenience function), correct? From a quick search of the UserWarning I found on [Stack Overflow](https://stackoverflow.com/questions/33126126/matplotlib-minor-ticks) that the normal minor tick locator is only designed for linear scales which is why it would work for the first projection, but doesn't agree with the log projection. It recommends using Log Locator. On first glance, this looks like a similar function but will recognize the log scale, but I haven't had the chance to fully dive into it. See if you can find a "normal" version to use, and if it works out, we may need to think about adding it to the convenience functions.
> 
> I didn't think of that thanks. I will use matplotlib functions instead of the geocat convenience ones for now. We should certainly make a utility function for log axes in the future.

Excellent idea. @hCraker could you please create an issue on geocat-viz to add support for this? Thanks!

hCraker:
Note: When this [issue](https://github.com/NCAR/geocat-viz/issues/7) is resolved, this PR should be updated to use the expanded function.

hCraker:
> Looks great so far! Minor changes I found: in line 92 `ax.yaxis.set_minor_locator(tic.LogitLocator(minor=True, nbins=1))` creates an error for the 'nbins' argument of an unexpected keyword argument, by removing 'nbins=1' completely, error is resolved and projection looks the same as yours does here. There is also a RuntimeWarning on my side, but I believe this is from the how vmax is being set because at the top pressure level, you will divide by 0 but I think this is a warning because it doesn't actually happen with this visualization. But as long as the graph is still running correctly, I don't think this is a big problem and probably doesn't need to be addressed past noting it exists.

I believe that I have removed the warning. Thank you for pointing it out.

Plots/XY/NCL_xy_16.py
```diff
@@ -0,0 +1,108 @@
...
+ax.axvline(x=0, color='black', linewidth=0.5)
+
+# Plot data
+plt.plot(U20.data, U20.lev, color='black', linestyle='-', label='20N')
+plt.plot(U30.data, U30.lev, color='black', linestyle='--', label='30N')
+plt.plot(U40.data, U40.lev, color='black', linestyle=':', label='40N')
+plt.plot(U50.data, U50.lev, color='black', linestyle='-.', label='50N')
+
+# Add legend
+plt.legend(loc='center right', frameon=False, fontsize=14, labelspacing=1)
```
clairefio, line 65: I'm not sure if demonstrating this feature will be of any value to the reader, but if you really want to make your graph look like the NCL one, you can reverse the order of the plot list in the legend.

```suggestion
handles, labels = ax.get_legend_handles_labels()
plt.legend(reversed(handles), reversed(labels), loc='center right', frameon=False, fontsize=14, labelspacing=1)
```

Plots/XY/NCL_xy_16.py
```diff
@@ -0,0 +1,108 @@
...
+               top=True, left=True, right=True)
+
+# Plot data
+plt.plot(U20.data, U20.lev, color='black', linestyle='-', label='20N')
+plt.plot(U30.data, U30.lev, color='black', linestyle='--', label='30N')
+plt.plot(U40.data, U40.lev, color='black', linestyle=':', label='40N')
+plt.plot(U50.data, U50.lev, color='black', linestyle='-.', label='50N')
+
+# Add legend
+plt.legend(loc='center right', frameon=False, fontsize=14, labelspacing=1)
```
clairefio, line 106: Same as line 65
```suggestion
handles, labels = ax.get_legend_handles_labels()
plt.legend(reversed(handles), reversed(labels), loc='center right', frameon=False, fontsize=14, labelspacing=1)
```

Plots/XY/NCL_xy_16.py
```diff
@@ -0,0 +1,108 @@
...
+
+###############################################################################
+# Plot with linear y axis:
+
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+ax = plt.axes()
+
+# Format axes limits, ticks, and labels
+gvutil.set_axes_limits_and_ticks(ax, xlim=(-20, 40), ylim=(1000, 0), xticks=np.arange(-20,60,10), yticks=np.arange(0,1200,200))
```
clairefio, line 51: Minor formatting error- add spaces to lists
```suggestion
gvutil.set_axes_limits_and_ticks(ax, xlim=(-20, 40), ylim=(1000, 0), xticks=np.arange(-20, 60, 10), yticks=np.arange(0, 1200, 200))
```

Plots/XY/NCL_xy_16.py
```diff
@@ -0,0 +1,108 @@
...
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+ax = plt.axes()
+
+# Format axes limits, ticks, and labels
+plt.yscale('log')
+ax.yaxis.set_major_formatter(ScalarFormatter())
+ax.yaxis.set_minor_formatter(NullFormatter())
+pressure_lvls = [1, 5, 10, 30, 50, 100, 200, 300, 400, 500, 700, 1000]
+gvutil.set_axes_limits_and_ticks(ax, xlim=(-20, 40), ylim=(1000, 4), xticks=np.arange(-20,60,10), yticks=pressure_lvls)
```
clairefio, line 81: Minor formatting error- add spaces to lists
```suggestion
gvutil.set_axes_limits_and_ticks(ax, xlim=(-20, 40), ylim=(1000, 4), xticks=np.arange(-20, 60, 10), yticks=pressure_lvls)
```

Plots/XY/NCL_xy_16.py
hCraker: I reversed the order and also added comments about how the order for legend elements is decided in commit `a619119`.

hCraker:
> Great work overall! Please play with the line thicknesses within both plots to make them look closer to NCL originals and to increase the readability due to multiple lines.

@erogluorhan the dash lengths are proportional to the line thickness. When I make the lines thinner, the dashes become harder to distinguish. Do you think this is legible enough?
image

erogluorhan:
> @erogluorhan the dash lengths are proportional to the line thickness. When I make the lines thinner, the dashes become harder to distinguish. Do you think this is legible enough or should I go back to what I had before?

That's right if you only adjust `'linewidth'`, but play also with `'dashes'` argument, e.g.:

`linewidth=0.5, dashes=(5, 8)`

Search about it and you will find it useful and meaningful information about it.

hCraker:
> > @erogluorhan the dash lengths are proportional to the line thickness. When I make the lines thinner, the dashes become harder to distinguish. Do you think this is legible enough or should I go back to what I had before?
> 
> That's right if you only adjust `'linewidth'`, but play also with `'dashes'` argument, e.g.:
> 
> `linewidth=0.5, dashes=(5, 8)`
> 
> Search about it and you will find it useful and meaningful information about it.

I was hoping to use this example to show the builtin dash styles provided by matplotlib, but I can make custom ones pretty easily. I'll push the changes soon.

Plots/XY/NCL_xy_16.py
```diff
@@ -0,0 +1,116 @@
...
+U50 = U.isel(lat=49).drop('lat')
+
+###############################################################################
+# Plot with linear y axis:
+
+# Generate figure (set its size (width, height) in inches) and axes
+plt.figure(figsize=(8, 8))
+ax = plt.axes()
+
+# Format axes limits, ticks, and labels
```
erogluorhan, line 50: In case a community user looked first and only at this example script, comments should be self-explanatory, especially about our **geocat-viz** use, so I'd suggest to use comments from other examples for each of these **gvutil** convenience functions.

hCraker: Thanks for catching it. It should be fixed now along with the formatting for the other comments.

Issue #88, opened 2020-06-01
polyg_19


clairefio:
Description: Drawing a map of mainland United States with Alaska, Hawaii, and Puerto Rico added as annotations at the bottom

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/polyg_19.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/polyg_19_lg.png

Category: polygons

clairefio:
Updated plot:
![polyg_19](https://user-images.githubusercontent.com/36773444/85063297-c8ac5200-b15e-11ea-9907-dd4426a9d47f.PNG)

clairefio:
@hCraker @michaelavs A couple things for this PR: 

The PR run keeps failing and I am pretty sure it's because I am pulling data from a shapefile in the same directory as my plot script, and not from geocat-datafiles (the different shapefile is also why this map looks less detailed than the NCL version). This folder is getting added soon, which will hopefully fix the problem.

Additionally, in the NCL plot the tick labels are all ints except for the second '2.5' tick. I'm not sure how to accomplish this in python, or if it's even important, but if either of you have an idea let me know.

hCraker:
> @hCraker @michaelavs A couple things for this PR:
> 
> The PR run keeps failing and I am pretty sure it's because I am pulling data from a shapefile in the same directory as my plot script, and not from geocat-datafiles (the different shapefile is also why this map looks less detailed than the NCL version). This folder is getting added soon, which will hopefully fix the problem.
> 
> Additionally, in the NCL plot the tick labels are all ints except for the second '2.5' tick. I'm not sure how to accomplish this in python, or if it's even important, but if either of you have an idea let me know.

The NCL plot has the same tick values as your plot including a tick mark with a value of 2.5. There is also a subtle color difference between the section of the color bar to the left of 2.5 and to the right of 2.5 although it's so slight that I think we should change it. You appear to be on the right track, so could you clarify what you are wanting help with?

clairefio:
> > @hCraker @michaelavs A couple things for this PR:
> > The PR run keeps failing and I am pretty sure it's because I am pulling data from a shapefile in the same directory as my plot script, and not from geocat-datafiles (the different shapefile is also why this map looks less detailed than the NCL version). This folder is getting added soon, which will hopefully fix the problem.
[...]

PR #114, opened 2020-06-17
Polyg_19

clairefio:
Issue #88 

image


clairefio:
Updated plot:

image

clairefio:
Updated plot:

polyg_19

Plots/Polygons/NCL_polyg_19.py
```diff
@@ -0,0 +1,200 @@
...
+    axis.get_yaxis().set_visible(False)
+
+###############################################################################
+# Define helper function to plot and color each state
+
+
+def plotRegion(shape, axis, xlim):
+
+    # Plot each shape within a region (ex. mainland Alaska and all of it's surrounding Alaskan islands)
+    for i in range(len(shape.shape.parts)):
```
hCraker, line 108: I would recommend refactoring the `shape` argument to either `state` or `region` since you are specifying a particular state when you call this function. The code is confusing with calls like `shape.shape.parts`.

Plots/Polygons/NCL_polyg_19.py
```diff
@@ -0,0 +1,200 @@
...
+file1 = open(gdf.get("shape_files/tl_2017_us_state.dbf"), 'r')
+file2 = open(gdf.get("shape_files/tl_2017_us_state.shp"), 'r')
+file3 = open(gdf.get("shape_files/tl_2017_us_state.shx"), 'r')
+file4 = open(gdf.get("shape_files/tl_2017_us_state.prj"), 'r')
+
+# Open the shapefile
+us = shp.Reader(gdf.get("shape_files/tl_2017_us_state.dbf"))
+
+# Open the text file with the population data
+statepopulationfile = open(gdf.get("ascii_files/us_state_population.txt"), 'r')
```
hCraker, line 43: I would refactor this and other variables in the file to include underscores between the words i.e. `state_population_file` or even shorten it to `state_pop_file`. This would make your code more readable.

Plots/Polygons/NCL_polyg_19.py
```diff
@@ -0,0 +1,200 @@
...
+# Create three inset axes on lower axis for Alaska, Hawaii, and Puerto Rico
+axin1 = ax2.inset_axes([0.0, 0.7, 0.30, 0.80], frameon=False)
+removeTicks(axin1)
+axin2 = ax2.inset_axes([0.40, 0.7, 0.20, 0.40], frameon=False)
+removeTicks(axin2)
+axin3 = ax2.inset_axes([0.70, 0.7, 0.30, 0.30], frameon=False)
+removeTicks(axin3)
+
+# Specify which shapes not to include when plotting the mainland U.S.
+nonmainlandus = ['Commonwealth of the Northern Mariana Islands', 'Guam', 'Puerto Rico', 'Hawaii', 'Alaska', 'American Samoa',
```
hCraker, line 165: I would refactor this variable. It took a second before I could tell what it was trying to say. `non_mainland_us` might be better

Plots/Polygons/NCL_polyg_19.py
```diff
@@ -0,0 +1,200 @@
...
+                 'United States Virgin Islands']
+
+# Get population of each state
+nameandpopdict = getStatePopulations(statepopulationfile)
```
hCraker, line 169: You've used this variable name in a few places, and it's hard to read. Again I would separate the words using underscores

hCraker:
@clairefio Are there more detailed shape files, or are the ones you are using the best you have? If you don't have higher resolution files, could you use cartopy to draw this instead?

michaelavs:
Looks great so far! I was wondering if we know why Wisconsin/Michigan/two of the Great Lakes are merging together? I would assume it may have something to do with boundaries in the shape files and them just not existing correctly(?). Its very weird it isn't doing that with Ohio/Pennsylvania/Lake Erie, I wonder if what Rose was saying with more detailed shape files and/or using Cartopy may help resolve this. I would be interested to hear what you find out, and will see if I can find anything too!

clairefio:
> @clairefio Are there more detailed shape files, or are the ones you are using the best you have? If you don't have higher resolution files, could you use cartopy to draw this instead?

Yes, I found a different, more-defined shapefile so I will be using that one!

clairefio:
> Looks great so far! I was wondering if we know why Wisconsin/Michigan/two of the Great Lakes are merging together? I would assume it may have something to do with boundaries in the shape files and them just not existing correctly(?). Its very weird it isn't doing that with Ohio/Pennsylvania/Lake Erie, I wonder if what Rose was saying with more detailed shape files and/or using Cartopy may help resolve this. I would be interested to hear what you find out, and will see if I can find anything too!

I found a different, more defined shapefile but it is still doing that which is very odd. I might have to find out what that region specifically is called and then include a condition so it doesn't get plotted/filled.

hCraker:
@clairefio Only parts of the Great Lakes are within US borders. It appears that the shapes for the states extend all the way to the national border which includes large areas of water. If you take a look at google maps, you can see the difference in shape between the states' coasts and the interstate and international borders. The discrepancy is most noticeable for Wisconsin and Michigan, but the coasts of NE Minnesota, NE Illinois, NW Indiana, N Ohio, NW Pennsylvania, and W New York are also altered. 

I recommend trying to find a shape file that accounts for the lake fronts in each of these states. Alternatively, you could find a shape file for the lakes, color those areas white, and overlay it onto the state shapes. This is "hacky" but it could work.

clairefio:
> @clairefio Only parts of the Great Lakes are within US borders. It appears that the shapes for the states extend all the way to the national border which includes large areas of water. If you take a look at google maps, you can see the difference in shape between the states' coasts and the interstate and international borders. The discrepancy is most noticeable for Wisconsin and Michigan, but the coasts of NE Minnesota, NE Illinois, NW Indiana, N Ohio, NW Pennsylvania, and W New York are also altered.
> 
> I recommend trying to find a shape file that accounts for the lake fronts in each of these states. Alternatively, you could find a shape file for the lakes, color those areas white, and overlay it onto the state shapes. This is "hacky" but it could work.
> 
> EDIT: I will mention that your shape file is _technically_ correct in terms of the area of land/water each state owns, but aesthetically it's not what you'd expect.

ahh, this makes sense. I think I might be able to make it show up correctly with my new shapefile, but if not I might try the second solution you mentioned. Thanks!

clairefio:
Updated plot:

polyg_19

There is a faint outline of the border of the great lakes, I have tried many things but I cannot seem to remove this.

hCraker:
That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to either prevent the border from being drawn.

clairefio:
> That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to prevent the border from being drawn.

No, it's not a separate shape unfortunately, those lines are the border of each state. So if I wanted to make the line white/thinner next to the great lakes, that formatting would also apply to the lines between the states as well.

hCraker:
> > That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to prevent the border from being drawn.
> 
> No, it's not a separate shape unfortunately, those lines are the border of each state. So if I wanted to make the line white/thinner next to the great lakes, that formatting would also apply to the lines between the states as well.

Then I'm not sure what you can do. I don't have much experience with shape files. @michaelavs any thoughts?

clairefio:
> > > That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to prevent the border from being drawn.
> > 
> > 
> > No, it's not a separate shape unfortunately, those lines are the border of each state. So if I wanted to make the line white/thinner next to the great lakes, that formatting would also apply to the lines between the states as well.
> 
> Then I'm not sure what you can do. I don't have much experience with shape files. @michaelavs any thoughts?

When I increase the linewidth of that particular lake shape and make the zorder higher than the state it is over, the lake covers up the lines. However, the lake shape becomes less defined. Do you think this is a worthwhile trade-off or should I keep it how it was before?

polyg_19

michaelavs:
> > > That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to prevent the border from being drawn.
> > 
> > 
> > No, it's not a separate shape unfortunately, those lines are the border of each state. So if I wanted to make the line white/thinner next to the great lakes, that formatting would also apply to the lines between the states as well.
> 
> Then I'm not sure what you can do. I don't have much experience with shape files. @michaelavs any thoughts?

I have maybe used shapefiles in a very limited sense in QGIS, but not with Python. From my understanding of them, it's a shape around data which is why they're used for population statistics (or similar types of data). That being said, I did try using the shapefile.fields command with this script to see if the field names would be any help, but unfortunately they are pretty cryptically named. My next thought would be to maybe find a shape file you could layer over the top of these that only has the regions data in it. But not sure if that you get rid of the the problem we are seeing with border lines. In that same degree, would there be any capability to use cartopy to add border lines and just reduce the ones from the shapefile to zero?

clairefio:
> > > > That line is the US/Canada border. Is the international border specified as a separate shape in the shape file? You may be able to use that fact to prevent the border from being drawn.
> > > 
> > > 
> > > No, it's not a separate shape unfortunately, those lines are the border of each state. So if I wanted to make the line white/thinner next to the great lakes, that formatting would also apply to the lines between the states as well.
> > 
> > 
> > Then I'm not sure what you can do. I don't have much experience with shape files. @michaelavs any thoughts?
> 
> I have maybe used shapefiles in a very limited sense in QGIS, but not with Python. From my understanding of them, it's a shape around data which is why they're used for population statistics (or similar types of data). That being said, I did try using the shapefile.fields command with this script to see if the field names would be any help, but unfortunately they are pretty cryptically named. My next thought would be to maybe find a shape file you could layer over the top of these that only has the regions data in it. But not sure if that you get rid of the the problem we are seeing with border lines. In that same degree, would there be any capability to use cartopy to add border lines and just reduce the ones from the shapefile to zero?

This is a good idea, but I don't think you can use cartopy borders unless you are plotting them over an axis defined by a cartopy projection. I'll try it and see what happens!

Plots/Polygons/NCL_polyg_19.py
```diff
@@ -0,0 +1,233 @@
+"""
+NCL_polyg_19.py
+===============
+This script illustrates the following concepts:
+   - Adding lines and polygons to a map
+   - Adding a map to another map as an annotation
+   - Coloring shapefile outlines based on an array of values
+   - Drawing a custom labelbar on a map
```
erogluorhan, line 8: Use "colorbar" to replace "labelbar", which is from NCL terminology

Issue #113, opened 2020-06-16
xy_4


hCraker:
Description: Create scatter plot with custom markers

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/xy_4.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/xy_4_1_lg.png
https://www.ncl.ucar.edu/Applications/Images/xy_4_2_lg.png

Category: XY Plots

hCraker:
This issue wasn't properly linked to the PR that was supposed to close it. That PR has been merged.

PR #116, opened 2020-06-22
xy_4

hCraker:
Closees #113 
image
image

These plots match the NCL example plots; however, I don't think this is a good representation of all the options available in matplotlib to customize plot markers. I'm thinking of using the second plot to show the various ways to make markers though `matplotlib.markers`, unicode symbols, math symbols, etc.


hCraker:
Added multiple different methods of drawing markers. This is what the second plot looks like now.
image

Plots/XY/NCL_xy_4.py
```diff
@@ -0,0 +1,112 @@
...
+# Add legend
+plt.legend()
+
+# Use geocat.viz.util convenience function to set titles and labels
+gvutil.set_titles_and_labels(ax, maintitle="Make your own marker", xlabel=t['time'].long_name, ylabel=t.long_name)
+
+# Use geocat.viz.util convenience function to add minor and major tick lines
+gvutil.add_major_minor_ticks(ax, x_minor_per_major=5, y_minor_per_major=4, labelsize=14)
+
+gvutil.set_axes_limits_and_ticks(ax, xlim=(131155, 131190),
```
clyne, line 107: Can we get the time info from the file rather than hard-coding it here?

hCraker: The minimum time is 131157 and the maximum time is 131187. I could retrieve that data and the round to the nearest multiple of 5 without much issue.

Do you also want me to generalize the `xticklabels`? Only every other tick label should be drawn so I figured this was the simplest solution.

clyne: Automatically picking nice, round-number tic labels is not the point of this example. On the other hand the NCL version seems to be doing a reasonable job here. Perhaps some sort of convenience function for geocat-viz is in order, but I'm not sure its a must do. I'm going to approve the PR and leave the choice up to you :-)

Issue #120, opened 2020-06-24
Polyg_14 Example


michaelavs:
Link to NCL code: https://www.ncl.ucar.edu/Applications/Scripts/polyg_14.ncl
Link to NCL visualizations: https://www.ncl.ucar.edu/Applications/Images/polyg_14_2_lg.png
https://www.ncl.ucar.edu/Applications/Images/polyg_14_1_lg.png

PR #122, opened 2020-06-24
Polyg 14

michaelavs:
Closes [#120](https://github.com/NCAR/GeoCAT-examples/issues/120)

*Outdated branch of [new](https://github.com/NCAR/GeoCAT-examples/pull/140) version of PR


michaelavs:
I found a way to add specific points to the trajectory line in this code (without having predetermined projection paths from a data set), but it requires I used a package that isn't a part of the geocat environment. Have either of you done anything similar? either with using packages not in the environment or know a way to define where the markers will be along the trajectory? I have used matplotlib functions, but this will only allow for a straight path with predetermined markers. I have also experimented with the marker type but that didn't fully project what I wanted it to. @clairefio @hCraker 
image
image

hCraker:
@michaelavs Have you looked at [this](https://scitools.org.uk/cartopy/docs/latest/cartopy/geodesic.html) page on cartopy? There are functions that can do some of the geodesic calculations in cartopy already, so you may not need the other package.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+###############################################################################
+# Read in data:
+###############################################
+
+def Plot(color, row, col, pos, title, style, pt):
+
+    fig = plt.figure(figsize=(12,12))
+    ax = fig.add_subplot(row, col, pos, projection=ccrs.PlateCarree())
+    
+    ax.set_extent([-125,-60,15,65], ccrs.PlateCarree())
```
hCraker, line 21: I recommend having these coordinates as arguments for this function instead of hard coded.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+    ax.add_feature(cfeature.LAND, color='lightgrey')
+    
+    plt.plot([-120, -64], [20, 60], style, color= color,  transform=ccrs.Geodetic())
```
hCraker, line 24: same as above

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+    
+    # start location
+    lon_fr = -120
+    lat_fr = 20
+    # end location
+    lon_to = -64
+    lat_to = 60
```
hCraker, line 31: Same as above

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+    # WGS84 ellipsoid is used
+    gl = Geodesic.WGS84.InverseLine(lat_fr, lon_fr, lat_to, lon_to)
+    
+    num_points = 10  #for points on geodesic
+    print("distance latitude longitude azimuth")
+    
+    # Compute points on the geodesic, and plot them as red dots
+    # gl.s13 is the total length of the geodesic
+    # the points are equally spaced by true distances, but not on the map
+    # due to the projection distortion
```
hCraker, line 43: Can you clarify this? The points are equally spaced using true distances, which makes sense. The second part "but not on the map due to the projection distortion" could use clarification.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+    # start location
+    lon_fr = -120
+    lat_fr = 20
+    # end location
+    lon_to = -64
+    lat_to = 60
+    
+    # This gets geodesic between the two points
+    # WGS84 ellipsoid is used
+    gl = Geodesic.WGS84.InverseLine(lat_fr, lon_fr, lat_to, lon_to)
```
hCraker, line 35: `cartopy.geodesic.Geodesic.inverse()` may be able to replace this line. That way you don't have to use a package that isn't in the geocat-examples environment

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,85 @@
...
+    
+    num_points = 10  #for points on geodesic
+    print("distance latitude longitude azimuth")
+    
+    # Compute points on the geodesic, and plot them as red dots
+    # gl.s13 is the total length of the geodesic
+    # the points are equally spaced by true distances, but not on the map
+    # due to the projection distortion
+    for ea in np.linspace(0, gl.s13, num_points):
+        g = gl.Position(ea, Geodesic.STANDARD | Geodesic.LONG_UNROLL)
```
hCraker, line 45: I don't see any functionality in cartopy that can easily do what you're trying to do here. We may have to add `geographiclib` to the environment if there is no work around.
cc: @clyne

clyne: geographiclib is distributed via conda. So I don't see any problem with making use of it. Checking with @erogluorhan

erogluorhan: I agree with @clyne as long as @michaelavs adds geographiclib as a dependency to the conda environment yml file within this PR.

erogluorhan:
There is already a directory called "Plots/Trajectories", so is there a specific reason for this PR to introduce a new folder called "Trajectory"?

michaelavs:
> There is already a directory called "Plots/Trajectories", so is there a specific reason for this PR to introduce a new folder called "Trajectory"?

I think my original branch did not have the trajectories directory so I had to create the Trajectory directory for myself. Because the directory already exists as "Trajectories", I'll go ahead and rename it on my side.

michaelavs:
This PR is going to be closed to resolve and issue on my side with pushing commits. A new PR (#140) now encompasses this PR

erogluorhan:
> > There is already a directory called "Plots/Trajectories", so is there a specific reason for this PR to introduce a new folder called "Trajectory"?
> 
> I think my original branch did not have the trajectories directory so I had to create the Trajectory directory for myself. Because the directory already exists as "Trajectories", I'll go ahead and rename it on my side.

Not just for this one, but for in general, you can always `rebase` your local feature branch to the latest of `master` branch to keep your local up-to-date with changes in `master`. This also reduces the number of possible conflicts at the time of merge.

Plots/Trajectory/NCL_polyg_14.py
michaelavs: @erogluorhan should I just push the conda_environment.yml with the geogrpahiclib dependency to the newer PR of this one? Or do I need to create a new PR for changing the yml?

erogluorhan: Yes, it would be good enough to push the `conda_environment.yml` with the `geographiclib` dependency to the newer PR of this one without a need of creating another new PR.

PR #128, opened 2020-06-30
Box 3

clairefio:
Matplotlib does not support changing the color of individual box whiskers/caps, so I left those black.

box_3


Plots/Boxplots/NCL_box_3.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Generate fake data
+
+np.random.seed(200)
+data = np.random.lognormal(size=(40, 3), mean=1, sigma=.7)
```
hCraker, line 27: Was there a reason for randomly generating data instead of specifying it manually like in the NCL script? This works so I don't think you have to change it. I'm just curious

Plots/Boxplots/NCL_box_3.py
```diff
@@ -0,0 +1,116 @@
...
+
+###############################################################################
+# Helper function to set edge color of boxes
+
+
+def setBoxColor(boxplot, number, edge_color):
+
+    # Set edge color of the outside of the boxes and the median lines
+    for element in ['boxes', 'medians']:
+        plt.setp(boxplot[element][number], color=edge_color)
```
hCraker, line 39: [This stackoverflow post](https://stackoverflow.com/questions/41997493/python-matplotlib-boxplot-color) talks about coloring box and whisker plots. You may be able to refer to this to figure out how to color the whiskers.

Plots/Boxplots/NCL_box_3.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Turn both major and minor ticks in overlayed axis off
+ax2.tick_params(which='both', top=False, bottom=False, left=False, right=False,
+                labelleft=False, labelbottom=False)
+
+# Get rid of right and top axis spines
+removeSpines(ax2)
+
+# Plot red x markers
+ax2.scatter(1, 7.7, marker='x', color='red', linewidth=.5, s=100)
```
hCraker, line 107: Are these Xs supposed to represent the outliers or something else that is statistically significant? If these are representative of outliers, `plt.boxplot` should take care of them for you. You shouldn't have to manually plot them.

Plots/Boxplots/NCL_box_3.py
clairefio: No reason!

Plots/Boxplots/NCL_box_3.py
clairefio: I looked at that while I was trying to find a solution, but in all of their examples, every box in each "boxplots" call uses the same colors. When I try to do "for element in ['whiskers', 'caps']" the coloring turns out weird and mismatched, (Like the upper whisker on one box is green and the lower one is blue).

Plots/Boxplots/NCL_box_3.py
clairefio: It looks like in the NCl script the coordinates of each polymarker are manually set, so I did the same. I'm not sure how common it is to want to manually plot markers on a boxplot though, I feel like it would be more useful to plot a statistically significant point on there instead.

hCraker: After looking at the NCL script, it does seem that those points are arbitrary. I wouldn't be too concerned with it then

Plots/Boxplots/NCL_box_3.py
hCraker: Have you considered plotting each box separately and specifying the different colors that way?

clairefio: Figured out how to access each box's actual whiskers and caps, so no need!

clairefio:
Updated plot:

box_3

PR #131, opened 2020-07-01
panel_13

### LINKED ISSUE #130: panel_13
Description: An example of paneling two plots each with filled contours and vectors

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/panel_13.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/panel_13_lg.png

Category: Panel

hCraker:
Closes #130 
image

### DIFF
--- a/Plots/Panels/NCL_panel_13.py
+++ b/Plots/Panels/NCL_panel_13.py
@@ -42,7 +42,8 @@ magnitude = np.sqrt(U.data**2 + V.data**2)
 
 # Create sublots and specify their projections
 projection = ccrs.PlateCarree()
-fig, axs = plt.subplots(2, 1, figsize=(8, 10), subplot_kw={"projection": projection}, gridspec_kw={'hspace': 0.3})
+fig, axs = plt.subplots(2, 1, figsize=(6, 10), subplot_kw={"projection": projection})
+plt.tight_layout(pad=4, h_pad=-8)
 
 # Add coastlines
 axs[0].coastlines(linewidth=0.5)
@@ -80,10 +81,10 @@ wind_ticks = np.arange(-12, 40, 4)
 
 # Plot filled contours
 speed = axs[0].contourf(U['lon'], U['lat'], magnitude, levels=speed_levels, cmap=newcmap)
-speed_cbar = plt.colorbar(speed, ax=axs[0], orientation='horizontal', ticks=speed_ticks, shrink=0.75, drawedges=True)
+speed_cbar = plt.colorbar(speed, ax=axs[0], orientation='horizontal', ticks=speed_ticks, shrink=0.8, drawedges=True, pad=0.1)
 
 wind = axs[1].contourf(U['lon'], U['lat'], U.data, levels=wind_levels, cmap=newcmap)
-plt.colorbar(wind, ax=axs[1], orientation='horizontal', ticks=wind_ticks, shrink=0.75, drawedges=True)
+plt.colorbar(wind, ax=axs[1], orientation='horizontal', ticks=wind_ticks, shrink=0.8, drawedges=True, pad=0.1)
 
 # Remove trailing zeros from speed color bar tick labels
 speed_cbar.ax.xaxis.set_major_formatter(FormatStrFormatter('%g'))
--- a/Plots/Panels/NCL_panel_13.py
+++ b/Plots/Panels/NCL_panel_13.py
@@ -49,8 +49,8 @@ magnitude = np.sqrt(U.data**2 + V.data**2)
 
 # Create sublots and specify their projections
 projection = ccrs.PlateCarree()
-fig, axs = plt.subplots(2, 1, figsize=(6, 10), subplot_kw={"projection": projection})
-plt.tight_layout(pad=4, h_pad=-8)
+fig, axs = plt.subplots(2, 1, figsize=(7, 10), subplot_kw={"projection": projection})
+plt.tight_layout(pad=4, h_pad=-5)
 
 # Add coastlines
 axs[0].coastlines(linewidth=0.5, zorder=1)
--- a/Plots/Panels/NCL_panel_13.py
+++ b/Plots/Panels/NCL_panel_13.py
@@ -107,12 +107,14 @@ quiver_speed = axs[0].quiver(U['lon'], U['lat'], U.data, V.data, width=0.0025, h
 quiver_wind = axs[1].quiver(U['lon'], U['lat'], U.data, V.data, width=0.0025, headwidth=4, zorder=2)
 
 # Add white box to go behind reference vector
-axs[0].add_patch(mpatches.Rectangle(xy=[0.9, 0], width=0.1, height=0.175,
+axs[0].add_patch(mpatches.Rectangle(xy=[0.775, 0], width=0.225, height=0.2,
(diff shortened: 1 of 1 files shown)


Plots/Panels/NCL_panel_13.py
```diff
@@ -0,0 +1,113 @@
...
+                                 yticks=np.arange(-90, 91, 30))
+gvutil.set_axes_limits_and_ticks(axs[1], xlim=[-180, 180], ylim=[-90, 90], xticks=np.arange(-180, 181, 30),
+                                 yticks=np.arange(-90, 91, 30))
+
+# Use geocat.viz.util convenience function to add minor and major tick lines
+gvutil.add_major_minor_ticks(axs[0])
+gvutil.add_major_minor_ticks(axs[1])
+
+# Use geocat.viz.util convenience function to make plots look like NCL plots by using latitude, longitude tick labels
+gvutil.add_lat_lon_ticklabels(axs[0])
```
clairefio, line 70: Remove the degree symbol from each tick mark label to match the NCL plot.

Plots/Panels/NCL_panel_13.py
```diff
@@ -0,0 +1,113 @@
...
+
+wind = axs[1].contourf(U['lon'], U['lat'], U.data, levels=wind_levels, cmap=newcmap, zorder=0)
+plt.colorbar(wind, ax=axs[1], orientation='horizontal', ticks=wind_ticks, shrink=0.8, drawedges=True, pad=0.1)
+
+# Remove trailing zeros from speed color bar tick labels
+speed_cbar.ax.xaxis.set_major_formatter(FormatStrFormatter('%g'))
+
+
+# Plotting vector field
+quiver_speed = axs[0].quiver(U['lon'], U['lat'], U.data, V.data, zorder=2)
```
clairefio, line 101: I would add a "width" parameter to make the vectors thinner to match the NCL plot. Also, are your vectors scaled differently than the ones in the NCL version?

Plots/Panels/NCL_panel_13.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Add white box to go behind reference vector
+axs[0].add_patch(mpatches.Rectangle(xy=[0.9, 0], width=0.1, height=0.175,
+             facecolor='white', transform=axs[0].transAxes, zorder=2))
+axs[1].add_patch(mpatches.Rectangle(xy=[0.9, 0], width=0.1, height=0.175,
+             facecolor='white', transform=axs[1].transAxes, zorder=2))
+# Add reference vector
+axs[0].quiverkey(quiver_speed, 0.95, 0.05, 20, 20, zorder=2)
+axs[1].quiverkey(quiver_wind, 0.95, 0.05, 20, 20, zorder=2)
+
```
clairefio, line 112: Add a textbox to the rectangle patch with the words "Reference vector" underneath the vector.

Plots/Panels/NCL_panel_13.py
hCraker: NCL scales vectors differently than matplotlib does. This was also noticed in #99.

Plots/Panels/NCL_panel_13.py
hCraker: It's fixed now

Plots/Panels/NCL_panel_13.py
hCraker: I just reduced the width and adjusted the arrow proportions. Let me know if you think they need to be any thinner.

Plots/Panels/NCL_panel_13.py
hCraker: I've fixed this now

hCraker:
image

hCraker:
I tried to make the arrows more legible. Is this good @clyne?
image

clyne:
That's better, but some of the arrows are still pretty much blobs unless you zoom in a lot. I see what the problem is; if you set the scale factor to something like 200 the smallest arrows are visible, but then the others are too long. I believe NCL has a notion of minimum arrow size, and I think @michaelavs or @clairefio may have run into this. I'm not sure if there is a simple solution, but perhaps they have some thoughts. If there is not a simple workaround, how about adding a comment that some arrows aren't legible, and opening an issue on geocat-viz to provide a means to set a minimum arrow size?

hCraker:
@clyne I have found a simple way to set minimum lengths using some trig. It works fine, but I'm afraid it is misleading because I had to manipulate the data to produce the desired lengths. Is this best practice? I could add a comment to the documentation explaining how this may not be a desirable feature when trying to make truly accurate plots.
image

jukent:
> @clyne I have found a simple way to set minimum lengths using some trig. It works fine, but I'm afraid it is misleading because I had to manipulate the data to produce the desired lengths. Is this best practice? I could add a comment to the documentation explaining how this may not be a desirable feature when trying to make truly accurate plots.

What type of data manipulation are you referring to?

hCraker:
> > @clyne I have found a simple way to set minimum lengths using some trig. It works fine, but I'm afraid it is misleading because I had to manipulate the data to produce the desired lengths. Is this best practice? I could add a comment to the documentation explaining how this may not be a desirable feature when trying to make truly accurate plots.
> 
> What type of data manipulation are you referring to?

I'm searching through the data for any wind vector with a magnitude less than `minLength` and then changing those vectors' components to make their magnitude equal to `minLength`

jukent:
> > > @clyne I have found a simple way to set minimum lengths using some trig. It works fine, but I'm afraid it is misleading because I had to manipulate the data to produce the desired lengths. Is this best practice? I could add a comment to the documentation explaining how this may not be a desirable feature when trying to make truly accurate plots.
> > 
> > 
> > What type of data manipulation are you referring to?
> 
> I'm searching through the data for any wind vector with a magnitude less than `minLength` and then changing those vectors' components to make their magnitude equal to `minLength`

It might be more honest to not display wind vectors below that minimum length than to change their size, but an argument could be made for still displaying wind direction. I would feel better about this if it were somehow indicated in a key (size X represents wind of value X and below, for example). Thoughts @clyne ?

clyne:
> > I'm searching through the data for any wind vector with a magnitude less than `minLength` and then changing those vectors' components to make their magnitude equal to `minLength`
> 
> It might be more honest to not display wind vectors below that minimum length than to change their size, but an argument could be made for still displaying wind direction. I would feel better about this if it were somehow indicated in a key (size X represents wind of value X and below, for example). Thoughts @clyne ?

I don't like the idea of misrepresenting the arrow length without clearly indicating in the plot that not all arrows are drawn to scale. Ideally, a different glyph would be drawn to indicate this. I don't think we're going to solve this problem with the time remaining and suggest falling back to the illegible glyphs and a comment in the documentation for the plot.

hCraker:
@clyne @jukent I have opened an issue regarding this discussion so we remember to fix this later. Please let me know if the added comment in the latest commit needs clarification.

clyne:
Thanks, Rose.

clyne:
> Thanks, Rose.

Rose, I don't see a comment in the example about the glyph legibility issue.

hCraker:
> Rose, I don't see a comment in the example about the glyph legibility issue.

I was having issues with git, but I just pushed the change. Sorry about that

PR #138, opened 2020-07-06
Sat 2

clairefio:
sat_2


Plots/Satellites/NCL_sat_2.py
```diff
@@ -0,0 +1,436 @@
+"""
+NCL_sat_2.py
+===============
+This script illustrates the following concepts:
+   - unpacking 'short' data
```
hCraker, line 5:
```suggestion
   - Unpacking 'short' data
```
Capitalize the U

Plots/Satellites/NCL_sat_2.py
```diff
@@ -0,0 +1,436 @@
...
+
+###############################################################################
+# Read in data:
+
+# Open a netCDF data file using xarray default engine and
+# load the data into xarrays
+ds = xr.open_dataset(gdf.get("netcdf_files/slp.1963.nc"), decode_times=False)
+
+# Get data from the 24th timestep
+pressure = ds.slp[21, :, :]
```
hCraker, line 37: This comment says the 24th time step but you're using what looks like the 21st time step. Can you change the comment or the code here so they match?

Plots/Satellites/NCL_sat_2.py
```diff
@@ -0,0 +1,436 @@
...
+fig = plt.figure(figsize=(8, 8))
+
+# Set global axes with an orthographic projection
+proj = ccrs.Orthographic(central_longitude=270, central_latitude=45)
+ax = plt.axes(projection=proj)
+ax.set_global()
+
+# Add land, coastlines, and ocean features
+ax.add_feature(cfeature.LAND, facecolor='lightgray', zorder=1)
+ax.add_feature(cfeature.COASTLINE, linewidth=.3)
```
hCraker, line 319: The coastlines and borders are getting covered up by the contours. Can you change the `zorder` for these features to fix this?

Plots/Satellites/NCL_sat_2.py
```diff
@@ -0,0 +1,436 @@
...
+ax.add_feature(cfeature.OCEAN, facecolor='white')
+ax.add_feature(cfeature.BORDERS, linewidth=.3)
+ax.add_feature(cfeature.LAKES, facecolor='white',
+               edgecolor='k', linewidth=.3)
+
+# Make array of the contour levels that will be plotted
+contours = np.arange(948, 1060, 4)
+contours = np.append(contours, 975)
```
hCraker, line 327: Do you still need to explicitly add this contour level? There aren't any low pressure contours that need to be drawn with a value of 975.
Additionally, would it be possible to add to the helper methods a way to draw the contours that aren't already drawn if they need a high or low pressure label? That way you wouldn't have to manually append the value to this array.

clairefio:
> This is a pretty good start. Please see my inline comments. In addition, I noticed that the low pressure near the Bering Strait is labeled as 975 instead of 972. Also the high pressure system over Greenland isn't labeled at all in your plot. See if you can figure out why that is happening and you should be good to go.

I think the problem is only having 30 contour levels plotted on the map. The more contour levels I plot (ex. 70 rather than 30), the closer the Low and High label values get to the ones in the NCL map. When I only have 30 levels, even if I try to manually click to label the areas shown, the label value will still be off because that is the value of the closest contour.

Also, the high-pressure area over Greenland isn't labeled because when the zeroes in my function are k-means clustered, they are done so by proximity. because the High-pressure system to the left is so large, that when I increase the number of k-means clusters that system is split into two before the Greenland area is even recognized as a cluster. I added a helper function where the user can manually add high, low, and regular pressure contours.

clairefio:
sat_2

hCraker:
> I think the problem is only having 30 contour levels plotted on the map. The more contour levels I plot (ex. 70 rather than 30), the closer the Low and High label values get to the ones in the NCL map. When I only have 30 levels, even if I try to manually click to label the areas shown, the label value will still be off because that is the value of the closest contour.
> 
> Also, the high-pressure area over Greenland isn't labeled because when the zeroes in my function are k-means clustered, they are done so by proximity. because the High-pressure system to the left is so large, that when I increase the number of k-means clusters that system is split into two before the Greenland area is even recognized as a cluster. I added a helper function where the user can manually add high, low, and regular pressure contours.

Have you considered using your current method of finding high and low pressure values but adding text to the figure using `plt.text` instead of `clabels`? For example, by using the latitude and longitude coordinates associated with a high/low pressure area and then transforming them to the axes coordinate system via `transform=ax.transAxes` your label might be more accurate.

In addition, I noticed you made a commit attempting to force the high and low labels to be drawn horizontally. `plt.text` would allow you to do that pretty easily via the `rotation` kwarg.

Plots/Satellites/NCL_sat_2.py
```diff
@@ -0,0 +1,516 @@
...
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/sat_2_lg.png
+"""
+
+###############################################################################
+# Import packages:
+import xarray as xr
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import numpy as np
+from sklearn.cluster import KMeans, DBSCAN
```
michaelavs, line 21:
```suggestion

```
This line is unused in the code. Otherwise, looking good so far!

michaelavs, line 21:
```suggestion
from sklearn.cluster import DBSCAN
```
KMeans is unused in the code. Also, will this package need to be added to the conda_environment.yml?

clairefio:
Updated plot:
sat_2

erogluorhan:
This PR cannot pass the build checks likely because **scikit-learn** dependency should have been added to `conda_environment.yml` as `scikit-learn `instead of `sklearn`.

clairefio:
@michaelavs would you mind reviewing this one?

clairefio:
@michaelavs can you review this one?

Plots/MapProjections/NCL_sat_2.py
```diff
@@ -0,0 +1,371 @@
+"""
+NCL_sat_2.py
+===============
+This script illustrates the following concepts:
+   - Unpacking 'short' data
```
clyne, line 5: Is this comment relevant for the Python version of this plot?

Plots/MapProjections/NCL_sat_2.py
```diff
@@ -0,0 +1,371 @@
...
+# low pressure contour levels- these will be plotted
+# as a subscript to an 'L' symbol.
+lowClevels = findLocalExtrema(pressure, lowVal=995, eType='Low')
+highClevels = findLocalExtrema(pressure, highVal=1042, eType='High')
+
+# regular pressure contour levels- These values were found by setting
+# 'manual' argument in ax.clabel call to 'True' and then hovering mouse
+# over desired location of countour label to find coordinate
+# (which can be found in bottom left of figure window).
+clevels = [(-145.27, 50.9), (-125.89, 32.33), (-112.62, 19.89),
```
clyne, line 335: Since this plot is so similar to sat_1, and we already have an example of manual label positioning, how about we let MPL pick the label locations automatically for this plot?

Plots/MapProjections/NCL_sat_2.py
clairefio: I do change the float data into 'short' data, so I just changed the comment to "Converting float data into short data".

Plots/MapProjections/NCL_sat_2.py
clairefio: ok, I'll change it to automatic labeling.

clairefio:
> This looks great! I am approving it despite agreeing with @clyne 's inline comments. So, please address them as well as copy the latest versions of `plotCLabels` and `plotELabels` from **geocat-viz** repo as there was some clean-up lately on them.

Ok! I updated the helper function and made the inline comment suggestions.

PR #140, opened 2020-07-06
Polyg_14

michaelavs:
Closes [#122](https://github.com/NCAR/GeoCAT-examples/issues/120)
*Updated branch for [previous](https://github.com/NCAR/GeoCAT-examples/pull/122) version of this PR
image
image


Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
+"""
+NCL_polyg_14.py
+================
+This script illustrates the following concepts:
+    - Drawing polylines and markers using great circle paths
+    - Using gc_latlon to calculate a great circle path
```
hCraker, line 6: Update your comments to reflect what this example does. `gc_latlon` is from NCL

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
...
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+#Plot
+
+def Plot(color, ext, xext, yext, title, style, pt):
+
+    plt.figure(figsize=(8,8))
+    ax = plt.axes( projection=ccrs.PlateCarree())
```
hCraker, line 32:
```suggestion
    ax = plt.axes(projection=ccrs.PlateCarree())
```
Removing the extra white space here. There are a few other formatting things that could be changed, but I'm not going to add a comment for all of them. I'd recommend using `flake8` to catch them all.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
...
+import matplotlib.pyplot as plt
+import cartopy.feature as cfeature
+from geographiclib.geodesic import Geodesic
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+#Plot
+
+def Plot(color, ext, xext, yext, title, style, pt):
```
hCraker, line 29: I recommend adding a comment explaining what each argument is and what this function does.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
...
+    lon_fr = xext[0]
+    lat_fr = yext[0]
+    # end location
+    lon_to = xext[1]
+    lat_to = yext[1]
+    
+    # This gets geodesic between the two points
+    # WGS84 ellipsoid is used
+    gl = Geodesic.WGS84.InverseLine(lat_fr, lon_fr, lat_to, lon_to) 
+    num_points = 10
```
hCraker, line 49: Could this value be added to the argument list for the function instead of hard coded?

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
...
+    gvutil.set_axes_limits_and_ticks(
+    ax,
+    xlim=(-125,-60),
+    ylim=(15,65),
+    xticks=np.linspace(-180, 180, 13),
+    yticks=np.linspace(0, 80, 5))
+    
+    # Use geocat.viz.util convenience function to make plots look like NCL plots by using latitude, longitude tick labels
+    gvutil.add_lat_lon_ticklabels(ax)
+
```
clairefio, line 74: I would get rid of the degree symbols in the lat/lon tick marks so the plot more closely resembles the NCL version.

Plots/Trajectory/NCL_polyg_14.py
```diff
@@ -0,0 +1,96 @@
...
+        maintitlefontsize=16,
+        righttitlefontsize=14,
+        xlabel="",
+        ylabel="")
+
+
+
+
+# plot first color map
+Plot("blue", [-125,-60,15,65],[-120, -64], [20, 60], "1st method: Two Points and Great Circle Path", '-', 'blue')
```
hCraker, line 91: The NCL example is using two methods to create the trajectories, but here you are using the same method for both plots. If Python doesn't have two methods to accomplish this, I think the titles should be changed to something more relevant like 'Great Circle Path' and 'Great Circle Path with Markers'.

Plots/Trajectory/NCL_polyg_14.py
clairefio, line 29: I could be wrong, but does having this as a separate "Plot" function make Sphinx unable to generate documentation for it?

hCraker: I think it might be because @michaelavs is using a package that isn't in the geocat-examples environment

michaelavs: I'm not completely sure, I was using the code from the "Comparisons" PR to create this version which Sphinx could generate. I think currently Sphinx isn't able to generate the documentation because there is the additional package installed for running this code that isn't in the environment currently, but I also don't have a very good understanding on how Sphinx works past running "make html".

hCraker: FYI when I ran `make html` locally just now, Sphinx worked fine.

Plots/Trajectory/NCL_polyg_14.py
michaelavs: I'm going to look into this a bit more, I think it would come down to just creating two functions within the code and having them run separately instead of together as they are now. I'll need to checkout which graph is dependent on what and if there are any major separations between them when it comes to using Python vs NCL on this.

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Updated in most recent push

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Added in most recent push

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Updated in most recent push

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Changed in most recent push

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Just kidding, pushed the wrong notebook!

Plots/Trajectory/NCL_polyg_14.py
michaelavs: Pushed the wrong notebook! One moment

Plots/Trajectory/NCL_polyg_14.py
hCraker: The most recent push still has this hard coded. Could you add it as an argument to the plot function? It would be easier for users to then copy your function and use it this way. If you don't think it's necessary to have it as an argument, you could move this assignment to the beginning of the function with a comment about what it represents.

hCraker: Sorry I should have specified. I meant `num_points`

michaelavs:
Updated plots:
image
image

Plots/Trajectories/conda_environment.yml
```diff
@@ -0,0 +1,20 @@
+name: geocat-examples
```
erogluorhan, line 1: This file should be removed from the PR because there is only one `conda_environment.yml` which is located under the root directory. Thus, the geographiclib dependency should be added to that environment file instead. I have done  adding it by another PR #163 in favor of this PR because the dependency should be merged before this PR to get all the Sphinx-builds in the correct way. Since I have already added the dependency, it is good from now on that you either add the same dependency to conda environment as well or not.

Plots/Trajectory/README.rst
```diff
@@ -0,0 +1,6 @@
+.. _trajectory_examples:
```
erogluorhan, line 1: This file as well as "Trajectory" folder should be removed because the correct folder is "Trajectories".

Plots/Trajectory/conda_environment.yml
```diff
@@ -0,0 +1,20 @@
+name: geocat-examples
```
erogluorhan, line 1: This should be removed also because of the same reasons with above two comments.

Plots/Trajectories/conda_environment.yml
michaelavs: I think I should have this resolved and the other two comments about the "Trajectory" directory. Let me know if I missed anything!

erogluorhan: Yes, that's correct.

erogluorhan:
@michaelavs There was a little miscommunication I think. I meant to remove **Plots/Trajectories/conda_environment.yml** and **Plots/Trajectory/conda_environment.yml** but the valid **conda_environment.yml** under the root directory seems to have been removed as well. It shouldn't be removed. Could you please restore it under the root directory?

michaelavs:
> @michaelavs There was a little miscommunication I think. I meant to remove **Plots/Trajectories/conda_environment.yml** and **Plots/Trajectory/conda_environment.yml** but the valid **conda_environment.yml** under the root directory seems to have been removed as well. It shouldn't be removed. Could you please restore it under the root directory?

I think I just got that re-added to this PR (copied conda_environment.yml with changes to "GeoCAT-examples" directory and pushed changes). If not let me know, for some reason I am not able to revert back to any previous versions so I wasn't able to go back to the version where it had been uploaded the first time.

erogluorhan:
> I think I just got that re-added to this PR (copied conda_environment.yml with changes to "GeoCAT-examples" directory and pushed changes). If not let me know, for some reason I am not able to revert back to any previous versions so I wasn't able to go back to the version where it had been uploaded the first time.

Yes the file is back now and the PR builds correctly!

Plots/Trajectories/NCL_polyg_14.py
```diff
@@ -0,0 +1,122 @@
...
+    plt.plot(lons, lats, style, color= color, transform=ccrs.Geodetic())
+    ax.plot(lons, lats, pt, transform=ccrs.PlateCarree())
+    
+    
+    plt.suptitle(title, y=0.90, fontsize=16)
+    
+    # Use geocat.viz.util convenience function to set axes parameters without calling several matplotlib functions
+    # Set axes limits, and tick values
+    gvutil.set_axes_limits_and_ticks(
+    ax,
```
erogluorhan, line 89: Indentation here looks inconsistent with other indentations throughout the code.

Plots/Trajectories/NCL_polyg_14.py
```diff
@@ -0,0 +1,122 @@
...
+    ax = plt.axes(projection=ccrs.PlateCarree())
+
+    ax.set_extent(ext, ccrs.PlateCarree())
+    ax.add_feature(cfeature.LAND, color='lightgrey')
+
+    # This gets geodesic between the two points
+    # WGS84 ellipsoid is used
+    # yext and xext refer to the start and stop points for the curve
+    # [0] being start, [1] being stop
+
```
erogluorhan, line 62: Removing this line break could help increase readability.

Plots/Trajectories/NCL_polyg_14.py
```diff
@@ -0,0 +1,122 @@
...
+    # [0] being start, [1] being stop
+   
+    gl = Geodesic.WGS84.InverseLine(yext[0],xext[0], yext[1], xext[1])
+    npoints = npts
+
+    # Compute points on the geodesic, and plot them 
+    # gl.s13 is the total length of the geodesic
+    # the points are equally spaced by 'true distance', but visually 
+    # there is a slight distortion due to curvature/projection style 
+
```
erogluorhan, line 70: Removing this line break could help increase readability.

Plots/Trajectories/NCL_polyg_14.py
```diff
@@ -0,0 +1,122 @@
...
+        g = gl.Position(ea, Geodesic.STANDARD | Geodesic.LONG_UNROLL)
+        lon2 = g['lon2']
+        lat2 = g['lat2']
+        lons.append(lon2)
+        lats.append(lat2)
+
+    plt.plot(lons, lats, style, color= color, transform=ccrs.Geodetic())
+    ax.plot(lons, lats, pt, transform=ccrs.PlateCarree())
+    
+
```
erogluorhan, line 83: Remove redundant line break if not intentional

Plots/Trajectories/NCL_polyg_14.py
```diff
@@ -0,0 +1,122 @@
...
+import cartopy.feature as cfeature
+from geographiclib.geodesic import Geodesic
+from cartopy.mpl.gridliner import LongitudeFormatter, LatitudeFormatter
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+#Plot
+
+def Plot(color, ext, xext, yext, npts, title, subt, style, pt):
```
erogluorhan, line 31: May be a brief comment about this plotting helper function such as "Helps generate two similar plots where differences are ..." or something you would find useful.

PR #141, opened 2020-07-06
Ncl_lcnative_1_lg.py

michaelavs:
Closes [#63 ](https://github.com/NCAR/GeoCAT-examples/issues/63)
*Updated branch for [previous](https://github.com/NCAR/GeoCAT-examples/pull/81) version of this PR

image
image
image


Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,73 @@
...
+    plt.figure(figsize=(14, 14))
+    # Generate axes using Cartopy and draw coastlines
+    projection = proj#LambertAzimuthalEqualArea()#central_longitude=45)#, standard_parallels=(36,55), globe=ccrs.Globe())
+    ax = plt.subplot(row, col, pos, projection=projection)
+    ax.set_extent((28, 57, 20, 47), crs=ccrs.PlateCarree())
+    ax.coastlines(linewidth=0.5)
+    
+    gl = ax.gridlines(draw_labels=True, dms=False, x_inline=False, y_inline=False)
+    gl.xlabels_top = False
+    gl.ylabels_right = False
```
hCraker, line 56: The NCL example has latitude and longitude labels on the top and right of the plot. I know that the labels don't look quite right when you have tried to add them in before so I think it's okay that they aren't included. Maybe we should add a comment here mentioning Python's limitation with these labels for certain projections.

michaelavs: I think it may be beneficial to actually turn them on but use your idea of having a comment here about the discrepancies  of Python/NCL with the Lambert Conformal projection. This would also make sense just overall so that it doesn't seem random that we have two extra projections that aren't Lambert Conformal for this example as well.

michaelavs:
New visualization:
image

hCraker:
@michaelavs while working on another example, I found that you can specify the tick labels to be horizontal using the following:
```
gl.xlabel_style = {'rotation': 0}
gl.ylabel_style = {'rotation': 0}
```
I think this will make your plots resemble the NCL style more closely.

michaelavs:
I have been experimenting with adding a fourth subplot to showcase the convenience functions which can be used on the cylindrical projection, and found we don't currently have a way of putting tick labels all the way around a projection as seen in the original NCL plot. After (very) briefly looking at the NCL map examples, it appears not many NCL maps have tick labels around the entire projection so I am not sure if it would super necessary to add to the existing function or if just mentioning that the convenience function is only set up to do two axes would be the best route. 

Additionally, I may have come across a bug in the "add_lat_lon_ticklabels" convenience function. The code I have running currently has all extents set correctly, but this function is generating what appears to be random coordinate points some of which are not in the defined extent. @erogluorhan Has this ever been a problem in the past with this function? Here is the code blurb I am use to define the axes:
 `# Use geocat.viz.util convenience function to set axes parameters without calling several matplotlib functions
    # Set axes limits, and tick values
    gvutil.set_axes_limits_and_ticks(
        ax,
        xlim=(30, 55),
        ylim=(20, 45),
        xticks=np.linspace(30, 55, 6),
        yticks=np.linspace(20, 45, 6))
    
    # Use geocat.viz.util convenience function to make plots look like NCL plots by using latitude, longitude tick labels
    gvutil.add_lat_lon_ticklabels(ax)

    # Use geocat.viz.util convenience function to add minor and major tick lines
    gvutil.add_major_minor_ticks(ax, labelsize=12)
    
    # Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
    gvutil.set_titles_and_labels(
        ax,
        maintitle=title,
        maintitlefontsize=16,
        righttitlefontsize=14,
        xlabel="",
        ylabel="")
`
and the output file:
image
Let me know if you see an error that may be causing this

michaelavs:
> I have been experimenting with adding a fourth subplot to showcase the convenience functions which can be used on the cylindrical projection, and found we don't currently have a way of putting tick labels all the way around a projection as seen in the original NCL plot. After (very) briefly looking at the NCL map examples, it appears not many NCL maps have tick labels around the entire projection so I am not sure if it would super necessary to add to the existing function or if just mentioning that the convenience function is only set up to do two axes would be the best route.
> 
> Additionally, I may have come across a bug in the "add_lat_lon_ticklabels" convenience function. The code I have running currently has all extents set correctly, but this function is generating what appears to be random coordinate points some of which are not in the defined extent. @erogluorhan Has this ever been a problem in the past with this function? Here is the code blurb I am use to define the axes:
> `
> 
> # Use geocat.viz.util convenience function to set axes parameters without calling several matplotlib functions
> ```
> # Set axes limits, and tick values
> gvutil.set_axes_limits_and_ticks(
>     ax,
>     xlim=(30, 55),
>     ylim=(20, 45),
>     xticks=np.linspace(30, 55, 6),
>     yticks=np.linspace(20, 45, 6))
> 
> 
> # Use geocat.viz.util convenience function to make plots look like NCL plots by using latitude, longitude tick labels
> gvutil.add_lat_lon_ticklabels(ax)
> 
> # Use geocat.viz.util convenience function to add minor and major tick lines
> gvutil.add_major_minor_ticks(ax, labelsize=12)
> 
> # Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
> gvutil.set_titles_and_labels(
>     ax,
>     maintitle=title,
>     maintitlefontsize=16,
>     righttitlefontsize=14,
>     xlabel="",
>     ylabel="")
> ```
> 
> `
> and the output file:
> image
> Let me know if you see an error that may be causing this

I am going to omit this change for the time being. It may not be a good use of time to dive into the potential bug on this as the example gallery is partly to showcase the capabilities of Python.

michaelavs:
Updated plots:
image

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,107 @@
...
+    you will be able to create the "rectangular" style of coordinates not classically 
+    represented by a Lambert Conformal map. Additionally, Cartopy does not currently support 
+    adding tick marks to a projection like NCL, this is why these Python projections do not 
+    have this feature. The GeoCAT Team is actively adding to the list of convenience functions
+    supported and hopes to add this functionality one day.
+    '''
+    
+    # Plot data and create colorbar
+    newcmp = gvcmaps.BlueYellowRed
+    t.plot.contourf(ax=ax, cmap=newcmp, transform=ccrs.PlateCarree(), levels = 14, cbar_kwargs={"orientation":"horizontal",
```
clyne, line 95: Can you shorten your line length so that generated notebook doesn't require horizontal scrolling?

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,107 @@
...
+    # Plot data and create colorbar
+    newcmp = gvcmaps.BlueYellowRed
+    t.plot.contourf(ax=ax, cmap=newcmp, transform=ccrs.PlateCarree(), levels = 14, cbar_kwargs={"orientation":"horizontal", 
+                                                                "ticks":np.arange(0, 240, 20),  "label":'', "shrink":0.9})
+
+    plt.title(title, loc='center', y=1.17, size=15)
+    plt.title(t.units, loc='right', y=1.08,  size=14)
+    plt.title("precipitation", loc='left', y=1.08, size=14)
+
+Plot(2,2,1, ccrs.LambertConformal(central_longitude=45, standard_parallels=(36,55), globe=ccrs.Globe()), "Lambert Conformal")
```
clyne, line 102: Ditto on line length

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+            number of columns necessary for subplotting 
+        pos (:class: 'int'): 
+            position of visualization in m x n subplot
+        proj (:class: 'cartopy.crs'): 
+            which projection to visualize
+        title (:class: 'str'): 
+            center title of respective visualization
+    '''    
+    # Generate axes using Cartopy and draw coastlines
+    projection = proj
```
erogluorhan, line 61: Could use `proj` directly throughout this function without creating a new variable  `projection`.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+    ax = plt.subplot(row, col, pos, projection=projection)
+    ax.set_extent((28, 57, 20, 47), crs=ccrs.PlateCarree())
+    ax.coastlines(linewidth=0.5)
+    # Import an NCL colormap
```
erogluorhan, line 65: Bettor to leave a line break before this

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+    newcmp = gvcmaps.BlueYellowRed
+
+
```
erogluorhan, line 68: Could remove one of these line breaks as well as put a descriptive comment before the next block begins.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+    gl.xlines = False
+    gl.ylines = False
+    gl.xlocator = mticker.FixedLocator([30,35,40,45,50,55])
+    gl.ylocator = mticker.FixedLocator([20,25,30,35,40,45])
+    gl.xlabel_style = {'rotation':0}
+    gl.ylabel_style = {'rotation':0}
+    
+    '''
+    When using certain types of projections in Cartopy, you may find that there
+    is not a direct 1-to-1 projection simularity. When looking at the three Lambert
```
erogluorhan, line 81: Typo "simularity"?

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+    supported and hopes to add this functionality one day.
+    '''
+    
+    # Plot data and create colorbar
+    newcmp = gvcmaps.BlueYellowRed
+    t.plot.contourf(ax=ax, cmap=newcmp, transform=ccrs.PlateCarree(), 
+                    levels = 14, cbar_kwargs={"orientation":"horizontal", 
+                    "ticks":np.arange(0, 240, 20),  "label":'', "shrink":0.9})
+
+    plt.title(title, loc='center', y=1.17, size=15)
```
erogluorhan, line 99: Could these three lines be replaced by `gvutil.set_titles_and_labels`?

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+    in the Cartopy version. By using Lambert Cylindrical over Lambert Conformal in Python,
+    you will be able to create the "rectangular" style of coordinates not classically 
+    represented by a Lambert Conformal map. Additionally, Cartopy does not currently support 
+    adding tick marks to a projection like NCL, this is why these Python projections do not 
+    have this feature. The GeoCAT Team is actively adding to the list of convenience functions
+    supported and hopes to add this functionality one day.
+    '''
+    
+    # Plot data and create colorbar
+    newcmp = gvcmaps.BlueYellowRed
```
erogluorhan, line 94: This was already set above, so there is no need for this assignment again.

Plots/MapProjections/NCL_lcnative_1_lg.py
```diff
@@ -0,0 +1,109 @@
...
+        title (:class: 'str'): 
+            center title of respective visualization
+    '''    
+    # Generate axes using Cartopy and draw coastlines
+    projection = proj
+    ax = plt.subplot(row, col, pos, projection=projection)
+    ax.set_extent((28, 57, 20, 47), crs=ccrs.PlateCarree())
+    ax.coastlines(linewidth=0.5)
+    # Import an NCL colormap
+    newcmp = gvcmaps.BlueYellowRed
```
erogluorhan, line 66: If that's intentional to pick a different color map than NCL, please clarify that in the function docstrings.

michaelavs: I did mention at the top in the "concepts illustrated" portion that using best practices for choosing color scheme is implemented in the code. However, do you think it would be better to have a comment about this in the code as well, or do one or the other (either comment about it in the code and not at the top, or leave it as is)?

erogluorhan: Ah, sorry!  I did not notice that. May be you add some information to that concept illustrated about what NCL's original was and what we used in this example.

michaelavs: I added a small blurb at the top and linked to the "colors" version of this in the GeoCAT-Examples gallery. I also updated the color scheme to be "Blues_r" because for the "colors" version, we stated using a solid color would be the best route for this type of data

Plots/MapProjections/NCL_lcnative_1_lg.py
michaelavs: I tried updating to using the gvutil version, but because we can't set the location of the titles/labels in either gvutil or with ax.gridlines, this will have to be hard coded unless one is updated to allow location settings

michaelavs:
>  The 'extend' parameter to Colorbar has no effect because it is overridden by the mappable; it is deprecated since 3.3 and will be removed two minor releases later.

I've been looking into this depreciation warning, but I'm not sure what is triggering it because I'm not using an "extend" parameter (unless the "shrink" kwarg is part of "extend"). I was looking at the [Matplolib depreciation list](https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html), and it doesn't really provide much on what "extend" variable would be depreciated, but it does mention the cmap argument. However, I think the way I use cmap in the code is part of contourf not cbar. Do you have any ideas what may be triggering this warning?

Plots/MapProjections/NCL_lcnative_1_lg.py
erogluorhan: It sounds good now except that you need to remove 

```
from geocat.viz import util as gvutil
from geocat.viz import cmaps as gvcmaps
```

because they are not used.

erogluorhan:
> I've been looking into this depreciation warning, but I'm not sure what is triggering it because I'm not using an "extend" parameter (unless the "shrink" kwarg is part of "extend"). I was looking at the [Matplolib depreciation list](https://matplotlib.org/3.3.0/api/prev_api_changes/api_changes_3.3.0/deprecations.html), and it doesn't really provide much on what "extend" variable would be depreciated, but it does mention the cmap argument. However, I think the way I use cmap in the code is part of contourf not cbar. Do you have any ideas what may be triggering this warning?

I first thought if it was a trivial thing to change some function call and overcome, we'd do it, but I agree with your findings; it's probably from within the API where xArray 's `contourf` function calls `matplotlib`. So, I don't think you need to do anything  on this for the time.

PR #142, opened 2020-07-06
xy_10

### LINKED ISSUE #139: xy_10
Description: Filling the area between two curves and using Greek characters

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/xy_10.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/xy_10_lg.png

Category: XY

hCraker:
Closes #139 
image


hCraker:
> Good start! The only thing I have noticed so far is that the original has North/South labels on the ticks, so I would add those to this version too if possible.

Good catch! I've made the change.

PR #145, opened 2020-07-07
Sat_3

clairefio:
sat_3


clairefio:
@hCraker @michaelavs  From what I've researched, there's not really any good mechanism for choosing tick placement on zoomed-in orthographic projections. I set the ticks myself with a helper function. If either of you know any good alternative, please let me know.

Plots/Satellites/NCL_sat_3.py
```diff
@@ -0,0 +1,155 @@
...
+                    )
+
+# Get rid of black outline on colorbar
+cbar.outline.set_visible(False)
+
+# Set plot titles
+main = r"$\bf{Example}$" + " " + r"$\bf{of}$" + " " + r"$\bf{Zooming}$" + \
+       " " + r"$\bf{a}$" + " " + r"$\bf{Sat}$" + " " + r"$\bf{Projection}$"
+left = 'Potential Temperature'
+right = 'Celsius'
```
hCraker, line 136: Instead of hardcoding these values, you may be able to use `t.long_name` and `t.units`. NetCDF files have variable names and units built in as part of the metadata and this is how xarray lets you access it.

Plots/Satellites/NCL_sat_3.py
```diff
@@ -0,0 +1,155 @@
...
+
+import geocat.datafiles as gdf
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Define a helper function for plotting lat/lon ticks on an orthographic plane
+
+
+def plotOrthoTicks(coords, loc):
```
hCraker, line 32: In #150 on lines 98-105 of `NCL_shapefiles_1.py` I had a more succinct way of drawing latitude and longitude ticks on a Lambert Conformal projection . I don't know if it will work for Orthographic, but you might be able to modify it for a solution that is shorter than this. If not, what you have is fine.

clairefio: I tried this, but there is no way to specify which ticks you want labeled, or where you want the labels to be. It just plots labels automatically based on which gridlines you specified. Do you think it would be better if I just used this built-in function even if the resulting plot doesn't match the NCL version? This is what it looks like:
sat_3 alt ticks

michaelavs: I agree with Rose that what you had before should be fine. It looks like cartopy did try to [add support](https://github.com/SciTools/cartopy/pull/1117) for adding lat/lon lines for all projections, but its pretty basic in the way it will be projected and I am assuming thats you used for this one. I think overall, it may be best to show off the capabilities of python over just using built-in functions when necessary. My impression is the examples we produce won't just be used for simply showing projections but also as a tool to show how to use Python for anyone coming over from NCL exclusively.

clairefio:
> Looks good, Claire. A couple of comments:
> 
> 1. The NCL version has automatic placement of the tick marks. We should document that Cartopy lacks this capability and the tick marks have to unfortunately be specified. It would be good to also go ahead and open an in issue in geocat-viz requesting this feature and referencing this PR.
> 2. From what I can tell the the main point of this (and the other sat* plots) is to demonstrate "satellite" map projections, which are really just orthographic projections. I think rather than creating a separate Satellite folder for these we should probably put them in the existing Map Projections directory.

Ok, I moved it to the Contours directory.

clyne:
> > Looks good, Claire. A couple of comments:
> > 
> > 1. The NCL version has automatic placement of the tick marks. We should document that Cartopy lacks this capability and the tick marks have to unfortunately be specified. It would be good to also go ahead and open an in issue in geocat-viz requesting this feature and referencing this PR.
> > 2. From what I can tell the the main point of this (and the other sat* plots) is to demonstrate "satellite" map projections, which are really just orthographic projections. I think rather than creating a separate Satellite folder for these we should probably put them in the existing Map Projections directory.
> 
> Ok, I moved all three satellite plots to the Contours directory.

Hmm. How about the "Map Projections" directory, not the Contours directory, since the main point of this example seems to be the use of the orthographic projection; the contouring component is really no different than any other contour plot.

clairefio:
> > > Looks good, Claire. A couple of comments:
> > > 
> > > 1. The NCL version has automatic placement of the tick marks. We should document that Cartopy lacks this capability and the tick marks have to unfortunately be specified. It would be good to also go ahead and open an in issue in geocat-viz requesting this feature and referencing this PR.
> > > 2. From what I can tell the the main point of this (and the other sat* plots) is to demonstrate "satellite" map projections, which are really just orthographic projections. I think rather than creating a separate Satellite folder for these we should probably put them in the existing Map Projections directory.
> > 
> > 
> > Ok, I moved all three satellite plots to the Contours directory.
> 
> Hmm. How about the "Map Projections" directory, not the Contours directory, since the main point of this example seems to be the use of the orthographic projection; the contouring component is really no different than any other contour plot.

Sorry, didn't know there was an actual MapProjections directory that had been created. Changed it!

clairefio:
> Looks great! However, I could not make sure if you would still work on this since it is still in Draft status. Are you going to continue improving this or should we merge it?

Thanks! And yes, it is ready to merge.

PR #147, opened 2020-07-08
Comparisons

michaelavs:
Resolves issue #92 
*Updated to allow local branch to push to PR


michaelavs:
@hCraker I just ran make html and it ran successfully but did not output any of the files in "Colors" nor make a Colors directory. @erogluorhan, do you know why this may be happening? In the past I haven't had any difficulties with this directory and it appears it isn't just a local issue as it was happening when Rose attempted as well

erogluorhan:
> @hCraker I just ran make html and it ran successfully but did not output any of the files in "Colors" nor make a Colors directory. @erogluorhan, do you know why this may be happening? In the past I haven't had any difficulties with this directory and it appears it isn't just a local issue as it was happening when Rose attempted as well

Yes, since "Colors" directory is a new folder with this PR, you should put a `README.rst `in it as in the other directories. Don't forget to modify its content properly for Colors folder.

Plots/Colors/CB_Height.py
```diff
@@ -0,0 +1,129 @@
+"""
+CB_Height.py
+================
+
+This script illustrates multiple color schemes for color maps which will allow for those
+impacted by color blindness to see visualizations. Using rainbow color schemes is also
+a poor choice in color scheme for images that may be transferred to a black and white
```
clyne, line 7: Can we make  a stronger case for why to avoid the rainbow? See the three reasons given in the Moreland paper: https://doi.org/10.2352/ISSN.2470-1173.2016.16.HVEI-133

michaelavs: @clyne I include the three reasons in the figure 1 descriptions, but can move them here instead and change that description if you think that would be the best route

clyne: OK. Looks good.

Issue #148, opened 2020-07-08
xy_35


clairefio:
description: Controlling color and opacity in xy plots

plot: https://www.ncl.ucar.edu/Applications/Images/xy_35_lg.png
script: https://www.ncl.ucar.edu/Applications/Scripts/xy_35.ncl

hCraker:
This issue was not linked to it's associated PR so it didn't close when PR #149 was merged.

PR #149, opened 2020-07-08
Xy_35

clairefio:
xy_35

### DIFF
--- a/Plots/XY/NCL_xy_35.py
+++ b/Plots/XY/NCL_xy_35.py
@@ -1,0 +1,176 @@
+"""
+NCL_xy_35.py
+============
+This script illustrates the following concepts:
+   - Use of new (with NCL 6.4.0) opacity resources
+     to control opacity of curves and markers in 
+     XYPlots.
+   - The per-curve/per-marker resource arrays override
+     the single-value opacity resources.      
+   - Unlike color resources, the opacity resources
+     are applicable with the Mono-color resources
+     are in effect.
+
+See following URLs to see the reproduced NCL plot & script:
+    - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/xy_35.ncl
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/xy_35_lg.png
+"""
+
+###############################################################################
+# Import packages:
+
+import numpy as np
+import xarray as xr
+from matplotlib import pyplot as plt
+from matplotlib import colors
+
+import geocat.datafiles as gdf
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Create data:
+
+# Make array of x-values, 64 evenly spaced values between 0 and 1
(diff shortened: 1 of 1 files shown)


Plots/XY/NCL_xy_35.py
```diff
@@ -0,0 +1,249 @@
...
+
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Create data:
+
+# Make array of x-values, 64 evenly spaced values between 0 and 1
+f = np.linspace(0., 1., 64)
+
+twopi = 2 * 3.141592654
```
hCraker, line 32: This is probably sufficient for pi, but if you want to you could use the math package instead and use `math.pi`

Plots/XY/NCL_xy_35.py
```diff
@@ -0,0 +1,249 @@
...
+
+# Create figure with 6 axes
+fig, axes = plt.subplots(figsize=(7, 10), nrows=3, ncols=2)
+plt.subplots_adjust(wspace=0.3, hspace=.4)
+
+# Plot first graph
+ax = axes[0, 0]
+
+# Use geocat-viz function to add major and minor ticks on the x and y axes
+gvutil.add_major_minor_ticks(ax,
```
hCraker, line 50: You have a lot of code here that is very similar. You may be able to use a loop to define all of the axes, ticks, tick labels, etc then after the loop you can plot the lines with the different capacities and markers as well as make the different titles. I think it might be neater and reader friendly.

Plots/XY/NCL_xy_35.py
michaelavs: Or to avoid importing another package, I would use np.pi. Both have the same value, but like Rose said up to you!

Plots/XY/NCL_xy_35.py
```diff
@@ -0,0 +1,168 @@
...
+line2 = axesList[4].plot(f, y, 'o', ls='-', color=lcolor2,
+                         ms=3, markevery=.1, mec='None', mfc=mcolor2)
+line3 = axesList[4].plot(f, z, 'o', ls='-', color=lcolor3,
+                         ms=3, markevery=.1, mec='None', mfc=mcolor3)
+
+
+# Plot sixth graph:
+
+# Create RGBA tuples for lines
+lcolor1 = colors.to_rgba('black', alpha=alpha)
```
hCraker, line 150: Why define `lcolor1` `lcolor2` and `lcolor3` if they are all the same value rather than using one variable say `lcolor` for all three lines? Your current method is a bit more readable, but condensing it wouldn't make this that much harder to understand. You can change it or keep what you have. Either is fine. This is just food for thought.

clairefio: I just did that for consistency, I think I will keep it in just because the 5th graph uses that same format of defining line colors as well.

hCraker:
Closes #148

PR #150, opened 2020-07-08
shapefiles_1

### LINKED ISSUE #143: shapefiles_1
Description: Using a shape file to color different states

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/shapefiles_1.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/shapefiles_1_lg.png

Category: Shape Files

hCraker:
Closes #143 
image

### DIFF
--- a/Plots/Shapefiles/NCL_shapefiles_1.py
+++ b/Plots/Shapefiles/NCL_shapefiles_1.py
@@ -1,0 +1,38 @@
+"""
+NCL_shapefiles_1.py
+===============
+This script illustrates the following concepts:
+   - Reading shapefiles
+   - Plotting data from shapefiles
+   - Using shapefile data to plot unemployment percentages in the U.S.
+   - Drawing a custom labelbar on a map
+   - Drawing filled polygons over a Lambert Conformal plot
+   - Drawing the US with a Lambert Conformal projection
+   - Zooming in on a particular area on a Lambert Conformal map
+   - Centering the labels under the labelbar boxes
+
+See following URLs to see the reproduced NCL plot & script:
+    - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/shapefiles_1.ncl
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/shapefiles_1_lg.png
+"""
+
+###############################################################################
+# Import packages:
+import matplotlib.pyplot as plt
+import shapefile as shp
+import numpy as np
+
+import geocat.datafiles as gdf
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Read in data:
+
+# Open all shapefiles and associated .dbf, .shp, and .prj files so sphinx can run and generate documents
+file1 = open(gdf.get("shape_files/states.dbf"), 'r')
+file2 = open(gdf.get("shape_files/states.shp"), 'r')
+file3 = open(gdf.get("shape_files/states.shx"), 'r')
+file4 = open(gdf.get("shape_files/states.prj"), 'r')
+
+# Open shapefiles
+shape = shp.Reader(gdf.get("shape_files/states.dbf"))
--- a/Plots/Shapefiles/NCL_shapefiles_1.py
+++ b/Plots/Shapefiles/NCL_shapefiles_1.py
@@ -104,4 +104,7 @@ gl.ylabel_style = {'rotation': 0}
 gl.xlines = False
 gl.ylines = False
 
+# Use geocat.viz.util convenience function to set titles and labels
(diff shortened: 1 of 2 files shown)


Plots/Shapefiles/NCL_shapefiles_1.py
```diff
@@ -0,0 +1,110 @@
...
+import matplotlib.pyplot as plt
+import matplotlib.patches as mpatches
+import matplotlib.colors as colors
+import matplotlib.cm as cm
+import matplotlib.ticker as mticker
+import shapefile as shp
+import numpy as np
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import cartopy.mpl.gridliner as cgrid
```
michaelavs, line 30: This line isn't used in the code
```suggestion

```

hCraker: Good catch thanks!

Plots/Shapefiles/NCL_shapefiles_1.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        patch = mpatches.Polygon(shape.points, facecolor=color, edgecolor='black', linewidth=0.5, transform=ccrs.PlateCarree(), zorder=2)
+        ax.add_patch(patch)
+
+# Create colorbar
+plt.colorbar(cm.ScalarMappable(cmap=colormap, norm=norm), ax=ax, boundaries=colorbounds, orientation='horizontal', shrink=0.75, ticks=[1, 2, 3, 4], label='percent', aspect=30, pad=0.075)
+
+# Add latitude and longitude labels
+gl = ax.gridlines(draw_labels=True, x_inline=False, y_inline=False)
+gl.xlocator = mticker.FixedLocator(np.linspace(-120, -80, 5))
```
clairefio, line 99: Is there any way you can set the top/bottom x-axes, and the left/right y-axes to have different tick mark locations? For example, having the bottom x-axis range from -120 to -80 but the top x-axis range from -135 to -75, like in the NCL plot.

hCraker: I'm not sure, but I'll see if I can figure it out.

hCraker: @clairefio Different labels for the top/bottom and left/right sides of maps doesn't seem to be supported in cartopy. I could turn off the labels being drawn on the top and right side and then add the labels in manually but I'm not sure if it's worth it since the plot is legible as is. What do you think?

clairefio: I've run into that problem as well. I think how you have it now should be fine!

jukent:
Could you add tick marks along the axis?

hCraker:
> Could you add tick marks along the axis?

I haven't found an easy way to do that yet, but I'll try again.

hCraker:
> Everything look great. However, can you adjust your line lengths so there is not so much scrolling in the Jupyter Notebook? I believe @erogluorhan and @pilotchute have settled on a line length limit of 99 characters, but it might make sense to make this even shorter for the examples gallery to avoid the horizontal scroll bar in the notebooks.

Yeah I can do that! I totally forgot to fix the line lengths before asking for review

hCraker:
@jukent Adding tick marks doesn't seem to have a intuitive solution. I've tried changing the major locator of the axes to a few different locators, but nothing works. Alternatively, I could add in grid lines to give the viewer a better idea of what the labels are marking which looks like this. Do you think this is a good solution for now, or should I try to find something else?
image

Plots/Shapefiles/NCL_shapefiles_1.py
```diff
@@ -0,0 +1,144 @@
+"""
+NCL_shapefiles_1.py
+===============
+This script illustrates the following concepts:
+   - Reading shapefiles
+   - Plotting data from shapefiles
+   - Using shapefile data to plot unemployment percentages in the U.S.
+   - Drawing a custom labelbar on a map
```
erogluorhan, line 8: Use "colorbar" instead of NCL's "labelbar" term

Plots/Shapefiles/NCL_shapefiles_1.py
```diff
@@ -0,0 +1,144 @@
...
+   - Drawing filled polygons over a Lambert Conformal plot
+   - Drawing the US with a Lambert Conformal projection
+   - Zooming in on a particular area on a Lambert Conformal map
+   - Centering the labels under the labelbar boxes
```
erogluorhan, line 12: Same as above

PR #155, opened 2020-07-10
panel_6

### LINKED ISSUE #154: panel_6
Description: Paneling four plots and demonstrating adding white space between panels

Script URL: https://www.ncl.ucar.edu/Applications/Scripts/panel_6.ncl
Plot URL: https://www.ncl.ucar.edu/Applications/Images/panel_6_1_lg.png
https://www.ncl.ucar.edu/Applications/Images/panel_6_2_lg.png

Category: Panel

hCraker:
Closes #154 
image

### DIFF
--- a/Plots/Panels/NCL_panel_6.py
+++ b/Plots/Panels/NCL_panel_6.py
@@ -15,6 +15,7 @@ See following URLs to see the reproduced NCL plot & script:
 # Import packages:
 import cartopy.crs as ccrs
 import cartopy.feature as cfeature
+import matplotlib.colors as mcolors
 import matplotlib.pyplot as plt
 import matplotlib.ticker as mticker
 import numpy as np
@@ -151,13 +152,13 @@ fig, axs = plt.subplots(2, 2, figsize=(8, 8), gridspec_kw=dict(wspace=0.5), subp
 for row in range(0,2):
     for col in range(0,2):
         # Add map features
-        axs[row][col].add_feature(cfeature.LAND, facecolor='silver')
-        axs[row][col].add_feature(cfeature.COASTLINE, linewidth=0.5)
+        axs[row][col].add_feature(cfeature.LAND, facecolor='silver', zorder=2)
+        axs[row][col].add_feature(cfeature.COASTLINE, linewidth=0.5, zorder=3)
         axs[row][col].add_feature(cfeature.LAKES, linewidth=0.5, edgecolor='black',
-                            facecolor='None')
+                            facecolor='None', zorder=4)
 
         # Add gridlines and latitude and longitude labels
-        gl = axs[row][col].gridlines(ccrs.PlateCarree(), y_inline=False, draw_labels=True, color='gray', linestyle="--")
+        gl = axs[row][col].gridlines(ccrs.PlateCarree(), y_inline=False, draw_labels=True, color='gray', linestyle="--", zorder=5)
         gl.xlocator = mticker.FixedLocator(np.linspace(-180, 150, 12))
         gl.xlabel_style = {'rotation': 0}
 
@@ -166,11 +167,16 @@ for row in range(0,2):
 
 # Import color map
 cmap = gvcmaps.gui_default
+# Plot filled contours
+data[0][0].plot.contourf(ax=axs[0][0], cmap=cmap, vmin=0, vmax=30, levels=16, transform=ccrs.PlateCarree(), add_colorbar=False, zorder=0)
+data[0][1].plot.contourf(ax=axs[0][1], cmap=cmap, vmin=-2, vmax=26, levels=15, transform=ccrs.PlateCarree(), add_colorbar=False, zorder=0)
+data[1][0].plot.contourf(ax=axs[1][0], cmap=cmap, transform=ccrs.PlateCarree(), add_colorbar=False, zorder=0)
+data[1][1].plot.contourf(ax=axs[1][1], cmap=cmap, transform=ccrs.PlateCarree(), add_colorbar=False, zorder=0)
 
-# Plot
-data[0][0].plot.contourf(ax=axs[0][0], cmap=cmap, vmin=0, vmax=30, levels=16, transform=ccrs.PlateCarree(), add_colorbar=False)
-data[0][1].plot.contourf(ax=axs[0][1], cmap=cmap, vmin=-2, vmax=26, levels=15, transform=ccrs.PlateCarree(), add_colorbar=False)
-data[1][0].plot.contourf(ax=axs[1][0], cmap=cmap, vmin=-0.0275, vmax=0.005, levels=14, transform=ccrs.PlateCarree(), add_colorbar=False)
-data[1][1].plot.contourf(ax=axs[1][1], cmap=cmap, vmin=-0.016, vmax=0.004, levels=11, transform=ccrs.PlateCarree(), add_colorbar=False)
+# Plot contour lines
+data[0][0].plot.contour(ax=axs[0][0], colors='black', linestyles='solid', linewidths=0.5, vmin=0, vmax=30, levels=16, transform=ccrs.PlateCarree(), zorder=1)
+data[0][1].plot.contour(ax=axs[0][1], colors='black', linestyles='solid', linewidths=0.5, vmin=-2, vmax=26, levels=15, transform=ccrs.PlateCarree(), zorder=1)
+data[1][0].plot.contour(ax=axs[1][0], colors='black', linestyles='solid', linewidths=0.5, transform=ccrs.PlateCarree(), zorder=1)
+data[1][1].plot.contour(ax=axs[1][1], colors='black', linestyles='solid', linewidths=0.5, transform=ccrs.PlateCarree(), zorder=1)
 
 plt.show()
--- a/Plots/Panels/NCL_panel_6.py
+++ b/Plots/Panels/NCL_panel_6.py
@@ -7,8 +7,7 @@ This script illustrates the following concepts:
 
 See following URLs to see the reproduced NCL plot & script:
     - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/panel_6.ncl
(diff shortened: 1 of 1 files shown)


hCraker:
I'm having some issues with the color maps. The data for the bottom plots (representing salinity) has values in the range of 30 to 40 psu while the NCL plot is showing values that are much closer to 0 and even negative. I don't see any obvious data manipulation in the NCL script. Do either of you have an idea of what might be causing this? @clairefio @michaelavs

michaelavs:
> I'm having some issues with the color maps. The data for the bottom plots (representing salinity) has values in the range of 30 to 40 psu while the NCL plot is showing values that are much closer to 0 and even negative. I don't see any obvious data manipulation in the NCL script. Do either of you have an idea of what might be causing this? @clairefio @michaelavs
> 
> Additionally, I'm using the same color map as the NCL version but the colors don't match even on the top two plots which have the same data as the upper plots on the NCL example.

My first thought is there may be a units problem, but I don't think that would be the solution based on the current values being plotted and that you were able to get at least 2/4 plotted. Another option may be how refined the data is with the number of levels used. 
For the actual coloring problem, I wonder if it's tied to the data range in the sense that the yellow color that should be the max from NCL is what Python may recognize only as a middle color and therefore sets the range as blue to red not blue to yellow as expected. If need be, I think it's more than fine to use a different color scheme for this if the issue doesn't resolve itself once the first issue is resolved for this

hCraker:
> My first thought is there may be a units problem, but I don't think that would be the solution based on the current values being plotted and that you were able to get at least 2/4 plotted. Another option may be how refined the data is with the number of levels used.

Can you elaborate on what you mean by "how refined" the data is? I've tried plotting all of the levels specified by `z_t` in the NetCDF file, and none of them look remotely similar to what the NCL script is outputting.

michaelavs:
> > My first thought is there may be a units problem, but I don't think that would be the solution based on the current values being plotted and that you were able to get at least 2/4 plotted. Another option may be how refined the data is with the number of levels used.
> 
> Can you elaborate on what you mean by "how refined" the data is? I've tried plotting all of the levels specified by `z_t` in the NetCDF file, and none of them look remotely similar to what the NCL script is outputting.

I'm not if it will be the same with this projection style, but I know when I used WRF Python for a class, if you didn't set enough levels for the contours, it would make the projection either one singular color or would not reflect the correct amount of contours on the map. So it's really more of just how refined the projection is: more levels being more "refined" and able to capture the changes better between the contours

hCraker:
@michaelavs I know you have worked with Polar Stereographic projections before. Have you figured out how to draw the longitude labels around the map edge? This is what I currently have after modifying the code you used for #68 

image

michaelavs:
> @michaelavs I know you have worked with Polar Stereographic projections before. Have you figured out how to draw the longitude labels around the map edge? This is what I currently have after modifying the code you used for #68
> 
> image

We had to use an incredibly hard coded work around which was:
`# Manipulate longitude labels (0, 30 E, 60 E, ..., 30 W, etc.)
ticks = np.arange(0, 210, 30)
etick = ['0'] + [r'%d$^\circ$E' %
                 tick for tick in ticks if (tick != 0) & (tick != 180)] + ['180']
wtick = [
    r'%d$^\circ$W' %
    tick for tick in ticks if (
        tick != 0) & (
            tick != 180)]
labels = etick + wtick
xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -114, -140, -151]
yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]


for xtick, ytick, label in zip(xticks, yticks, labels):
    ax.text(xtick, ytick, label, transform=ccrs.Geodetic())
`
The way to move the labels around (which I just had to experiment with, there wasn't necessarily a rhyme or reason to it) was by modifying the 'yticks' parameters. Something similar may work with the 'xticks' parameters if that is necessary but I don't think I had to adjust those at all

hCraker:
Thanks @michaelavs. That was the code I was trying to modify. If hard coded values are necessary, I will give them a try.

hCraker:
@michaelavs I have found an easy way to draw the longitude labels without hard coding anything. I tried it out on your polar projection plot #68 and it worked there too, so I think this is a robust solution. If you don't mind, I'd like to open up a issue and PR to change the polar_1 code to use this method.

hCraker:
image

michaelavs:
> @michaelavs I have found an easy way to draw the longitude labels without hard coding anything. I tried it out on your polar projection plot #68 and it worked there too, so I think this is a robust solution. If you don't mind, I'd like to open up a issue and PR to change the polar_1 code to use this method.

Definitely! I think we had originally planned on returning to that projection to update it if/when a solution came about for the hard coded side, so re-opening an issue on it is more than ok to me!

hCraker:
Plot without any added white space:
image

Plot with added white space:
image

Plots/Panels/NCL_panel_6.py
```diff
@@ -0,0 +1,410 @@
...
+import cartopy.feature as cfeature
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mticker
+import matplotlib.contour as mcontour
+from mpl_toolkits.axes_grid1.inset_locator import inset_axes
+import numpy as np
+import xarray as xr
+
+import geocat.datafiles as gdf
+from geocat.viz import cmaps as gvcmaps
```
michaelavs, line 32:
```suggestion

```
This line is unused in the code

Plots/Panels/NCL_panel_6.py
```diff
@@ -0,0 +1,410 @@
...
+            this if the western edge of the plot is cut off. Defaults to 0.
+        res (:class:`int`):
+            The size of the incrementation for vertices in degrees. Default is
+            a vertex every one degree of longitude. A higher number results in
+            a lower resolution boundary.
+    """
+    import cartopy.crs as ccrs
+    import matplotlib.path as mpath
+    import matplotlib.patches as mpatches
+    import numpy as np
```
michaelavs, line 92:
```suggestion

```

Plots/Panels/NCL_panel_6.py
```diff
@@ -0,0 +1,410 @@
...
+            A constant to be subtracted from the first entry in lon_range. Use
+            this if the western edge of the plot is cut off. Defaults to 0.
+        res (:class:`int`):
+            The size of the incrementation for vertices in degrees. Default is
+            a vertex every one degree of longitude. A higher number results in
+            a lower resolution boundary.
+    """
+    import cartopy.crs as ccrs
+    import matplotlib.path as mpath
+    import matplotlib.patches as mpatches
```
michaelavs, line 91:
```suggestion

```

hCraker:
> Code looks nice! Plots are looking good except one point: In the first plot, the colorbars of the left column overlaps with the plot of the right column as in the red ellipse below:
> 
> image

That is the intent with this example. The first image shows what the plot looks like without added white space just like in the NCL example. The difference is that `matplotlib` has a default arrangement that leads to overlap while NCL's default looks nicer.

erogluorhan:
> That is the intent with this example. The first image shows what the plot looks like without added white space just like in the NCL example. The difference is that `matplotlib` has a default arrangement that leads to overlap while NCL's default looks nicer. Should I change this plot?

Actually, I thought about that would be intentional, but yes, I am confused since NCL plot has no overlap. I don ot know if we should have this as is with the overlap, or get rid of the overlap but not put a large whitespace as in the 2nd plots? Any thoughts @clyne @jukent ?

hCraker:
I mis-clicked and didn't intend to close this.

jukent:
> > That is the intent with this example. The first image shows what the plot looks like without added white space just like in the NCL example. The difference is that `matplotlib` has a default arrangement that leads to overlap while NCL's default looks nicer. Should I change this plot?
> 
> Actually, I thought about that would be intentional, but yes, I am confused since NCL plot has no overlap. I don ot know if we should have this as is with the overlap, or get rid of the overlap but not put a large whitespace as in the 2nd plots? Any thoughts @clyne @jukent ?

This is an interesting issue because as @hCraker points out the two examples are one without spacing and one fixing it. I think we have 3 options. 1 - leave it as is and use the overlap as motivation in the documentation for why we'd want to space out the plots. 2 - use minimum spacing to avoid overlap in the first plot (leaving both examples kind of redundant to each other. 3 - only showing the second plot that does not have overlap.

hCraker:
> This is an interesting issue because as @hCraker points out the two examples are one without spacing and one fixing it. I think we have 3 options. 1 - leave it as is and use the overlap as motivation in the documentation for why we'd want to space out the plots. 2 - use minimum spacing to avoid overlap in the first plot (leaving both examples kind of redundant to each other). 3 - only show the second plot that does not have overlap.

Option 1 is what I was trying to do. If we only leave the second plot with the fixed spacing, that circumvents the opportunity to teach users how to control panel spacing. I think showing users this feature is the intent with the example given that the NCL script comments state that "Adding white space around paneled plots" is an illustrated concept. Thoughts? @erogluorhan @jukent @clyne

erogluorhan:
> Option 1 is what I was trying to do. If we only leave the second plot with the fixed spacing, that circumvents the opportunity to teach users how to control panel spacing. I think showing users this feature is the intent with the example given that the NCL script comments state that "Adding white space around paneled plots" is an illustrated concept. Thoughts? @erogluorhan @jukent @clyne

Taking into attention the original intent, I think it is better to kepp whatever @hCraker is already implemented.

clyne:
> > Option 1 is what I was trying to do. If we only leave the second plot with the fixed spacing, that circumvents the opportunity to teach users how to control panel spacing. I think showing users this feature is the intent with the example given that the NCL script comments state that "Adding white space around paneled plots" is an illustrated concept. Thoughts? @erogluorhan @jukent @clyne
> 
> Taking into attention the original intent, I think it is better to keep whatever @hCraker is already implemented.

Works for me.

PR #156, opened 2020-07-10
Changing xy_5 to plot data using fractional dates

### LINKED ISSUE #153: xy_5: Convert dates from YYYYMM to floats
In the NCL version of the [xy_5 example](https://www.ncl.ucar.edu/Applications/Scripts/xy_5.ncl), the datafile represents dates in a `YYYYMM` format. To make the plot, the NCL example first converts the date array into floats using a format where the decimal component represents a fraction of the year in place of months. This was not done for the GeoCAT version of xy_5. This should be changed not only for consistency, but to produce an accurate graph more simply. It is easier to add ticks and tick labels to a graph that is produced with fractional years for the x axis than with months from the starting date as the x axis.

hCraker:
Closes #153 
In the NCL version of the xy_5 example, the datafile represents dates in a YYYYMM format. To make the plot, the NCL example first converts the dates into floats using a format where the decimal component represents a fraction of the year in place of months. 

The GeoCAT version of xy_5 uses another method to accurately plot the date while also converting the dates; however the fractional dates are never used and are also calculated incorrectly. This PR makes the GeoCAT example more accurate and consistent with the NCL one.

### DIFF
--- a/Plots/XY/NCL_xy_5.py
+++ b/Plots/XY/NCL_xy_5.py
@@ -6,7 +6,7 @@ This script illustrates the following concepts:
    - Drawing a Y reference line in an XY plot
    - Filling the areas of an XY curve above and below a reference line
    - Using named colors to indicate a fill color
-   - Creating array of dates to use as x-axis tick labels
+   - Converting dates from YYYYMM format to floats
    - Creating a main title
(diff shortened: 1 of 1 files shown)


hCraker:
Visually the plot is unchanged.
image

PR #158, opened 2020-07-13
Meteo 1

clairefio:
meteo_1

### DIFF
--- a/Plots/Meteograms/NCL_meteo_1.py
+++ b/Plots/Meteograms/NCL_meteo_1.py
@@ -51,8 +51,8 @@ vgrid = ds.vgrid
 rain03 = ds.rain03
 tempht = ds.tempht
 
-smoothtemp = smooth_n_point(tempisobar, n=5, passes=1)
-smoothrh   = smooth_n_point(rh, n=5, passes=1)
+#smoothtemp = smooth_n_point(tempisobar, n=9, passes=1)
+#smoothrh   = smooth_n_point(rh, n=9, passes=1)
 smoothtemp=tempisobar
 smoothrh=rh
 
@@ -71,11 +71,11 @@ ax1.set_aspect(1.5)
 
 # Import an NCL colormap
 cmap = gvcmaps.wgne15
-colors = ListedColormap(np.array(['white', 'honeydew', 'palegreen', 'limegreen', 'green', 'darkgreen']))
+#colors = ListedColormap(np.array(['white', 'white', 'white', 'honeydew', 'honeydew', 'palegreen', 'palegreen', 'springgreen', 'limegreen', 'green', 'darkgreen']))
+colors = ListedColormap(np.array(['white', 'white', 'white', 'white', 'white', 'mintcream', "#DAF6D3", "#B2FAB9", "#B2FAB9", 'springgreen', 'lime', "#54A63F"]))
+bounds = [-20, -10, 0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100]
 
-bounds = [-20, -10, 0, 10, 20, 30, 40, 50, 60, 70, 80, 90]
-
-norm = BoundaryNorm(boundaries=bounds, ncolors=5) 
+norm = BoundaryNorm(boundaries=bounds, ncolors=12) 
 
 # Plot filled contour
 contour1 = ax1.contourf(smoothrh, transform=ccrs.PlateCarree(), cmap=colors, norm=norm,
@@ -105,7 +105,7 @@ cont3labels = ax1.clabel(contour3, manual=cont3Labels, fmt='%d', inline=True, fo
 
 gvutil.set_titles_and_labels(ax1, maintitle='Meteogram for LGSA, 28/12Z', maintitlefontsize=18, ylabel='Pressure (mb)', labelfontsize=12)
 
-yticklabels = np.array(levels)
+yticklabels = np.array(levels, dtype=np.int)
 xticklabels = ['12z', '15z', '18z', '21z', 'Apr29', '03z', '06z', '09z', '12z', '15z', '18z',
                '21z', 'Apr30', '03z', '06z', '09z', '12z', '15z', '18z', '21z', 'May01', '03z', '06z', '09z', '12z']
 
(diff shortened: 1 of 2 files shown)


clairefio:
Checks are not passing because the metpy module was imported for nine-point smoothing functionality.

clairefio:
Updated plot:

meteo_1

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+                       colors='red',
+                       levels=[-20, -10, 0, 10, 20, 30, 40, 50, 60],
+                       linewidths=0.4,
+                       linestyles='solid',
+                       zorder=4)
+
+# Create lists of coordinates where the contour labels are going to go
+cont2Labels = [(1.71, 3.82), (5.49, 3.23), (9.53, 4.34), (9.27, 3.53),
+               (14.08, 4.81), (19.21, 2.24), (17.74, 1.00), (22.23, 3.87),
+               (12.87, 2.54), (10.45, 6.02), (11.51, 4.92)]
```
hCraker, line 104: Can you add a comment explaining how you got these coordinates? They don't match what the x and y scales show.

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+                                 xlim=[0, 24],
+                                 ylim=[0, 7],
+                                 xticks=np.arange(0, 25),
+                                 yticks=np.arange(0, 8),
+                                 xticklabels=xticklabels,
+                                 yticklabels=yticklabels)
+
+# Make the ticks on each axis point inwards instead of outwards
+ax1.tick_params(axis="x", direction="in")
+ax1.tick_params(axis="y", direction="in", labelsize=9)
```
hCraker, line 156: Your plot doesn't have inward facing ticks. I don't know if these two lines did what you expected them to do.

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+ugrid = ds.ugrid
+vgrid = ds.vgrid
+rain03 = ds.rain03
+tempht = ds.tempht
+
+###############################################################################
+# Plot:
+
+# Generate figure (set its size (width, height) in inches)
+fig = plt.figure(figsize=(6, 10))
```
hCraker, line 55: Instead of using `plt.figure` you could to use `plt.subplots` and specify the ratio between the axes heights using the key words `height_ratios` and `gridspec_kw`. That way the pressure and wind barb map won't be squished vertically and can have an aspect ratio similar to the NCL version. This would also allow you to make three subplots stacked vertically with different heights instead of using inset axes. 

It's up to you if you want to use `plt.figure` or `plt.subplots`, but either way the top set of axes should be a bit taller.

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+gvutil.set_axes_limits_and_ticks(axin1,
+                                 xlim=[0, 72],
+                                 ylim=[0, .5],
+                                 xticks=np.arange(0, 75, 3),
+                                 yticks=np.arange(0, .6, 0.1),
+                                 xticklabels=xticklabels,
+                                 yticklabels=yticklabels)
+
+# Use the geocat.viz function to add minor ticks
+gvutil.add_major_minor_ticks(axin1,
```
hCraker, line 242: The rain total plot doesn't have minor xticks. You can remove them by specifying `x_minor_per_major=1`

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+gvutil.set_axes_limits_and_ticks(axin2,
+                                 xlim=[0, 72],
+                                 ylim=[59, 64.5],
+                                 xticks=np.arange(0, 75, 3),
+                                 yticks=np.arange(59, 65),
+                                 xticklabels=xticklabels,
+                                 yticklabels=yticklabels)
+
+# Use the geocat.viz function to add minor ticks
+gvutil.add_major_minor_ticks(axin2,
```
hCraker, line 280: The temperature plot doesn't have minor xticks. You can remove them by specifying `x_minor_per_major=1`

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,291 @@
...
+
+# Add a pad between the y axis label and the axis spine
+ax1.yaxis.labelpad = 4
+
+# Rotate the labels on the x axis so they are vertical
+for tick in ax1.get_xticklabels():
+    tick.set_rotation(90)
+
+# Make an axis to overlay on top of the contour plot
+axin = fig.add_subplot(311)
```
hCraker, line 166: Are you sure you need to make a subplot over the contour plot? You should be able to simply call `ax1.barbs()` after calling `ax1.contourf` and `ax.contour`. You may need to specify `zorder` to get the layering right, but I believe there is a simpler solution than this.

If you can do this without making another subplot, can you move the creation of the wind barbs by the code for the contour lines? It will make this more readable.

clairefio: No, because the contour plots are on a plate carree projection, but the x and y values for the barbs are not coordinates. The scale is different, which is also why the x and y tick units in the first plot do not make sense with the contour map.

hCraker: Could you use cartopy's transform points?

clairefio:
> This looks good so far! Please see my inline comments. Additionally, you have more circles in your wind barb field than in the NCL plot. Is there a way you can specify at what value a circle should be drawn instead of a wind barb?

Unfortunately there's no way that I have found to change whether or not you plot a circle or a wind barb.

clairefio:
Updated plot:

meteo_1

hCraker:
> > This looks good so far! Please see my inline comments. Additionally, you have more circles in your wind barb field than in the NCL plot. Is there a way you can specify at what value a circle should be drawn instead of a wind barb?
> 
> Unfortunately there's no way that I have found to change whether or not you plot a circle or a wind barb.

At a later date, we may need to make a function with that capability and add it to geocat-viz. For now, I think what you have will do.

Plots/Meteograms/NCL_meteo_1.py
```diff
@@ -0,0 +1,276 @@
...
+ax1.coastlines(linewidths=0.5, alpha=-1)
+
+# Set aspect ratio of the first axis
+ax1.set_aspect(2)
+
+# Create a color map with a combination of matplotlib colors and hex values
+colors = ListedColormap(np.array(['white', 'white', 'white', 'white', 'white',
+                                  'mintcream', "#DAF6D3", "#B2FAB9", "#B2FAB9",
+                                  'springgreen', 'lime', "#54A63F"]))
+bounds = [-20, -10, 0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100]
```
clyne, line 71: "contour_levels", or even "levels" would be more a more meaningful variable name than "bounds". Ditto "normalized_levels" would be better than "norm"?

clairefio: Ok, I made this change.

PR #172, opened 2020-07-22
Mask_5

### LINKED ISSUE #171: Mask_5
link to NCL code: https://www.ncl.ucar.edu/Applications/Scripts/mask_5.ncl
link to NCL projection: https://www.ncl.ucar.edu/Applications/Images/mask_5_lg.png

michaelavs:
closes #171 
NCL code: https://www.ncl.ucar.edu/Applications/Scripts/mask_5.ncl
NCL plot: https://www.ncl.ucar.edu/Applications/Images/mask_5_lg.png
image


michaelavs:
I am having trouble with two things in this projection: 1. The white space between the plots is large and I can'f find a good fix for this and 2. the numbers on the colorbar are not separating out nicely. I've tried to play with the cbar kwargs and the position arguments for the subplots but haven't had any luck with that so far. Have either of you had anything similar happen on your projections or know a work around for these issues? @clairefio @hCraker

hCraker:
> I am having trouble with two things in this projection: 1. The white space between the plots is large and I can'f find a good fix for this and 2. the numbers on the colorbar are not separating out nicely. I've tried to play with the cbar kwargs and the position arguments for the subplots but haven't had any luck with that so far. Have either of you had anything similar happen on your projections or know a work around for these issues? @clairefio @hCraker

The easiest way to manage the spacing is to use `plt.subplots` and use `gridspec_kw` and `hspace`. The paneling examples do this and it works well. 

For the color bar, I would try making your figure wider and reducing the font size of the color bar tick labels. To manipulate the tick labels, you will have to make a color bar using `plt.colorbar` instead of passing arguments into the xarray `contourf` function. Then you can use a line similar to this `cbar.ax.tick_params(labelsize=14)`. I would look at [this example](https://geocat-examples.readthedocs.io/en/latest/gallery/Contours/NCL_lb_3.html#sphx-glr-gallery-contours-ncl-lb-3-py) to learn more about how you can manipulate a color bar.

michaelavs:
> > I am having trouble with two things in this projection: 1. The white space between the plots is large and I can'f find a good fix for this and 2. the numbers on the colorbar are not separating out nicely. I've tried to play with the cbar kwargs and the position arguments for the subplots but haven't had any luck with that so far. Have either of you had anything similar happen on your projections or know a work around for these issues? @clairefio @hCraker
> 
> The easiest way to manage the spacing is to use `plt.subplots` and use `gridspec_kw` and `hspace`. The paneling examples do this and it works well.
> 
> For the color bar, I would try making your figure wider and reducing the font size of the color bar tick labels. To manipulate the tick labels, you will have to make a color bar using `plt.colorbar` instead of passing arguments into the xarray `contourf` function. Then you can use a line similar to this `cbar.ax.tick_params(labelsize=14)`. I would look at [this example](https://geocat-examples.readthedocs.io/en/latest/gallery/Contours/NCL_lb_3.html#sphx-glr-gallery-contours-ncl-lb-3-py) to learn more about how you can manipulate a color bar.

That worked out for me! Thanks!

michaelavs:
Updated plot:
image

michaelavs:
Had to re-do part of the plot due to some formatting issues in Sphinx. The one issue/weird thing on this plot is that the subplots would not scale to be the same size for some reason, but a work around is using multiple subplots (in this case I used 20) and having the index extend for multiple of the subplots. If either of you know a better way to fix the scaling issue, let me know, otherwise the plot looks fine after being run through sphinx when using the 20 subplots method.

Updated plot:
image

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,154 @@
...
+    newcmp = 'magma' 
+    
+    # Contourf-plot data
+    contour = wrap_t.plot.contourf(ax=ax1, transform=ccrs.PlateCarree(),
+                    vmin = 240, vmax = 315, levels = 18, cmap = newcmp, add_colorbar=False)
+                 
+    #Create a colorbar for projection 
+    cbar = plt.colorbar(contour, ax=ax1, orientation='horizontal', shrink=0.70,
+                    pad=0.11, extendrect=True, extendfrac='auto', 
+                    ticks = [240,245,250,255,260,265,270,275,280,285,290,295,300,305,310])
```
hCraker, line 135: Can you use `range` `np.arange` or `np.linspace` here instead of typing all of the ticks out?

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,154 @@
...
+ds = xr.open_dataset(gdf.get("netcdf_files/atmos.nc"), decode_times=False)
+t = ds.TS.isel(time=0)
+
+# Fix the artifact of not-shown-data around 0 and 360-degree longitudes
+wrap_t = gvutil.xr_add_cyclic_longitudes(t, "lon")
+                    
+###############################################################################
+#Plot:
+
+fig = plt.figure(figsize=(12, 14))
```
hCraker, line 43: This is where you should call `plt.subplots`. Notice that this function has an 's' at the end. It does something different than `plt.subplot` which is what you are using right now elsewhere in this file. You might want to use something similar to what I have below. You will likely need to change `figsize` and `hspace`, but this is the easiest way to stack two plots on top of each other. [Here](https://matplotlib.org/3.2.2/api/_as_gen/matplotlib.pyplot.subplots.html) is the documentation page for this function. The `gridspec_kw` gives you a lot of control over the layout.
```fig, ax = plt.subplots(2, 1, figsize=(6, 10), gridspec_kw=dict(hspace=0.3), subplot_kw={"projection": projection})```

File: Plots/Masking/NCL_mask_5.py [hunk 3]
```diff
@@ -0,0 +1,154 @@
...
+
+def Plot(row, col, pos, title):
```
hCraker, line 45: If you use my suggestion above, you could then pass in the `axes` instance instead of `row`, `col`, and `pos`. This will also make your helper functions more intuitive.

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,154 @@
...
+        ax,
+        maintitle=title,
+        maintitlefontsize=16)
+
+# Plot first color map
+Plot(20, 1, (1,7), "land sea mask using 'atmos.nc'")
+
+# Plot second subplot
+
+def Plot2(row, col, pos, title):
```
hCraker, line 103: Same as above

Plots/Masking/NCL_mask_5.py
michaelavs: So I've been playing around with using subplots over subplot and the biggest issue I've been coming across is that it seems when using subplots, you need to reference data that is being displayed on the project, but I'm not sure if this is 100% the case (there aren't many online sources talking about the differences between the two and when to use one over the other).  The issue I am seeing comes with having one plot that uses no data (masking only) and the other is using data (masking with data). I did try using a couple of the panel plots we already have and some of the examples from the documentation page as a reference to what to do, but even within those they are using same source but plotting different data. I may keep playing with it for a bit longer, but if I can't resolve this issue, I will most likely just use the original work around I have for this plot.

hCraker: Using what you have is perfectly fine just make sure you add comments thoroughly explaining why the `row` argument is what it is in lines 99 and 153 since it seems to be a "hacky" solution.

I'm not sure if I fully understand your issue with needing to "to reference data that is being displayed on the project". If you'd like to meet and talk about to try and find a simpler solution, I'm open to that. I've made quite a few panel plots in this project and in others, so maybe I can help.

michaelavs:
> I'm seeing a color for ice shelves on the original image that is not on the python generated one. Is this a difference in the data used or could this be included in the new version?

From what I remember (for this projection), Cartopy doesn't support ice shelfs or the way that it does has been depreciated. I'll see if I can find what resource I had used to come to that conclusion. I believe it was a set of shapefiles that could be loaded in as a package, but something was corrupted/incorrect with ice shelf and islands specifically

michaelavs:
@hCraker @anissa111 I just pushed a newer version of this example that doesn't use subplots, which was what was hanging up this PR initially. The NCL version does have both projections on one figure, but the two projections were not sizing correctly in python which resulted in me using like 20 (?) subplots and forcing sizing. 

I changed to no subplots mainly for the ease of the code, but if we determine this route isn't the best, I can continue to tinker with the plotting functions to figure out a different way that may still use subplots. Let me know your thoughts on this!

hCraker:
> @hCraker @anissa111 I just pushed a newer version of this example that doesn't use subplots, which was what was hanging up this PR initially. The NCL version does have both projections on one figure, but the two projections were not sizing correctly in python which resulted in me using like 20 (?) subplots and forcing sizing.
> 
> I changed to no subplots mainly for the ease of the code, but if we determine this route isn't the best, I can continue to tinker with the plotting functions to figure out a different way that may still use subplots. Let me know your thoughts on this!

I'm hesitant about making two separate figures given that the NCL example comments reference "Paneling two plots on a page". The [NCL_panel_13.py](https://geocat-examples.readthedocs.io/en/latest/gallery/Panels/NCL_panel_13.html) example shows how to do exactly that, so maybe you can rework some of the preexisting code. I know we talked about using `plt.subplots` before and you had issues with it, but this example should help.

michaelavs:
> > @hCraker @anissa111 I just pushed a newer version of this example that doesn't use subplots, which was what was hanging up this PR initially. The NCL version does have both projections on one figure, but the two projections were not sizing correctly in python which resulted in me using like 20 (?) subplots and forcing sizing.
> > I changed to no subplots mainly for the ease of the code, but if we determine this route isn't the best, I can continue to tinker with the plotting functions to figure out a different way that may still use subplots. Let me know your thoughts on this!
> 
> I'm hesitant about making two separate figures given that the NCL example comments reference "Paneling two plots on a page". The [NCL_panel_13.py](https://geocat-examples.readthedocs.io/en/latest/gallery/Panels/NCL_panel_13.html) example shows how to do exactly that, so maybe you can rework some of the preexisting code. I know we talked about using `plt.subplots` before and you had issues with it, but this example should help. I wrote `NCL_panel_13.py`, so if you have questions let me know.

I'm not sure why it was so hard for me to get the subplots to work previously, but I FINALLY got the two plots to work with each other and build in sphinx nicely. So this should now be ready for review!

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,142 @@
...
+=============
+This script illustrates the following concepts:
+    - Using cartopy.feature options to display map information
+    - Paneling two plots on a page with 'subplots' command
+    - Drawing contours
+    - Explicitly setting the fill colors for contours
+    - Drawing contours over land only
+    - Using draw order resources to mask areas in a plot
+    - Adding a label bar
+    - Implementing best practices when choosing a color scheme
```
hCraker, line 12: As we have done with other plots, we can link the color examples that are already in the gallery. You could do something like "For more information on colormap choices, refer to [these examples](https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#colors)"

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,142 @@
...
+               loc='lower center', bbox_to_anchor=(0.5, -0.2), ncol=3)
+
+    # Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
+    gvutil.set_titles_and_labels(
+        ax,
+        maintitle=title,
+        maintitlefontsize=14)
+
+# #Plot first color map
+Plot(2, 3, (1,3), "land sea mask using 'atmos.nc'")
```
hCraker, line 91: I'm having a hard time figuring out why the `row` `col` and `pos` need to be these numbers. I'm looking at the `subplot` documentation and I'm still not sure. Can you add an elaborating comment or make some changes to make the code more intuitive.

hCraker (commented):
I'm glad you got the plots to panel correctly, but I'm still not sure why `row` `col` and `pos` need to be the numbers that they are. Please add clarifying comments or see if you can find a simpler way to do this. You could use `subplots` (notice it is plural) to create a grid with two rows and one column. It would return two sets of axes which you could manipulate individually. It may be a solution that is easier to read with fewer lines.

Also both of your helper functions are used only once, so I'm not sure if they need to be functions. You may be able to restructure that as well. This is good so far, it just needs some tidying up.

Plots/Masking/NCL_mask_5.py
michaelavs: Originally, my IDE was making the 2x1 subplot look like this:
image
where the size of the two subplots were not the same and there was overlap that wasn't going to be fixable with position calls. So those numbers were there because that was the only way to get the plots to be formatted correctly. But as of this morning, my IDE has decided it wants to make the 2x1 subplot correctly, and I have updated with that fix. In the IDE and after testing with sphinx the output is this:
image

hCraker:
> I'm not sure why it was so hard for me to get the subplots to work previously, but I FINALLY got the two plots to work with each other and build in sphinx nicely. So this should now be ready for review!

So I just remembered something. When you make a colorbar, if you use the `ax` argument the colorbar "steals" some of the area devoted to that set of axes. If you use `cax`, the colorbar fills the area for the specified axes. That is why your plots weren't scaling the same. The axes had the same area, but calling `plt.colorbar` doesn't add any new area to the plot it just reallocates some of the space meant for the plot.

If you want, you could play around with making a `gridspec` with four rows and one column. Using the argument `height_ratios` for `gridspec` you can make the second and forth row very short for the legend and colorbar and the other two the size you need for the maps. This forces the plots to scale the same and keeps the colorbar and legend from heavily influencing the formatting. This may be better practice, but it is more complicated. It would require rewriting this, which I'm not sure is the best approach. I'm bringing this up primarily for future reference.

hCraker (commented):
I see that you fixed the issue with the figures being different sizes after adding in the color bar! I'm glad you got that figured out; however, there are still a few things I think need to be changed about this example. Firstly, there are some inconsistencies with your formatting within this file. I recommend using `flake8` to catch them all.

In addition, I'm not sure how useful the helper functions you have made are since they are only called once and are named similar things. They also take arguments that are passed into `plt.subplot`, but the argument descriptions you've added in for `col`, `row`, and `pos` aren't very clear for someone who may be new to Python.

I see the potential usefulness of wrapping the code used to make each plot, but I think they need to be reworked a bit and renamed so they are easier to understand. I made an inline comment explaining this further.

anissa111 (approved):
This looks good, thank you!

erogluorhan:
Hi @hCraker latest commits seem to have addressed your concerns. Could you provide a newer review for this?

hCraker:
> Hi @hCraker latest commits seem to have addressed your concerns. Could you provide a newer review for this?

Sorry for the delay. When trying to run these files locally I'm encountering some errors that are preventing me from being able to properly review this. I will update you when I make some progress. @erogluorhan @michaelavs

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,120 @@
+"""
+NCL_mask_5.py
+=============
+This script illustrates the following concepts:
+    - Using cartopy.feature options to display map information
+    - Paneling two plots on a page with 'subplots' command
+    - Drawing contours
+    - Explicitly setting the fill colors for contours
```
hCraker, line 8: I don't think this concept is illustrated in this example anymore. You could reword this to refer to setting the fill colors for the map features on the first plot, or remove this bullet all together.

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,120 @@
...
+lakes = mpatches.Rectangle((0, 0), 1, 1, facecolor="plum")
+ocean = mpatches.Rectangle((0, 0), 1, 1, facecolor="blue")
+
+labels = ['Land ', 'Lakes', 'Ocean']
+
+# Add a legend to plot
+plt.legend([land, lakes, ocean], labels,
+           loc='lower center', bbox_to_anchor=(0.5, -0.2), ncol=3)
+
+# Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
```
hCraker, line 74:
```suggestion
# Use geocat.viz.util convenience function to set titles and labels without
# calling several matplotlib functions
```
This comment can't be completely read without scrolling to the right on the webpage

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,120 @@
...
+                    extendfrac='auto',
+                    use_gridspec=False,
+                    ticks=np.arange(240, 315, 5))
+
+cbar.ax.tick_params(labelsize=10)
+
+# Mask ocean data by changing adding ocean feature and changing its zorder
+ax1.add_feature(cfeature.OCEAN, zorder=10, edgecolor='k')
+
+# Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
```
hCraker, line 111:
```suggestion
# Use geocat.viz.util convenience function to set titles and labels without
# calling several matplotlib functions
```

Plots/Masking/NCL_mask_5.py
```diff
@@ -0,0 +1,120 @@
...
+NCL_mask_5.py
+=============
+This script illustrates the following concepts:
+    - Using cartopy.feature options to display map information
+    - Paneling two plots on a page with 'subplots' command
+    - Drawing contours
+    - Explicitly setting the fill colors for contours
+    - Drawing contours over land only
+    - Using draw order resources to mask areas in a plot
+    - Adding a label bar
```
hCraker, line 11:
```suggestion
    - Adding a color bar
```
Label bar is NCL terminology, so let's use equivalent Python terms instead

hCraker (commented):
I have made some suggestions to make the comments easier to read, but otherwise this looks good to me.

hCraker (approved):
This looks great! Thanks for making those small comment changes.

PR #174, opened 2020-07-22
Hov_3

### LINKED ISSUE #173: hov_3
Time vs. Longitude (Hovmueller) contour plot with hatching

plot: https://www.ncl.ucar.edu/Applications/Images/hov_3_lg.png
script: https://www.ncl.ucar.edu/Applications/Scripts/hov_3.ncl

clairefio:
Closes #173:

hov_3

### DIFF
--- a/Plots/Contours/NCL_hov_3.py
+++ b/Plots/Contours/NCL_hov_3.py
@@ -53,7 +53,6 @@ cf = ax.contourf(lon,
                  chi,
                  levels=[-7, -6],
                  colors='None',
-                 edge_color='gray',
                  hatches=['///'])
 
 # Draw contour lines at levels [-6, -4, -2, 0, 2, 4, 6, 8, 10]
--- a/Plots/Contours/NCL_hov_3.py
+++ b/Plots/Contours/NCL_hov_3.py
@@ -1,0 +1,105 @@
+"""
+NCL_hov_3.py
+================
+This script illustrates the following concepts:
+   - Creating a Hovmueller plot
+   - Hatching fill between contours
+
+See following URLs to see the reproduced NCL plot & script:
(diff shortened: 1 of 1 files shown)


clairefio:
> This looks good, but there is another type of hatching that I don't see on your plot. I have circled it in the NCL example. Other than that, I think this is great.
> image

Thanks for catching that! Fixed it.

Plots/Contours/NCL_hov_3.py
```diff
@@ -0,0 +1,113 @@
...
+cs = ax.contour(lon,
+                times,
+                chi,
+                levels=np.arange(-6, 12, 2),
+                colors='black',
+                linestyles="-",
+                linewidths=.2)
+
+# Set 0 level contour line to a thicker linewidth
+# cs.levels = [-6, -4, -2, 0, 2, 4, 6, 8, 10]
```
hCraker, line 68: Is this comment still needed?

clairefio: It was put in there to demonstrate how to access the levels attribute of a contour set, I tried to clarify it in my comment. Does the new comment make more sense?

clairefio:
> Nice work. A couple of things:
> 
> 1. In the NCL version of this example the contour lines in the hatched regions are still clearly visible. In the Python version they are not. Is there any control over the hatch marks (perhaps a different pattern) that would make the contour lines more visible?
> 2. I'm getting a warning message:
> 
> Plots/Contours/NCL_hov_3.py:57: UserWarning: The following kwargs were not used by contour: 'edge_color'
> hatches=['///'])

I fixed the error message. Unfortunately, you can't make the hatch marks any smaller unless you increase the density of the dots (which makes the area even darker). Instead I made the hatched areas light gray so you can still see the contour lines.

hov_3

PR #162, opened 2020-07-15
Update NCL_polar_1.py implementation for drawing longitude labels

hCraker:
NCL_polar_1.py originally from this PR #75 used hard coded values to position the longitude labels around the plot. A better method of doing this has been found, and it doesn't require any hard coding. This PR modifies NCL_polar_1.py to use the new method.

PR #177, opened 2020-07-23
ConOncon 5

clairefio:
conOncon_5

### DIFF
--- a/Plots/Contours/NCL_conOncon_5.py
+++ b/Plots/Contours/NCL_conOncon_5.py
@@ -108,9 +108,9 @@ p = slon.plot.contour(ax=ax,
 
 # Create a color list for each of the next 18 contours
 colorlist = ["crimson", "green", "blue", "yellow", "cyan", "hotpink",
-             "crimson", "skyblue", "navy", "lightyellow", "mediumorchid", "orange",
-             "slateblue", "palegreen", "magenta", "springgreen", "pink",
-             "forestgreen", "violet"]
+             "crimson", "skyblue", "navy", "lightyellow", "mediumorchid",
+             "orange", "slateblue", "palegreen", "magenta", "springgreen",
+             "pink", "forestgreen", "violet"]
 
 # Iterate through 18 different timesteps
 for x in range(18):
--- a/Plots/Contours/NCL_conOncon_5.py
+++ b/Plots/Contours/NCL_conOncon_5.py
@@ -83,12 +83,14 @@ yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]
 for xtick, ytick, label in zip(xticks, yticks, labels):
     ax.text(xtick, ytick, label, transform=ccrs.Geodetic(), fontsize=10)
 
-# Contour-plot data
+# Get slice of data at the 0th timestep
+p = ds.HGT.isel(time=0)
 
-newds = ds.isel(time=0)
-s = newds.HGT
-slon = gvutil.xr_add_cyclic_longitudes(s, "lon")
+# Use geocat-viz utility function to handle the no-shown-data
+# artifact of 0 and 360-degree longitudes
+slon = gvutil.xr_add_cyclic_longitudes(p, "lon")
 
+# Plot contour data at pressure level 5500 at the first timestep
 p = slon.plot.contour(ax=ax,
                       transform=ccrs.PlateCarree(),
                       linewidths=1.5,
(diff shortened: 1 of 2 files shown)


Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+ticks = np.arange(0, 210, 30)
+etick = ['0'] + [r'%dE' %
+                 tick for tick in ticks if (tick != 0) & (tick != 180)] + ['180']
+wtick = [
+    r'%dW' %
+    tick for tick in ticks if (
+        tick != 0) & (
+            tick != 180)]
+labels = etick + wtick
+xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -114, -140, -151]
```
michaelavs, line 83:
```suggestion
xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -117, -145, -151]
```
This will bring the ticks for 120W and 150W closer to the line they represent

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-7]
```
michaelavs, line 84:
```suggestion
yticks = [-3] + [-2] + [-1] + [-1] * 2 + [-1] + [-3] + [-7] + [-7] * 3 + [-5]
```
Same as above

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+# Set boundary to a circle
+theta = np.linspace(0, 2 * np.pi, 100)
+center, radius = [0.5, 0.5], 0.5
+verts = np.vstack([np.sin(theta), np.cos(theta)]).T
+circle = mpath.Path(verts * radius + center)
+ax.set_boundary(circle, transform=ax.transAxes)
+
+# Manipulate longitude labels (0, 30 E, 60 E, ..., 30 W, etc.)
+ticks = np.arange(0, 210, 30)
+etick = ['0'] + [r'%dE' %
```
hCraker, line 75: An updated way to draw the longitude labels can be found in #162. This makes sure the 0 and 180 marks are the same distance from the map as all of the other labels. It also allows you to easily change how far away all of the labels are from the map.

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+p = slon.plot.contour(ax=ax,
+                      transform=ccrs.PlateCarree(),
+                      linewidths=1.5,
+                      levels=[5500],
+                      colors='k',
+                      add_labels=False)
+
+# Create a color list for each of the next 18 contours
+colorlist = ["r", "green", "blue", "yellow", "cyan", "hotpink",
+             "r", "skyblue", "navy", "lightyellow", "mediumorchid", "orange",
```
hCraker, line 106: The red color is used twice. You might want to chose another shade of red. Additionally, since all of the other color names are fully written out, using "red" vs "r" would be easier for the user to interpret

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+
+# Get slice of data at the 0th timestep
+p = ds.HGT.isel(time=0)
+
+# Use geocat-viz utility function to handle the no-shown-data
+# artifact of 0 and 360-degree longitudes
+slon = gvutil.xr_add_cyclic_longitudes(p, "lon")
+
+# Plot contour data at pressure level 5500 at the first timestep
+p = slon.plot.contour(ax=ax,
```
hCraker, line 97: I see that you are plotting each line using a loop later on in the script but that the 0th time step uses a thicker line. To make this more readable, I would pull out the data for the 0th time step and plot that line just before you go into the loop but after you define the list of colors. Also, it would be beneficial to add a comment pointing out that the 0th timestep is using a thicker line and that is why it is outside of the loop.

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+                      transform=ccrs.PlateCarree(),
+                      linewidths=1.5,
+                      levels=[5500],
+                      colors='k',
```
hCraker, line 101: For readability use the full color name instead of the abreviation.
```suggestion
                      colors='black',
```

Plots/Contours/NCL_conOncon_5.py
clairefio: I just used the color list that was in the NCL script, I tried originally having red be listed as "red" but I think because it's a 3 element string python gets confused and tries to interpret it as an RGB value.

hCraker: That's strange. You could use 'firebrick' and 'crimson' as two distinct reds that have a readable name

Plots/Contours/NCL_conOncon_5.py
```diff
@@ -0,0 +1,138 @@
...
+ticks = np.arange(0, 210, 30)
+etick = ['0'] + [r'%dE' %
+                 tick for tick in ticks if (tick != 0) & (tick != 180)] + ['180']
+wtick = [
+    r'%dW' %
+    tick for tick in ticks if (
+        tick != 0) & (
+            tick != 180)]
+labels = etick + wtick
+xticks = [-0.8, 28, 58, 89.1, 120, 151, 182.9, -36, -63, -89, -117, -145, -151]
```
erogluorhan, line 83: @hCraker has recently found a nice way to avoid this hard-coded values. Please consult with her to modify here.

clairefio: Oops, I had added that in but it looks like my latest commit didn't go through.

PR #189, opened 2020-07-28
mask_2

### LINKED ISSUE #188: mask_2
Description: Show how to draw land masses on top of filled contour

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/mask_2.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/mask_2_lg.png

Category: Masking

hCraker:
Closes #188 
image


hCraker:
In the NCL example, the contours appear over lakes. I have played around with `zorder` but have been unable to achieve the same result. @clairefio @michaelavs have either of you done this before?

michaelavs:
> In the NCL example, the contours appear over lakes. I have played around with `zorder` but have been unable to achieve the same result. @clairefio @michaelavs have either of you done this before?

I did something semi-similar in my mask_5 example, but I was only drawing over land which covered the lakes so it may not work the same because I think you will be drawing over 2 separate objects. I believe I used [this](https://stackoverflow.com/questions/36525825/mask-ocean-or-land-from-data-using-cartopy) stack overflow question as a basis for my line of code, so It may be of use to you

hCraker:
> I did something semi-similar in my mask_5 example, but I was only drawing over land which covered the lakes so it may not work the same because I think you will be drawing over 2 separate objects. I believe I used [this](https://stackoverflow.com/questions/36525825/mask-ocean-or-land-from-data-using-cartopy) stack overflow question as a basis for my line of code, so It may be of use to you

I think the issue stems from the `LAND` feature in `cartopy` covering where lakes are. I would have to find a feature that draws land masses but doesn't cover lakes, and I don't think `cartopy` has that. I could use a shapefile, but I don't know if it would be worth adding to the example since shapefiles are a bit complicated. I'll continue to work on this but if you and @clairefio could review this PR with this in mind, that would be helpful.

clairefio:
> > I did something semi-similar in my mask_5 example, but I was only drawing over land which covered the lakes so it may not work the same because I think you will be drawing over 2 separate objects. I believe I used [this](https://stackoverflow.com/questions/36525825/mask-ocean-or-land-from-data-using-cartopy) stack overflow question as a basis for my line of code, so It may be of use to you
> 
> I think the issue stems from the `LAND` feature in `cartopy` covering where lakes are. I would have to find a feature that draws land masses but doesn't cover lakes, and I don't think `cartopy` has that. I could use a shapefile, but I don't know if it would be worth adding to the example since shapefiles are a bit complicated. I'll continue to work on this but if you and @clairefio could review this PR with this in mind, that would be helpful.

The rest of the script looks good! Here's an interesting thread: https://stackoverflow.com/questions/13796315/plot-only-on-continent-in-matplotlib Maybe you could tweak this solution to plot contour data on the lakes?

hCraker:
> The rest of the script looks good! Here's an interesting thread: https://stackoverflow.com/questions/13796315/plot-only-on-continent-in-matplotlib Maybe you could tweak this solution to plot contour data on the lakes?

Thank you for the suggestions, but that stack overflow thread uses `matplotlib.basemap` instead of `cartopy`. Thank you for searching for a solution.

Plots/Masking/NCL_mask_2.py
```diff
@@ -0,0 +1,83 @@
...
+                           levels=np.arange(216, 315, 3), extend='neither',
+                           add_colorbar=False, add_labels=False, zorder=0)
+plt.colorbar(contour, ax=ax, ticks=np.linspace(219, 303, 8),
+             orientation='horizontal', pad=0.075, drawedges=True)
+
+# Use geocat.viz.util convenience function to set axes limits & tick values
+gvutil.set_axes_limits_and_ticks(ax, xlim=(-180, 180), ylim=(-90,90),
+                                 xticks=np.linspace(-180, 180, 13),
+                                 yticks=np.linspace(-90, 90, 7))
+
```
michaelavs, line 71:
```suggestion
# Remove the degree symbol from tick labels
    ax.yaxis.set_major_formatter(LatitudeFormatter(degree_symbol=''))
    ax.xaxis.set_major_formatter(LongitudeFormatter(degree_symbol=''))
```

Plots/Masking/NCL_mask_2.py
```diff
@@ -0,0 +1,83 @@
+"""
+NCL_mask_2.py
+===============
+This script illustrates the following concepts:
+   - Using keyword zorder to mask areas in a plot
+   - Drawing filled land areas on top of a contour plot
+
```
michaelavs, line 7: I would mention why you used a different color scheme in this projection. I added a comment in the concepts illustrated section of [lcnative 1](https://github.com/NCAR/GeoCAT-examples/blob/master/Plots/MapProjections/NCL_lcnative_1_lg.py) that you may choose to copy over, or even just a short line about using best practices when choosing color schemes would be fine

jukent:
Here is an example of creating a custom cartopy shape feature:
https://geocat-examples.readthedocs.io/en/latest/gallery/Overlays/NCL_overlay_11a.html#sphx-glr-gallery-overlays-ncl-overlay-11a-py

You might need to make a new land file that excludes area in the lake feature.

hCraker:
> Here is an example of creating a custom cartopy shape feature:
> https://geocat-examples.readthedocs.io/en/latest/gallery/Overlays/NCL_overlay_11a.html#sphx-glr-gallery-overlays-ncl-overlay-11a-py
> 
> You might need to make a new land file that excludes area in the lake feature.

I will look into that and see what I can do. Thanks for the link

hCraker:
@jukent I've been having issues trying to mask only the land masses and not the lakes using cartopy and natural earth shape files. I tried looking for a shape file that already had what I wanted, but I couldn't find one. We may want to merge this PR as is and work on creating an appropriate shape file later.

PR #192, opened 2020-07-29
table_2

### LINKED ISSUE #191: table_2
Drawing adjacent tables/filling cells.

script: https://www.ncl.ucar.edu/Applications/Scripts/table_2.ncl
plot: https://www.ncl.ucar.edu/Applications/Images/table_2_lg.png

clairefio:
closes #191:

table_2


michaelavs:
Looks great so far! The only suggestion I have is to maybe make the fill color for the "Case A" column white instead of transparent/none. This is only because in my my IDE (Spyder) the background is black, so the numbers barely come through at all. But, because the transparent is true to the NCL version, I could also see just adding a note about the potential issue with the transparent background and mentioning filling it with white could also reproduce the NCL style projection. The projection looks fine when run through Sphinx so this suggestion is 100% up to you if you want to implement it or not!

clairefio:
> This looks good! For aesthetics, you might want to see if you can make the grid lines thinner, but I don't think it is that important.

Unfortunately there's no way to thin the grid lines in a matplotlib table.

clairefio:
> Looks great so far! The only suggestion I have is to maybe make the fill color for the "Case A" column white instead of transparent/none. This is only because in my my IDE (Spyder) the background is black, so the numbers barely come through at all. But, because the transparent is true to the NCL version, I could also see just adding a note about the potential issue with the transparent background and mentioning filling it with white could also reproduce the NCL style projection. The projection looks fine when run through Sphinx so this suggestion is 100% up to you if you want to implement it or not!

I think that's probably a good call, just changed it.

Plots/Tables/README.rst
```diff
@@ -0,0 +1,6 @@
+.. _table_examples:
+
+.. _table-examples-index:
+
+Tables
+============
```
erogluorhan, line 6: Please make the length of "=" signs exactly the same with above "Tables" string.

Plots/Tables/NCL_table_2.py
```diff
@@ -0,0 +1,103 @@
...
+
+# Set axis aspect ratio
+ax.set_aspect(2)
+
+# Make axis spines invisible
+fig.patch.set_visible(False)
+ax.axis('off')
+
+# Set row headers (first column)
+rows = ["", "", "SLP_ERA40", "Tsfc_ERA40", "Prc_GPCP", "Prc 30S-30N_GPCP",
```
erogluorhan, line 32: How about moving the initialization of variables `rows, rowcolors, cell_text, colors` into a new cell in between "import packages" and "plot" cells and calling it "generate data"?

PR #195, opened 2020-07-29
write_csv_1

### LINKED ISSUE #193: write csv 1
Link to NCL script: https://www.ncl.ucar.edu/Applications/Scripts/write_csv_1.ncl
no projection available (CSV file output)

michaelavs:
closes #193


hCraker:
@michaelavs Do you know what the resulting CSV file should look like? It would help me review this.

hCraker:
> @michaelavs Do you know what the resulting CSV file should look like? It would help me review this.

Never mind, I found the expected output in the [NCL gallery](https://www.ncl.ucar.edu/Applications/write_csv.shtml)

Plots/CSV/NCL_write_csv_1.py
```diff
@@ -0,0 +1,52 @@
...
+# Creat CSV file using 'pandas'
+
+# Create a data frame to contain all data
+df = pd.DataFrame([x1, x2, x3])
+
+# Transpose data frame so that it will be read in column by column not row by row
+df = df.T
+
+# Export data frame to csv file
+# setting 'header' and 'index' to False will remove a default crow and column
```
hCraker, line 49:
```suggestion
# setting 'header' and 'index' to False will remove a default row and column
```

michaelavs:
> This looks good. I only found a typo. Do you think opening the files and displaying them would be good for the gallery or would it be better to simply link the output files on the example page? All of the other examples clearly show the output and this example could benefit from doing the same.

I've been trying to figure this out myself. It is definitely weird not having a clear output file to display. From what I've found so far, we could have the pandas dataframe print out (with the header and index being in their "True" form) but I haven't really found a way to have the csv version to display. But even on the NCL webpage for this example, there isn't a true output file to see, it it just listed what the variables should look like in their respective CSV files. So I may play with it a bit more, but I think it may just end up having to be a comment that the output file will be stored locally and the user will have to manually open/view it. But if you have any experience in printing out a csv file in the kernel, definitely let me know because I would prefer to have it displayed for the user instead of manually done on their end.

hCraker:
> I've been trying to figure this out myself. It is definitely weird not having a clear output file to display. From what I've found so far, we could have the pandas dataframe print out (with the header and index being in their "True" form) but I haven't really found a way to have the csv version to display. But even on the NCL webpage for this example, there isn't a true output file to see, it it just listed what the variables should look like in their respective CSV files. So I may play with it a bit more, but I think it may just end up having to be a comment that the output file will be stored locally and the user will have to manually open/view it. But if you have any experience in printing out a csv file in the kernel, definitely let me know because I would prefer to have it displayed for the user instead of manually done on their end.

I was thinking of using a pandas dataframe too. I'm not sure if there is another way to print it out. At the very least, we could have a note at the beginning with the expected format types out like on the write_csv NCL gallery page

michaelavs:
> This looks good!
> 
> This might be an issue for a different PR, but is there anything that could be done to keep the csv files generated by this example from sticking around when we generate the html locally? Or at least keep git from trying to add them to tracking?

I was trying to find a fix for this, like maybe defining where to store the files after/with their creation, but haven't had much luck. It is kinda weird that they stick to the file because when I run the code locally, they save to the working directory, which is 'Downloads' for me. But then again, I suppose the working directory for Sphinx is within the `_build/html/gallery` directory that it builds from. I was also thinking maybe to define the working directory in the beginning and maybe that would work, but that wouldn't allow for a universally usable code as the directory has to be specific to the machine running the code. If it isn't casuing any problems with the build, I think that it would be best to investigate options in another PR. That PR could also address Sphinx build documentation/code so that we may be able to display the csv files on the webpage as well.

anissa111:
> I was trying to find a fix for this, like maybe defining where to store the files after/with their creation, but haven't had much luck. It is kinda weird that they stick to the file because when I run the code locally, they save to the working directory, which is 'Downloads' for me. But then again, I suppose the working directory for Sphinx is within the `_build/html/gallery` directory that it builds from. I was also thinking maybe to define the working directory in the beginning and maybe that would work, but that wouldn't allow for a universally usable code as the directory has to be specific to the machine running the code. If it isn't casuing any problems with the build, I think that it would be best to investigate options in another PR. That PR could also address Sphinx build documentation/code so that we may be able to display the csv files on the webpage as well.

Yeah, sounds good! This issue doesn't affect that the example here is working properly. I assume anybody actually trying to generate a csv would want that file to hang around, haha.

PR #197, opened 2020-07-30
color_1

### LINKED ISSUE #196: color_1
Description: Using NCL default colormap to create a filled contour plot

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/color_1.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/color_1_lg.png

Category: Contour

anissa111:
closes #196 

Original NCL plot:
image

Generated python plot:
image


anissa111:
image

anissa111:
> This example is a little tricky because the colormap is pretty close to the "rainbow", which we are trying to steer people away from. Thus, there is a tension between trying to reproduce NCL plots as closely as possible, but also not enforcing the same bad habits that many of the NCL plots exhibited (particularly wrt the rainbow colormap). Unfortunately, one of the goals for this example is to draw contours with "the default color map" (which, more accurately, is the default _NCL_ colormap). A couple of thoughts on how to move forward:
> 
> 1. Replace the colormap with a better choice. E.g. https://geocat-examples.readthedocs.io/en/latest/gallery/Contours/NCL_coneff_16.html#sphx-glr-gallery-contours-ncl-coneff-16-py
> 
> 
> and remove references to default colormap from this example, replacing with explanation of why a different colormap was chosen, possibly referencing the GeoCAT-examples that highlight colormap choices: https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#colors
> 
> 2. Leave the color map, but make clear that this is the default _NCL_ color map, and provide references to https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#colors for better colormap options.

My interpretation of including this example more closely aligned with your second suggestion, so that the default color map could be recreated in Python if desired. 

I could definitely include directions to the better colormap options, so that the original colormap recreation stands, but is included next to links to preferred options?

clyne:
OK. Let's just make clear in the documentation that it is the default *NCL* colormap that is being demonstrated, and let's add the caveat that this colormap may not be the best choice for interpretation of the data. See the Colors examples in the gallery for more discussion on colormap selection.

Plots/Contours/NCL_color_1.py
```diff
@@ -0,0 +1,115 @@
...
+                                 xticks=np.linspace(-180, 180, 13),
+                                 yticks=np.linspace(-90, 90, 7))
+
+# Use geocat.viz.util convenience function to add minor and major tick lines
+gvutil.add_major_minor_ticks(ax, labelsize=10)
+
+# Use geocat.viz.util convenience function to make latitude, longitude tick labels
+gvutil.add_lat_lon_ticklabels(ax)
+# Remove degree symbol from tick labels
+ax.yaxis.set_major_formatter(LatitudeFormatter(degree_symbol=''))
```
erogluorhan, line 56: I know removing the degree symbol is to make plot identical to NCL's original, but you could just draw them as we have preferred to do so for couple of other examples even though NCL plots were without the symbol.

If you decide to remove these lines, then do not forget to also remove `from cartopy.mpl.gridliner import LongitudeFormatter, LatitudeFormatter` since it will likely be no use.

Plots/Contours/NCL_color_1.py
```diff
@@ -0,0 +1,115 @@
...
+                  linewidths=0.4,
+                  linestyles='solid',
+                  add_labels=False,
+                  levels=levels,
+                  transform=ccrs.PlateCarree())
+
+plt.xlabel("")
+plt.ylabel("")
+plt.title("")
+plt.suptitle("Default Color", fontsize=16, y=0.9, fontweight='bold')
```
erogluorhan, line 112: `gvutil.set_titles_and_labels()` should be enough with setting its `maintitle` arguments instead of `plt.suptitle`. Again, may have a look at `NCL_coneff_16.py`.

Plots/Contours/NCL_color_1.py
```diff
@@ -0,0 +1,115 @@
...
+                             righttitle=ds.U.units,
+                             lefttitlefontsize=12,
+                             righttitlefontsize=12)
+
+# Import the default color map
+# Note: this may not be the best colormap to interpret the information, but was
+# included here in order to demonstrate how to recreate the original NCL colormap
+# For more information on colormap choices, see the Colors examples in the GeoCAT-examples documentation:
+# https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#colors
+newcmp = gvcmaps.amwg_blueyellowred
```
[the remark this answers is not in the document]
anissa111: I would agree, except when I use the `gvcmaps.gui_default`, it produces a plot that looks like:

image

Which is not what the original plot shows. I'm not sure why using the default colormap does not reproduce the default colormap example, but I'm sure the one I've used in this example is the one that actually reproduces the original plot, for whatever reason.

Plots/Contours/NCL_color_1.py
```diff
@@ -0,0 +1,115 @@
...
+ds.U.plot.contour(ax=ax,
+                  colors='k',
+                  alpha=0.8,
+                  linewidths=0.4,
+                  linestyles='solid',
+                  add_labels=False,
+                  levels=levels,
+                  transform=ccrs.PlateCarree())
+
+plt.xlabel("")
```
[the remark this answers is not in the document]
anissa111: Addressed, thank you!

Plots/Contours/NCL_color_1.py
anissa111: Okay, good! I've updated this to reflect the newer preferences.

Plots/Contours/NCL_color_1.py
anissa111: Addressed!

Plots/Contours/NCL_color_1.py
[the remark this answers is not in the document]
erogluorhan: Sorry, I did give you the wrong color-map; you should use `ncl_default` instead. It makes the plot identical.

Plots/Contours/NCL_color_1.py
```diff
@@ -0,0 +1,109 @@
...
+                                 yticks=np.linspace(-90, 90, 7))
+
+# Use geocat.viz.util convenience function to add minor and major tick lines
+gvutil.add_major_minor_ticks(ax, labelsize=10)
+
+# Use geocat.viz.util convenience function to make latitude, longitude tick labels
+gvutil.add_lat_lon_ticklabels(ax)
+
+# Import the default color map
+# Note: this may not be the best colormap to interpret the information, but was
```
erogluorhan, line 56: Could you move this important `Note` comment to the upper function docstring so that the user could initially see this?

anissa111: Added it, thanks for the suggestion!

PR #214, opened 2020-08-07
Updating sat_1 and sat_2 

michaelavs:
Updating projections to use convenience functions instead of hard code


michaelavs:
> I looked into where the `posx and posy should be finite values` errors were coming from and it seems to be coming from lines 94-95 in sat_1 and 90-91 in sat_2, though I'm not sure how to fix the issue. I agree with @hCraker that this should probably be resolved before merging.
> 
> Additionally, the Sphinx documentation for sat_1 does not seem to be generating correctly when I generate the docs locally. The preview and page itself exists, but no content is loaded on the page. Let me know if this is just my local generation issue or if the problem exists on your side, too.

It looks like the error is from not specifying levels in the code. I was able to update the code on line 91 by deleting 'highval' and 'lowval', but this leads to the map projections not coming out as similar to the NCL projection. This error also occurs in the original code, so I'm not sure if we chose to overlook this issue originally or what happened there. @hCraker do you recall anything like that being the case previously (overlooking the error)?

hCraker:
> It looks like the error is from not specifying levels in the code. I was able to update the code on line 91 by deleting 'highval' and 'lowval', but this leads to the map projections not coming out as similar to the NCL projection. This error also occurs in the original code, so I'm not sure if we chose to overlook this issue originally or what happened there. @hCraker do you recall anything like that being the case previously (overlooking the error)?

I was loosely involved in the creation for the sat_1 and sat_2 examples as well as the extrema utility function in geocat-viz, so I'm not sure why these warnings were put in. The only [mention](https://github.com/NCAR/geocat-viz/pull/17#discussion_r455275695) of warnings in the various PRs related to these examples was regarding adding a warning for when no extrema were found using the given parameters.

michaelavs:
@anissa111 @hCraker I've been playing around with both the overall codes and each util function to narrow down why there is a warning showing up and further how to maybe bypass the warning all together. So far, there isn't much luck. Yesterday I tried using the "warnings" function and having the warning be ignored, but that didn't work. I also tried to implement sat_2 code in sat_1 (there are less instances of the warning in sat_2) but in sat_1, the code would throw an error about specified levels not matching available levels, even if they did. When I bypassed that error, I wound up back up at square one because then I would still get 8 "posx and posy" warnings when running the code. Overall, the warnings are most likely coming from "Clabels" but within that function, I can't find a place that defining the levels actually resolved the warning. What is also semi-interesting is that 4/8 of the warnings come immediately after the "CLabels" function call, but the other 4/8 come in almost randomly (as in I can't locate what function is causing them to occur). If either of you have any time to dissect this and maybe see something I have missed, I would definitely appreciate and new set of eyes on this!

hCraker:
@michaelavs I did some digging and found [this](https://stackoverflow.com/questions/57674890/plotting-issue-matplotlib-valueerror-posx-and-posy-should-be-finite-values) stack overflow post that explains what this error means. It has to do with `np.nan` values causing problems with how `matplotlib.Text()` decides to place text. I think if we remove `np.nan` values from the arrays passed into `plotCLabels` and `plotELabels` that may fix the issue.

michaelavs:
> @michaelavs I did some digging and found [this](https://stackoverflow.com/questions/57674890/plotting-issue-matplotlib-valueerror-posx-and-posy-should-be-finite-values) stack overflow post that explains what this error means. It has to do with `np.nan` values causing problems with how `matplotlib.Text()` decides to place text. I think if we remove `np.nan` values from the arrays passed into `plotCLabels` and `plotELabels` that may fix the issue.

I've been looking into the link you provided and a couple of resources to try and get around this problem. I checked both projection's data for any NaN values and both returned there weren't any NaN values being used in the data. So far the most success I've had with changes in the code came from deleting the low and high val options, but this lead to the projection being cluttered with many level labels, instead of just the ones we want. Looking at the geoCAT-viz code for `plotCLabels` and `plotELabels', I feel there may be a way to modify the code to merge the success from removing low and high val along with fixing the output having all labels. I'll see where I get with this idea and report back if I am successful or need any assistance.

pilotchute:
@michaelavs building them now, they look fine? Is there a problem with them as built?

erogluorhan:
> @michaelavs building them now, they look fine? Is there a problem with them as built?

They seemed to have worked in my local and then I had pushed whatever I got done on my end. It still seems to be testing for that push, which may indicate some issues.

erogluorhan:
@anissa111 @hCraker FYI: Looks like this PR is ready for review now

anissa111 (approved):
This looks good, thank you!

PR #215, opened 2020-08-07
Turning polyg_19 inactive 

michaelavs:
Creating a non .py file to avoid build failures while exploring options for making code less massive/resource intensive


erogluorhan:
@michaelavs The following links show similar discussions about having Hawaii and Alaska near CONUS in a cartopy map. They are also iterating through shape files of the states but using different plotting methods than what we already used in this function. The packages and functions may be of help to reduce the running time of the current implementation.

PR #218, opened 2020-08-11
Pdf 1

### LINKED ISSUE #204: pdf_1
Description: A figure with three subplots illustrating different probability distribution functions

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/pdf_1.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/pdf_1_lg.png

Category: Probability Distribution Functions (PDF)

anissa111:
Closes #204
image


anissa111:
@michaelavs I am having trouble generating a gamma distribution that matches the NCL one. I have the type and general shape, but the scale seems to be off, any thoughts?

michaelavs:
> @michaelavs I am having trouble generating a gamma distribution that matches the NCL one. I have the type and general shape, but the scale seems to be off, any thoughts?

I think the code you uploaded may be corrupted (It isn't displaying in normal format), do you mind re-uploading it to see if that fixes it on the GitHub side? Afterwards, I'll definitely take a look!

anissa111:
> > @michaelavs I am having trouble generating a gamma distribution that matches the NCL one. I have the type and general shape, but the scale seems to be off, any thoughts?
> 
> I think the code you uploaded may be corrupted (It isn't displaying in normal format), do you mind re-uploading it to see if that fixes it on the GitHub side? Afterwards, I'll definitely take a look!

Ahh, yeah that keeps happening to my commits, something about line break styles. I've fixed it here and am going to look into why it keeps happening.

michaelavs:
> > > @michaelavs I am having trouble generating a gamma distribution that matches the NCL one. I have the type and general shape, but the scale seems to be off, any thoughts?
> > 
> > 
> > I think the code you uploaded may be corrupted (It isn't displaying in normal format), do you mind re-uploading it to see if that fixes it on the GitHub side? Afterwards, I'll definitely take a look!
> 
> Ahh, yeah that keeps happening to my commits, something about line break styles. I've fixed it here and am going to look into why it keeps happening.

So this may not be the best work around, but if you define a range in the np.histograms() call for the gamma distribution, you can make the random numbers not exceed 18 (as they appear to be doing). By setting `range=(0,10)` you are able to make the graph look more like the NCL version and it will not jump around so much between each run. What is also interesting, and may be causing the issue, is that in the NCL version the x-axis is in the scale of 0 to 0.18. I tried setting the range to that and it makes the projection a low level, nearly flat line. Just something you may be able to play around with and see what the outcomes are

anissa111:
After a long pause, I think I've sorted out what to do with this PR. I don't want to force the data to fit the original axes, since the original axes are arbitrary and the data used to form the plots is meant to be random. I've added some axis magic to allow for variable axes upper limits on the bottom two plots while maintaining the original tick spacing. I've also added using matplotlib.ticker as a concept demonstrated in this script.

image

Just note while testing, the script may produce different axes from the ones on the original ncl plot.

michaelavs (approved):
This runs correctly locally and I see no glaring issues with the code, so it looks good to me! I'm glad you were able to do some magic with the plot axis and get the code to work!

erogluorhan (changes requested):
A README.rst file should be added under the folder "ProbabilityDitributions" for this to be generated properly by Sphinx.

Other than that, the implementation and docs look great!

anissa111:
> A README.rst file should be added under the folder "ProbabilityDitributions" for this to be generated properly by Sphinx.
> 
> Other than that, the implementation and docs look great!

Thanks for catching this! The README.rst file is now included.

erogluorhan (approved):

PR #223, opened 2020-08-14
conLab_4

### LINKED ISSUE #222: conLab_4
Description: Changing the backgrounds of contour labels white

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/conLab_4.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/conLab_4_lg.png

Category: Contours

hCraker:
Closes #222 
image


Plots/Contours/NCL_conLab_4.py
```diff
@@ -0,0 +1,114 @@
...
+cbar = plt.colorbar(colors,
+                    ticks=np.arange(-12, 44, 4),
+                    orientation='horizontal',
+                    drawedges=True,
+                    aspect=12,
+                    shrink=0.8,
+                    pad=0.075)
+cbar.ax.tick_params(labelsize=14)  # Make the labels larger
+
+# By defualt, `clabel` labels every contour line
```
anissa111, line 105:
```suggestion
# By default, `clabel` labels every contour line
```

anissa111 (commented):
This is looking good so far! I just have some concerns about the contour labels. I'm playing around with them, and it seems like it's plotting clabels based on the entire projection range instead of the axes limits, which I think is why the labels in this example aren't in the most obvious locations. 

I'm concerned about this because if the region the end user would want to zoom in on does not contain enough/any contour labels, following this example would not be enough, if that makes sense. Also, it seems like the contour label positions _should_ be based on the observable plot area. Let me know your thoughts.

hCraker:
> This is looking good so far! I just have some concerns about the contour labels. I'm playing around with them, and it seems like it's plotting clabels based on the entire projection range instead of the axes limits, which I think is why the labels in this example aren't in the most obvious locations.
> 
> I'm concerned about this because if the region the end user would want to zoom in on does not contain enough/any contour labels, following this example would not be enough, if that makes sense. Also, it seems like the contour label positions _should_ be based on the observable plot area. Let me know your thoughts.

You're right in that `matplotlib` doesn't always place contour labels in the most appropriate places. I'll start looking into ways to improve this without having to manually specify where labels should go.

hCraker:
@anissa111 I found a viable solution. While there is no parameter for `clabels` to limit the region where the labels are drawn, masking the data works. This forces `clabels` to plot all of the labels in the region we want to focus on. Let me know if this looks good to you.

image

anissa111:
> @anissa111 I found a viable solution. While there is no parameter for `clabels` to limit the region where the labels are drawn, masking the data works. This forces `clabels` to plot all of the labels in the region we want to focus on. Let me know if this looks good to you.

This is a great solution! I am curious how you generated the graph in your comment, because generating mine locally in my IDE and through terminal when checking the make html makes this plot:
image

which I feel is not as clearly labeled as the plot you generated. 

This is mainly a curiosity, though, because I'm wondering what is causing the difference in generation. What happens on your machine when you generate the html? 

@michaelavs What plot do you see with the most recent changes?

hCraker:
@anissa111 The graph from my earlier comment was generated locally on my machine. When I use `make html` I get a plot that's exactly like the one in your comment. I'm going to look into whether or not this is due to differences in package versions between my local environment and the environment sphynx is using.

hCraker:
I'm not sure why the plot generated using `make html` differs from the one I'm making locally. I've checked the versions of the packages in my environment and in the environment we are using to generate the documentation, and they appear to be the same. Any thoughts on what's happening? @jukent @michaelavs

erogluorhan:
> I'm not sure why the plot generated using `make html` differs from the one I'm making locally. I've checked the versions of the packages in my environment and in the environment we are using to generate the documentation, and they appear to be the same. other than sphinx using conda version `4.8.3` instead of `4.8.4`. Any thoughts on what's happening? @jukent @michaelavs

If you are using an IDE (such as PyCharm or Spyder) in your local, it might be handling a number of rendering stuff on their own where you may not control. When you run `make html` command in your local though, it will make use of `Sphinx` gallery generator, which will help you see the exact copy of what would be generated in the actual gallery website. Hope it helps.

hCraker:
> > I'm not sure why the plot generated using `make html` differs from the one I'm making locally. I've checked the versions of the packages in my environment and in the environment we are using to generate the documentation, and they appear to be the same. other than sphinx using conda version `4.8.3` instead of `4.8.4`. Any thoughts on what's happening? @jukent @michaelavs
> 
> If you are using an IDE (such as PyCharm or Spyder) in your local, it might be handling a number of rendering stuff on its own where you may or may not control. When you run `make html` command in your local though, it will make use of `Sphinx` gallery generator, which will help you see the exact copy of what would be generated in the actual gallery website.
> 
> That being said, whenever finalizing an implementation or reviewing others' work, the correct thing to make 100% sure about the cosmetics of the final plot is to run `make html` in local.
> 
> Hope it helps.

I'll keep that in mind. Thank you.

hCraker:
I was unable to find a way to get matplotlib to draw the contour labels the way we wanted, so I decided to manually specify the coordinates for where to draw them. This isn't ideal since it is inconvenient, but it produces the figure we want. I added comments describing how to do this. @jukent @anissa111 you both are good to review this now.

hCraker:
Current plot produced with `make html`:
image

anissa111 (approved):
This looks good, thank you for finding a work around for the label positions.

PR #228, opened 2020-08-17
Radar 1

### LINKED ISSUE #200: radar_1
Description: Plotting radar data using (r, theta) coordinates

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/radar_1.ncl
NCL Plots: https://www.ncl.ucar.edu/Applications/Images/radar_1_1_lg.png
                  https://www.ncl.ucar.edu/Applications/Images/radar_1_2_lg.png

Category: Projections

anissa111:
Closes #200 
image
image


michaelavs:
Looks great! I noticed you left the grey version as a commented out block for the user to decide. Do you think having this would be necessary overall? If not, I think it would be fine to exclude it. Otherwise, I think making the two graphs display would be the best route.

anissa111:
> Looks great! I noticed you left the grey version as a commented out block for the user to decide. Do you think having this would be necessary overall? If not, I think it would be fine to exclude it. Otherwise, I think making the two graphs display would be the best route.

Thank you for the input! I've included a helper function so that both plots can be displayed in the sphinx docs.

Plots/MapProjections/NCL_radar_1.py
```diff
@@ -0,0 +1,127 @@
+"""
+NCL_radar_1.py
+===============
+This script illustrates the following concepts:
+   - Fitting radial data to a cartesian grid
+   - Creating a vertical colorbar
```
michaelavs, line 6:
```suggestion
   - Creating a horizontal colorbar
```
Colorbar is horizontal not vertical

anissa111: Whoops, good catch, thanks!

anissa111:
> Thanks for adding a helper function! Only one thing to address in the code just for accuracy. I think your branch may have brought in another branch/file (box_1), if you can delete that here/locally that should resolve that issue.

Ahh, addressed! Thank you!

PR #232, opened 2020-08-19
Revert 160 native 2

michaelavs:


michaelavs:
I merged the PR and then read the second half of your comment! I looked at the build fail and it is getting stuck on panel_35 due to an issue with the locator function. However, it doesn't have this problem when running build html locally. Also, now this is reverted, it is  doing the weird thing where it adds box_1 but now has added a third file. Not sure why this is happening on this PR because it was created via GitHub, not command line.

PR #233, opened 2020-08-19
Native_2 New

michaelavs:
New PR for native_2

### DIFF
--- a/Plots/MapProjections/NCL_native_2.py
+++ b/Plots/MapProjections/NCL_native_2.py
@@ -1,0 +1,90 @@
+"""
+NCL_native_2.py
+================
+
+This script illustrates the following concepts:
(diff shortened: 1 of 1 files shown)


michaelavs:
Here is the version of this pull request that should be able to pass the documentation build checks. Nothing has been changed on the code since the previous version was approved

michaelavs:
> This looks good, but for future reference do we know why changes in formatting can cause the documentation builds to fail?

Anissa was able to find out that there is actually a bug in the most recent matplotlib update which causes problems when using tight_layout() and constrained_layout(). So while more recent PRs weren't necessarily using either of those functions, the already merged projections that had used it were causing build failures. Anissa reverted the version we were using to build on as a temporary fix while matplotlib works on fixing the bug, which should be available in the next update.

PR #236, opened 2020-08-19
Documentation generation failure

### LINKED ISSUE #235: Documentation generation failure
Fix failing documentation builds

anissa111:
Closes #235 

Fixes sphinx documentation generation errors by:

- Removing `plt.tight_layout()` calls from the following files:
Plots/Bar/NCL_bar_7.py
Plots/Bar/NCL_unique_5.py
Plots/Contours/NCL_conOncon_1.py
Plots/Contours/NCL_conOncon_5.py
Plots/Contours/NCL_hov_3.py
Plots/MapProjections/NCL_sat_1.py
Plots/MapProjections/NCL_sat_2.py
Plots/MapProjections/NCL_sat_3.py
Plots/Panels/NCL_panel_13.py
Plots/Scatter/NCL_scatter_4.py
Plots/Streamlines/NCL_stream_1.py
Plots/Tables/NCL_table_2.py
Plots/Vectors/NCL_vector_4.py
Plots/XY/NCL_xy_18.py
Plots/XY/NCL_xy_2_1.py
Plots/XY/NCL_xy_7_2.py

- Removing `constrained_layout=True `and re doing plot layouts on:
Plots/Panels/NCL_panel_1.py
Plots/Panels/NCL_panel_3.py

### DIFF
--- a/Plots/Bar/NCL_bar_7.py
+++ b/Plots/Bar/NCL_bar_7.py
@@ -113,5 +113,4 @@ gvutil.set_titles_and_labels(ax,
                              ylabel="Number of Deaths")
 
 # Show the plot
-plt.tight_layout()
 plt.show()
--- a/Plots/Bar/NCL_unique_5.py
+++ b/Plots/Bar/NCL_unique_5.py
@@ -113,5 +113,4 @@ gvutil.set_titles_and_labels(ax,
                              ylabel="(" + u'\N{DEGREE SIGN}' + "C)")
 
 # Show the plot
-plt.tight_layout()
 plt.show()
--- a/Plots/Contours/NCL_conOncon_1.py
+++ b/Plots/Contours/NCL_conOncon_1.py
@@ -89,5 +89,4 @@ axRHS.yaxis.label.set_size(20)
 axRHS.tick_params('both', length=20, width=2, which='major', labelsize=18)
 
 # Show the plot
-plt.tight_layout()
 plt.show()
--- a/Plots/Contours/NCL_conOncon_5.py
+++ b/Plots/Contours/NCL_conOncon_5.py
@@ -145,8 +145,5 @@ gvutil.set_titles_and_labels(ax,
                              lefttitle=slon.long_name,
                              righttitle=slon.units)
 
-# Make tight layout
-plt.tight_layout()
-
 # Show the plot
 plt.show()
--- a/Plots/Contours/NCL_hov_3.py
+++ b/Plots/Contours/NCL_hov_3.py
@@ -116,5 +116,4 @@ ax.text(1,
                   facecolor='white',
                   edgecolor='black'))
 
-plt.tight_layout()
 plt.show()
--- a/Plots/MapProjections/NCL_sat_1.py
+++ b/Plots/MapProjections/NCL_sat_1.py
@@ -413,7 +413,4 @@ ax.text(0.40,
         fontsize=16,
         bbox=props)
 
-# Make layout tight
-plt.tight_layout()
-
 plt.show()
--- a/Plots/MapProjections/NCL_sat_2.py
+++ b/Plots/MapProjections/NCL_sat_2.py
@@ -401,7 +401,4 @@ gl = ax.gridlines(color='gray', linestyle='--')
 gl.xlocator = mticker.FixedLocator(np.arange(-180, 180, 20))
 gl.ylocator = mticker.FixedLocator(np.arange(-90, 90, 20))
 
-# Make layout tight
-plt.tight_layout()
-
 plt.show()
--- a/Plots/MapProjections/NCL_sat_3.py
+++ b/Plots/MapProjections/NCL_sat_3.py
@@ -170,5 +170,4 @@ plotOrthoTicks([(-120, 60), (-60, 82.5)], 'top')
 plotOrthoTicks([(-75, 16.0), (-60, 25.0), (-45, 29.0), (-30, 29.5),
                 (-15, 26.5)], 'bottom')
 
-plt.tight_layout()
 plt.show()
--- a/Plots/Panels/NCL_panel_1.py
+++ b/Plots/Panels/NCL_panel_1.py
@@ -37,6 +37,8 @@ ds = xr.open_dataset(gdf.get("netcdf_files/uv300.nc")).isel(time=1)
 projection = ccrs.PlateCarree()
 fig, ax = plt.subplots(3,
                        1,
+                       figsize=(7, 10),
+                       gridspec_kw=dict(hspace=0.3),
                        subplot_kw={"projection": projection})
 
 # Set figure size (width, height) in inches
@@ -81,9 +83,9 @@ ax[0].clabel(hdl, np.arange(0, 32, 8), fontsize="small", fmt="%.0f")
 # Use geocat.viz.util convenience function to add left and right title to the plot axes.
 gvutil.set_titles_and_labels(ax[0],
                              lefttitle="Zonal Wind",
-                             lefttitlefontsize=12,
+                             lefttitlefontsize=10,
                              righttitle=ds.U.units,
-                             righttitlefontsize=12)
+                             righttitlefontsize=10)
 
 # Panel 2 (Subplot 2)
 # Contour-plot V data (for borderlines)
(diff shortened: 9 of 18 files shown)


erogluorhan:
Thanks for the fix!
Unfortunately there are some issues appeared with directly removing `tight_layout` function calls or `constrained_layout=True`arguments: For example, please have a look at the before/after of `NCL_bar_7.py` second plot, and you will see that the y-axis label disappeared after this PR. This helped me remember the reason of why we had to add `tight_layout()` or `constrained_layout=True`; it was these type of behaviors. Another example is NCL_unique_5.py where the very bottom legend-like text box disappeared after this PR. I haven't checked the others but it is likely there are some issues with them as well.

I am requesting re-review of @michaelavs to let her know the issue, too, and you may both want to work together on the issue.

anissa111:
> Thanks for the fix! Thanks also for the clear list of which functions are changed with this PR; it is very helpful!
> 
> Unfortunately there are some issues appeared with directly removing `tight_layout` function calls or `constrained_layout=True`arguments: For example, please have a look at the before/after of `NCL_bar_7.py` second plot, and you will see that the y-axis label disappeared after this PR. This helped me remember the reason of why we had to add `tight_layout()` or `constrained_layout=True`; it was these type of behaviors. Another example is NCL_unique_5.py where the very bottom legend-like text box disappeared after this PR. I haven't checked the others but it is likely there are some issues with them as well.
> 
> I am requesting re-review of @michaelavs to let her know the issue, too, and you may both want to work together on the issue.

Oh no! When I was comparing before and after plots, I was looking for sizes/whitespace/layout changes. I'll poke around and consult more with @michaelavs.

michaelavs:
@erogluorhan, is the problem with bar_7 coming strictly from removing plt.tight_layout()? I tried to recreate the issue locally, but both projections look identical/the y-axis is still labeled and occurs. The only main difference I see is width of the plot is larger without tight_layout than with.

michaelavs:
@anissa111 For the two plots that use `constrained_layout`, the issue can be solved with a couple quick formatting changes in the code. for panel_1, all that needed changed was the figsize parameter, which I increased to (14,14). With panel_3, fig size needed to be changed to (14,14) as well, and the color bar needed a couple of kwargs updated. If you are ok with it, I will go ahead and track this branch locally and update those two projections manually. Additionally, I found an inconsistency with the color bar used on panel_3 (needs square edges), so I was thinking I could use this branch to update that as well.

erogluorhan:
> @erogluorhan, is the problem with bar_7 coming strictly from removing plt.tight_layout()? I tried to recreate the issue locally, but both projections look identical/the y-axis is still labeled and occurs. The only main difference I see is width of the plot is larger without tight_layout than with.

Is your local comparison based on the Sphinx-generated html file or the plot generated by your IDE (i.e. Spyder)? If IDE, then I can tell that every single IDE might have their own handling mechanisms to show the plot in nice ways. However, if it is based on the local Sphinx generated html file (i.e. with `make html`), then we should further look into it because we want local and remote Sphinx creations to be identical.

michaelavs:
> > @erogluorhan, is the problem with bar_7 coming strictly from removing plt.tight_layout()? I tried to recreate the issue locally, but both projections look identical/the y-axis is still labeled and occurs. The only main difference I see is width of the plot is larger without tight_layout than with.
> 
> Is your local comparison based on the Sphinx-generated html file or the plot generated by your IDE (i.e. Spyder)? If IDE, then I can tell that every single IDE might have their own handling mechanisms to show the plot in nice ways. However, if it is based on the local Sphinx generated html file (i.e. with `make html`), then we should further look into it because we want local and remote Sphinx creations to be identical.

Oh, ok now I see! I was running it only in the IDE, but just did a `make html` to verify my output and I now have the missing y-axis as well

erogluorhan:
> Oh, ok now I see! I was running it only in the IDE, but just did a `make html` to verify my output and I now have the missing y-axis as well

Oh, ok! Yes, development in any IDE is fine, but checking the plots from `make html` is helpful to see the actual final look of them. As I mentioned in the previous message, every IDE might have its own plot handling mechanisms (may be such as applying a hidden tight layout even if you do not explicitly call `plt.tight_layout()`), and this makes it hard to see the actual final look of plots from within IDE.

michaelavs:
@erogluorhan @anissa111 From what I've seen now on bar_7, it looks like again if we just increase the figure size (I used 10,10 for this one but maybe a different, "standard", range can work for all) and the Sphinx output works, even with `tight_layout` not being activated, and will also show the y-axis. If this is the best route, I think from there we can also tweak other plots as necessary (like panel_3 needed with its colorbar kwargs) and this could resolve the issues.

erogluorhan:
> @erogluorhan @anissa111 From what I've seen now on bar_7, it looks like again if we just increase the figure size (I used 10,10 for this one but maybe a different, "standard", range can work for all) and the Sphinx output works, even with `tight_layout` not being activated, and will also show the y-axis. If this is the best route, I think from there we can also tweak other plots as necessary (like panel_3 needed with its colorbar kwargs) and this could resolve the issues.

Thanks for this! It is a quick nice solution. A shortcoming with it is that since the figure size is bigger now, all the fonts (of labels, legends, titles, etc.) get much smaller. Therefore, all of them needs to be re-adjusted. I'd suggest to research a bit about what is recently changed with tight_layout, what are possible tight layout solutions for such plots, and in case of no solution, then I'd play with figure size and related fonts. This is because having a tight or constrained layout worked well before even if the figure size (6,5), not (10,10), and this makes me think there still would be potential layout solutions. Thoughts?

anissa111:
> > @erogluorhan @anissa111 From what I've seen now on bar_7, it looks like again if we just increase the figure size (I used 10,10 for this one but maybe a different, "standard", range can work for all) and the Sphinx output works, even with `tight_layout` not being activated, and will also show the y-axis. If this is the best route, I think from there we can also tweak other plots as necessary (like panel_3 needed with its colorbar kwargs) and this could resolve the issues.
> 
> Thanks for this! It is a quick nice solution. A shortcoming with it is that since the figure size is bigger now, all the fonts (of labels, legends, titles, etc.) get much smaller. Therefore, all of them needs to be re-adjusted. I'd suggest to research a bit about what is recently changed with tight_layout, what are possible tight layout solutions for such plots, and in case of no solution, then I'd play with figure size and related fonts. This is because having a tight or constrained layout worked well before even if the figure size (6,5), not (10,10), and this makes me think there still would be potential layout solutions. Thoughts?

So I'm seeing in the newest version (3.3.1) of matplotlib that was released seven days ago, some documentation around tight_layout changed:
image

There are a couple of changes to the way that it is implemented, too, but I do not know what specifically would have caused the bug. Here specifically is the [change log](https://github.com/matplotlib/matplotlib/compare/v3.3.0...v3.3.1)

anissa111:
Oh! Also since the examples actually causing the failure and not throwing errors were panel_1 and panel_3, could we revert the changes to the plots with changes and allow the errors since the plots still generate?

erogluorhan:
> Oh! Also since the examples actually causing the failure and not throwing errors were panel_1 and panel_3, could we revert the changes to the plots with changes and allow the errors since the plots still generate?
> 
> Maybe with the next matplotlib release some of the errors but not failures will be resolved?

Do you meant warnings when saying "error but not failure"?

erogluorhan:
> > Do you meant warnings when saying "error but not failure"?
> 
> Kind of. The actual print out for one of these is `WARNING: GeoCAT-examples/some-example failed to execute correctly: Traceback (most recent call last):`
> 
> Which if you then trace back the "failure to execute" comes back to an error: `ValueError: zero-size array to reduction operation minimum which has no identity`. With these examples, the plot itself still generates, though.
> 
> But the examples that are causing the actual failure to build the docs (I believe panel_1 and panel_3) have the same error, but cause the whole sphinx build to fail.

Thanks for the clarification! 

> With these examples, the plot itself still generates, though. 

This is interesting. There is a ValueError, but plots can generate. Did you have a chance to research about the error `ValueError: zero-size array to reduction operation minimum which has no identity`

> But the examples that are causing the actual failure to build the docs (I believe panel_1 and panel_3) have the same error, but cause the whole sphinx build to fail.

This is also interesting. Could you please elaborate if I am missing some thing here: What is the indicator of thinking only `panel_1 and panel_3` to cause the whole sphinx build to fail (or, why not others)?

anissa111:
> This is interesting. There is a ValueError, but plots can generate. Did you have a chance to research about the error `ValueError: zero-size array to reduction operation minimum which has no identity`

Yeah! But I didn't find anything that seemed implementable at the time. It has something to do with how matplotlib is using numpy I believe. Something like at some point it's trying to display an empty array? Maybe this has something to do with the new update feature that incorporates all artist objects on the axes instead of just labels and titles.

> This is also interesting. Could you please elaborate if I am missing some thing here: What is the indicator of thinking only `panel_1 and panel_3` to cause the whole sphinx build to fail (or, why not others)?

The difference between the panel examples and the other examples is that the panel examples use the `constrained_layout=True` tag instead of calling `tight_layout()`. Interestingly, there is another example in the gallery that uses `constrained_layout` that is still generating fine (Plots/Contours/NCL_eof_1_1.py), though I do not have an explanation as to why that one is still fine.

erogluorhan:
Thank you very much for these clarifications! I will have a look into it.

PR #242, opened 2020-08-28
Wrf_zoom_1_2

### LINKED ISSUE #241: wrf_zoom_1_2
NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/wrf_zoom_1.ncl
NCL Plot:https://www.ncl.ucar.edu/Applications/Images/wrf_zoom_1_2_lg.png

michaelavs:
Closes #241 
Original Plot:https://www.ncl.ucar.edu/Applications/Images/wrf_zoom_1_2_lg.png

Recreated Plot:
image


michaelavs:
@anissa111 @hCraker I've added you both for review on this PR, but the datafile hasn't been added to the datafiles repo yet, just to get the 'add reviewer' stage out of the way for this PR. So you guys don't have to review this one for the time being, unless you would like to just look a the code. I can also individually send the datafile to you if you would prefer so you can fully review the PR in the mean time.

erogluorhan:
> @anissa111 @hCraker I've added you both for review on this PR, but the datafile hasn't been added to the datafiles repo yet, just to get the 'add reviewer' stage out of the way for this PR. So you guys don't have to review this one for the time being, unless you would like to just look a the code. I can also individually send the datafile to you if you would prefer so you can fully review the PR in the mean time.

Once datafile and PR is actually ready for review, please do not forget to update "Draft" status and to add a label of "Ready for review" for the reviewers' convenience. Thanks!

Plots/WRF/NCL_WRF_zoom_1_2.py
```diff
@@ -0,0 +1,128 @@
...
+fig = plt.figure(figsize=(12,12))
+ax = plt.axes(projection=cart_proj)
+
+# Add features to the projection
+states = NaturalEarthFeature(category="cultural",
+                             scale="50m",
+                             facecolor="none",
+                             name="admin_1_states_provinces_shp")
+
+ax.add_feature(cfeature.LAKES, edgecolor='lightgray', facecolor='gold')
```
hCraker, line 68: Is this line needed? I don't see any lakes drawn on the original plot.

hCraker (commented):
This is a great start! I just have a few questions. Could you explain why there are the large yellow dots on the Python plot? Two of them exist in the original NCL plot, but the one on the northern California boarder is unique to the Python version. Also, what value is represented by the yellow color? In the NCL version, the spots appear to represent values between 6 and 8, but there is ambiguity in the Python version.

Additionally, the NCL plot has values from the WRF model written out at the bottom of the figure. Can you add that text to the Python version?

Plots/WRF/NCL_WRF_zoom_1_2.py
michaelavs: This line came from the outline that is around Lake Tahoe, which would imply some call to adding lakes to the projection. After you pointed this out, I double checked the NCL code and original NCL projection to see if there was any mention of lakes. There was not, so I wanted to double check that maybe the color is representative to the temperature over the lake, not just a geographic overlay. It appears this is what the reasoning is, but I still don't know why Tahoe is outlined and no other lakes are. That being said, I did also run the projection with the default NCL color scheme to double check consistency with the Python version, so I am attaching that here for your reference as well (but the color scheme will remain magma in the code). In the mean time, I will go ahead and take out the line for adding lake outlines/fills!

image

hCraker: Sounds good thank you!

anissa111 (approved):
This looks good!

hCraker (commented):
This looks good to me. One thing I want to point out before approving this is that in the original NCL plot, there is some information at the bottom of the figure about the data (shown below). Have you attempted to add that into the Python example yet?

image

Plots/WRF/NCL_WRF_zoom_1_2.py
```diff
@@ -0,0 +1,128 @@
...
+gl.ylocator = mticker.FixedLocator([38, 39, 40, 41, 42])
+gl.xlabel_style = {"rotation": 0, "size": 15}
+gl.ylabel_style = {"rotation": 0, "size": 15}
+gl.xlines = False
+gl.ylines = False
+
+# Add titles to the plot
+plt.title("2m Dewpoint Temperature (C)", loc='left', y=1.02, size=12)
+plt.title("Zoomed in plot", loc='center', y=1.1, size=14)
+plt.title("OUTPUT FROM WRF V3.3.1 MODEL \nWE=481; SN=547; Levels=32; Dis=2km ; Phys Opt=2; PBL Opt=1; Cu Opt=0",
```
hCraker, line 124: Is any of the model information in the data file? I'm curious if we can parse out this info versus hard coding it in, that way someone using a different dataset won't have to change this line manually.

hCraker (commented):
Looks good. I made one last comment about the model information at the bottom of the figure.

Plots/WRF/NCL_WRF_zoom_1_2.py
michaelavs: So the answer is yes and no. The model information does exist, but I cannot find a way to format it in the way that we see in the NCL version. The problem comes (for me) from the combination of both text and `getattr` from the dataset. I've been playing around with different methods to extract the variables (xarray and netcdf Dataset) and tack them onto the end of text, but it seems like you can have one or the other, but not both. If you do have both, they will appear as a list such as: ['WE=', "481] or ('WE=', '481') without a way to be read as "WE=481" as is seen in the NCL version.

The version I am pushing now is the closest I could get which has the title of the projection taken from the data, but still hard codes the second line of data. Let me know what you think, or if you know of a way to combine text with data attributes.

hCraker: I'm not sure if I totally understand what issue you're having. Couldn't you pull the numbers from the file using `getattr` and then use the String `format()` method to mesh the data with the text you want? For example, say you have already pulled `WE` and `SN` from the file. You could do the following:
```
str_format = "WE={}; SN={};"
print(str_format.format(WE, SN))
```
Which prints out `WE=481; SN=547;`. You could expand this to include the other variables.

hCraker (approved):
This looks great! Thanks for doing the tedious string formatting. It looks very polished now.

hCraker:
Also, I noticed that your builds are failing. There is a new version of conda and the warning about that is causing the failure in this PR and in others. I think we are good to merge this PR once we update conda.

michaelavs:
> Also, I noticed that your builds are failing. There is a new version of conda and the warning about that is causing the failure in this PR and in others. I think we are good to merge this PR once we update conda.

I noticed this too. I'll talk to Orhan about it a little later today and hopefully we will get it resolved!

Issue #240, opened 2020-08-27
h_lat_6


hCraker:
Description: a filled contour plot showing zonal wind with y axis scales in hPa and in km. Note the data is already in pressure levels so no interpolation from hybrid coordinates to pressure levels is needed.

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/h_lat_6.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/h_lat_6_lg.png

Category: Contours

PR #246, opened 2020-08-28
h_lat_6

hCraker:
Closes: #240 
Original NCL Plot:
image

Python Recreation:
image


Plots/Contours/NCL_h_lat_6.py
```diff
@@ -0,0 +1,120 @@
...
+# how to make this work properly yet
+axRHS = ax.twinx()
+axRHS.set_ylim(0, 32)
+axRHS.set_yticks(np.arange(4, 32, 4))
+axRHS.tick_params(labelsize=10)
+axRHS.set_ylabel('Height (km)')
+axRHS.yaxis.label.set_size(14)
+
+# Force the plot to be square by setting the aspect ratio to 1
+ax.set_box_aspect(1)
```
michaelavs, line 116: Locally, I get an Attribute Error of: `'AxesSubplot' object has no attribute 'set_box_aspect'`. Is this consistent on your side too @anissa111? I still have an output projection even with the Attribute Error.

hCraker: That's really odd because when I'm running it locally, I don't get a single warning or error.

Plots/Contours/NCL_h_lat_6.py
```diff
@@ -0,0 +1,120 @@
+"""
+NCL_conOncon_1.py
```
anissa111, line 2:
```suggestion
NCL_h_lat_6.py
```

Plots/Contours/NCL_h_lat_6.py
anissa111: I am also not getting that error locally in my IDE or when I build the html, that is strange.

michaelavs: I looked the error up and it looks like it may be due to a version of matplotlib I am using? Not sure completely as we should all be on the same version, but if it isn't happening for Anissa, then I think it's fine to ignore!

michaelavs: I'm also using Spyder, so it could maybe be an IDE thing too (if you are both not using Spyder)

anissa111: I just recently switched over to PyCharm, so maybe. Did you remake the geocat-examples conda environment after we switched back to the previous version?

Plots/Contours/NCL_h_lat_6.py
hCraker: Thanks for catching these. I always start by recycling some code from a similar example and lately I've been forgetting to change the names.

Plots/Contours/NCL_h_lat_6.py
hCraker: For reference I recently rebuilt my conda environment, and I'm using Visual Studio Code as my IDE.

hCraker:
Updated plot:
image

michaelavs (approved):
Looks good!

Issue #248, opened 2020-08-29
xy_13


hCraker:
Description: An XY plot with added error bars

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/xy_13.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/xy_13_lg.png

Category: XY

PR #249, opened 2020-08-29
xy_13

hCraker:
Closes: #248 
Original NCL Plot:
image

Python equivalent:
image


Plots/XY/NCL_xy_13.py
```diff
@@ -0,0 +1,79 @@
+"""
+NCL_xy_12.py
+===============
+This script illustrates the following concepts:
+   - Adding error bars on an XY plot
```
michaelavs, line 5: I think this needs to be indented one space

Plots/XY/NCL_xy_13.py
```diff
@@ -0,0 +1,79 @@
+"""
+NCL_xy_12.py
```
michaelavs, line 2:
```suggestion
NCL_xy_13.py
```

Plots/XY/NCL_xy_13.py
hCraker: Good catch! I just fixed it

michaelavs (approved):

hCraker:
@erogluorhan @michaelavs Given the new workflow, should I tag this as ready to merge now or do I still need to request review from @jukent?

erogluorhan:
> @erogluorhan @michaelavs Given the new workflow, should I tag this as ready to merge now or do I still need to request review from @jukent?

In general, we should be good with 2 approving reviews without having several people in the loop from now on. However, my concern in particular with this PR is that although the PR still did not pass build tests (i.e. it is not all green, which means we cannot guarantee if it would not break the `master` branch when merged), we have two approving reviews. In other words, as a rule of thumb, please make sure to see PRs all green with their build tests in order to approve them.

hCraker:
> In general, we should be good with 2 approving reviews without having several people in the loop from now on. However, my concern in particular with this PR is that although the PR still did not pass build tests (i.e. it is not all green, which means we cannot guarantee if it would not break the `master` branch when merged), we have two approving reviews. In other words, as a rule of thumb, please make sure to see PRs all green with their build tests in order to approve them.

Okay thank you for clarifying. Have we made any progress on figuring out why the build tests are stalling?

PR #259, opened 2020-09-02
Reopening dataonmap_10

### LINKED ISSUE #253: dataonmap_10
NCL script: https://www.ncl.ucar.edu/Applications/Scripts/datagrid_4.ncl
NCL Projection: https://www.ncl.ucar.edu/Applications/Images/datagrid_4_lg.png

michaelavs:
closes #253 
NCL Plot:
image

Python Plot:
image


hCraker:
@michaelavs I'm having some issues running this locally. On line 30 when the data is being read in, I get the following error `OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\hcraker\\AppData\\Local\\geocat\\geocat\\Cache\\netcdf_files\\wrfout_d01_2003-07-15_00:00:00'`.
I believe this is due to colons being a reserved character in Windows so they cannot be used for filenames. Is anyone else having this issue or have an idea for a work around? I'm currently searching for the file in my cache so I can attempt to rename it.

michaelavs:
> @michaelavs I'm having some issues running this locally. On line 30 when the data is being read in, I get the following error `OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\hcraker\\AppData\\Local\\geocat\\geocat\\Cache\\netcdf_files\\wrfout_d01_2003-07-15_00:00:00'`.
> I believe this is due to colons being a reserved character in Windows so they cannot be used for filenames. Is anyone else having this issue or have an idea for a work around? I'm currently searching for the file in my cache so I can attempt to rename it.

Thats interesting. I think on the actual NCL landing site for this projection it mentioned two versions of the same file and one had colons and the other had underscores, so I think you're right that that is the problem. If you need to rename locally, I think it would be best to create a PR that renames the file in the geocat-datafiles repo so that this doesn't happen for users too, and I think that the use of the underscore in naming  is universal between mac and windows machines anyways.

hCraker:
> Thats interesting. I think on the actual NCL landing site for this projection it mentioned two versions of the same file and one had colons and the other had underscores, so I think you're right that that is the problem. If you need to rename locally, I think it would be best to create a PR that renames the file in the geocat-datafiles repo so that this doesn't happen for users too, and I think that the use of the underscore in naming is universal between mac and windows machines anyways.

I have opened an [issue](https://github.com/NCAR/geocat-datafiles/issues/4) for this in geocat datafiles.

Plots/WRF/NCL_dataonmap_10.py
```diff
@@ -0,0 +1,93 @@
+"""
+NCL_datagrid_4_lg.py
+==================
```
anissa111, line 3:
```suggestion
====================
```

Plots/WRF/NCL_dataonmap_10.py
```diff
@@ -0,0 +1,93 @@
...
+lats, lons = latlon_coords(q2)
+
+# The `get_cartopy` wrf function will automatically find and use the 
+# intended map projection for this dataset 
+fig = plt.figure(figsize=(10,10))
+cart_proj = get_cartopy(q2)
+ax = plt.axes(projection=cart_proj)
+
+# Add filled contours
+plt.contourf(to_np(lons),
```
anissa111, line 46: Could we include something here or in the top docstring linking back to the color examples/tutorials explaining best practices?

Plots/WRF/NCL_dataonmap_10.py
```diff
@@ -0,0 +1,93 @@
+"""
+NCL_datagrid_4_lg.py
```
anissa111, line 2:
```suggestion
NCL_dataonmap_10.py
```

anissa111 (commented):
This is looking good so far! I've included a couple small comments below.

Also, I noticed that there were originally two plots for this example, and the second one overlaid a coastline, do you think adding that option here would add anything to this example?

michaelavs:
> This is looking good so far! I've included a couple small comments below.
> 
> Also, I noticed that there were originally two plots for this example, and the second one overlaid a coastline, do you think adding that option here would add anything to this example?

I was thinking of adding that feature in this example, but not adding the second projection. Overall, I think the capabilities it is showing aren't relevant in python but open to more discussion later. I think adding the coastline feature would help combine the relevant points from each NCL projection into one singular python example.

Plots/WRF/NCL_dataonmap_10.py
michaelavs: I just added a line in the "concepts illustrated" portion which links to one of the colormap examples

Plots/WRF/NCL_dataonmap_10.py
```diff
@@ -0,0 +1,99 @@
...
+import cartopy.feature as cfeature
+from wrf import (getvar, to_np, latlon_coords, get_cartopy)
+
+import geocat.datafiles as gdf
+from geocat.viz import util as gvutil
+
+###############################################################################
+# Read in the data
+
+wrfin = Dataset(gdf.get("netcdf_files/wrfout_d01_2003-07-15_00:00:00"), decode_times=True)
```
hCraker, line 33: Once the filename is changed in geocat-datafiles, we should update this example to use the name without colons so users have an OS agnostic example.

hCraker (commented):
This looks good so far. I left a note about the file name and found a solution for the colorbar levels.

Plots/WRF/NCL_dataonmap_10.py
michaelavs: I updated this in the most recent push, but you may have to do a `git pull` on the geocat-datafiles before it will reflect locally

hCraker (commented):
For some reason I am getting errors on my side that we may need to address before merging. They are the same errors that are showing up when the build check fails.

anissa111 (commented):
I think I've isolated the error to line 49: `ax.add_feature(cfeature.LAND, facecolor="", edgecolor="k")`, though I can't pin down why it's happening here. This is looking good so far, any ideas on this?

michaelavs:
> I think I've isolated the error to line 49: `ax.add_feature(cfeature.LAND, facecolor="", edgecolor="k")`, though I can't pin down why it's happening here. This is looking good so far, any ideas on this?

I would venture that it may be related to when I was getting an error on one of Rose's examples that wasn't occurring on either of your sides. When that occurred, I double checked which matplotlib version my environment was using and when using conda list, it is version 3.3.2, where as the conda_environment is on 3.3.0. However, when I run `make html` locally, the build is stuck on h_lat_6 and gives an attribute error for `set_box_aspect`.  So I'm not sure if looking into h_lat_6 would be a good place to start, or if the matplotlib version may be the issue

michaelavs:
@anissa111 @hCraker I just reverted my matplotlib version to 3.3.0 and now I get `ValueError: Expected 2-dimensional array, got 1`, so version was the issue. I'll look into it a bit more now and see if I can find a fix

hCraker (commented):
This is what is produced when I run `make html` locally. I checked the `matplotlib` version I'm using and it's `3.3.0`. This looks good so far, just don't forget to add the latitude and longitude tick marks. In addition, please address some suggestions I made a few weeks ago. One is regarding the name of the file in the leading comment block and the other is asking for a comment to clarify what you're doing on line 40.

image

michaelavs:
> This is what is produced when I run `make html` locally. I checked the `matplotlib` version I'm using and it's `3.3.0`. This looks good so far, just don't forget to add the latitude and longitude tick marks. In addition, please address some suggestions I made a few weeks ago. One is regarding the name of the file in the leading comment block and the other is asking for a comment to clarify what you're doing on line 40.
> 
> image

I hadn't seen the two comments about the name and adding a comment about line 40, so apologies for not addressing those in versions pushed since then! 

I've been working to add gridlines to the projection and came across something weird with how the data is being read in. It appears the latitude lines are being read in at a value of .00001:1, so they would need to be multiplied by a value of 100,000 to be accurate. The longitude lines are even more confusing as they are at between 94.0009 and 94.0011 when the values should be between 80 and 105. I'm not really sure what to do from here for adding lines of latitude and longitude. 

A couple of things I have tried are: 
-Adding a simple ax.gridlines() to see what the output would be then adding arguments line by line to see what worked and didn't. Basically once I got to adding in the specific tick markers, the function no longer worked.
-Using geocat-viz functions. This didn't work at all and added nothing new to the projection
-Adding boundary arguments to ax
-In plt.contourf, using only the q2 call, to see if the issue was coming from the use of to_np or the latlon_coords functions. 
-Adding a "zorder" argument to ax.gridlines

I'm going to push a version with the "best" tick labels I've had so far so you can also see what the code is and maybe find what needs to be added/changed to make the lat/lon lines/ticks appear accurately

Here is a current version of the projection:
image

hCraker:
> I've been working to add gridlines to the projection and came across something weird with how the data is being read in. It appears the latitude lines are being read in at a value of .00001:1, so they would need to be multiplied by a value of 100,000 to be accurate. The longitude lines are even more confusing as they are at between 94.0009 and 94.0011 when the values should be between 80 and 105. I'm not really sure what to do from here for adding lines of latitude and longitude.

@michaelavs Regarding your issued with the latitude and longitude data, remember that this data is natively on a Mercator projection so the coordinates that you printed out may be correct but not what you were expecting since it's not on a Plate Carree projection. As to how to fix your gridline issue, I'm not sure what you could do. I did notice that the `ax` is of type `wrf.projection.MercatorWithLatTS` instead of the usual `cartopy.crs.Mercator` which may be causing some issues. This [website](https://fabienmaussion.info/2018/01/06/wrf-projection/) has information about WRF map projections, and it has a cartopy example at the bottom of the page. Hopefully it is of some use to you.

michaelavs:
> > I've been working to add gridlines to the projection and came across something weird with how the data is being read in. It appears the latitude lines are being read in at a value of .00001:1, so they would need to be multiplied by a value of 100,000 to be accurate. The longitude lines are even more confusing as they are at between 94.0009 and 94.0011 when the values should be between 80 and 105. I'm not really sure what to do from here for adding lines of latitude and longitude.
> 
> @michaelavs Regarding your issued with the latitude and longitude data, remember that this data is natively on a Mercator projection so the coordinates that you printed out may be correct but not what you were expecting since it's not on a Plate Carree projection. As to how to fix your gridline issue, I'm not sure what you could do. I did notice that the `ax` is of type `wrf.projection.MercatorWithLatTS` instead of the usual `cartopy.crs.Mercator` which may be causing some issues. This [website](https://fabienmaussion.info/2018/01/06/wrf-projection/) has information about WRF map projections, and it has a cartopy example at the bottom of the page. Hopefully it is of some use to you.

I looked at the website you linked and found that the necessary change was just defining the projection in `ax` as PlateCarree. So I now have the code updated to show the lat/lon labels and lines. The `get_cartopy` function ended up not being necessary so I also removed that from the code

hCraker (approved):
I'm glad you got the latitude and longitude labels figured out. This looks good.

anissa111 (approved):
This looks good!

Apologies if I missed it in the comments, but is there a reason that the NCL style tick marks weren't used, but a grid was?

michaelavs:
> This looks good!
> 
> Apologies if I missed it in the comments, but is there a reason that the NCL style tick marks weren't used, but a grid was?

For some reason, the geoCAT-viz functions will sometimes not overlay on the projection. There are a couple of other examples where we did this work around as well, one example is [native_2](https://geocat-examples.readthedocs.io/en/latest/gallery/MapProjections/NCL_native_2.html?highlight=native). We decided to use the full grid lines instead of just floating lat/lon ticks to make it easier to distinguish where the lat/lon coordinates were lying.

PR #261, opened 2020-09-03
panel_10

### LINKED ISSUE #260: panel_10
Description: Paneling plots with multiple axes

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/panel_10.ncl and https://www.ncl.ucar.edu/Applications/Scripts/panel_attach_10.ncl

NCL Plots: https://www.ncl.ucar.edu/Applications/Images/panel_10_lg.png and https://www.ncl.ucar.edu/Applications/Images/panel_attach_10_lg.png

Category: Panel

hCraker:
Closes #260 
Original plots: https://www.ncl.ucar.edu/Applications/Images/panel_10_lg.png and https://www.ncl.ucar.edu/Applications/Images/panel_attach_10_lg.png

Python Recreation thus far:
image


Plots/Panels/NCL_panel_10.py
```diff
@@ -0,0 +1,258 @@
...
+ax1.contour(lon,
+            times,
+            chi,
+            levels=np.arange(-12, 13, 2),
+            colors='black',
+            linestyles='solid',
+            linewidths=.5)
+
+# Draw filled contours
+cf = ax1.contourf(lon,
```
michaelavs, line 76:
```suggestion
ax1.contourf(lon,
```
cf is defined and unused, but contourf is used. So just removing the variable definition from this line works

Plots/Panels/NCL_panel_10.py
```diff
@@ -0,0 +1,258 @@
...
+    ax1.contour(lon,
+                times,
+                chi,
+                levels=np.arange(-12, 13, 2),
+                colors='black',
+                linestyles='solid',
+                linewidths=.5)
+
+    # Draw filled contours, save the mappable to create colorbar later
+    cf = ax1.contourf(lon,
```
michaelavs, line 162:
```suggestion
    ax1.contourf(lon,
```
cf is defined and unused, but contourf is used. So just removing the variable definition works here

Plots/Panels/NCL_panel_10.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Use geocat.viz.util convenience function to add minor and major tick lines
+    gvutil.add_major_minor_ticks(ax1,
+                                 x_minor_per_major=2,
+                                 y_minor_per_major=3,
+                                 labelsize=10)
+
+    # Remove tick marks on right side of ax1
+    ax1.tick_params('y', which='both', right=False)
+
```
michaelavs, line 187: I'm not sure if you were able to find another way to remove the degree symbol other than with the `ax.xaxis.set_major_formatter()` used in the past (looking at the first plot), but I would add that feature to the 4 panel plot as well for NCL consistency

Plots/Panels/NCL_panel_10.py
hCraker: `cf` is used to make the colorbar on line 250. I will change this function to return the color mappable so it can be used later

Plots/Panels/NCL_panel_10.py
hCraker: Fixed! Thanks for pointing it out

anissa111 (approved):
This looks great, Rose!

I'm getting a runtime waring when running it on my IDE and when generating the html:
`RuntimeWarning: Mean of empty slice return np.nanmean(a, axis=axis, dtype=dtype)`
But it seems like this is a known issue and could be suppressed, but I'm not overly concerned about addressing it here.

hCraker:
> This looks great, Rose!
> 
> I'm getting a runtime waring when running it on my IDE and when generating the html:
> `RuntimeWarning: Mean of empty slice return np.nanmean(a, axis=axis, dtype=dtype)`
> But it seems like this is a known issue and could be suppressed, but I'm not overly concerned about addressing it here.

I forgot to suppress this. I looked into it earlier and it's just caused by numpy functions checking for NaNs

michaelavs (approved):

PR #262, opened 2020-09-03
Fix RTD Build Issue

erogluorhan:
This PR mainly fixes the ReadTheDocs building issue in the last 3 days. As a side work, it also removes the file `.rtd_conda_environment.yml` from the root directory because that file is no more of use. 

How Fixed ReadTheDocs building issue:

The issue seemed to be triggered with the commit [f63e923](https://github.com/NCAR/GeoCAT-examples/commit/f63e923071ed42aa025b8256c352f2555529309d).

The changed file in that commit was `Plots/CSV/README.rst`, and the issue was comming from the following lines:

```
Write CSV
==========
```

where RTD builds turns out to be requiring a number of '=' (i.e. equal signs) that is exactly equal to the heading string put into preceding line. I would not personally expect that to result in a RTD building error though.


anissa111 (approved):

michaelavs (approved):

.github/workflows/build.yml
```diff
@@ -3,7 +3,7 @@ name: Build documentation pages
 on:
   push:
     branches:
-      - '*'
+    - '*'
```
pilotchute, line 6: indentation change not needed, but also has no effect, Github has both indentation styles in the guide: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

Plots/CSV/README.rst
```diff
@@ -3,4 +3,4 @@
 .. _CSV-examples-index:
 
 Write CSV
-==========
+=========
```
pilotchute, line 6: Can't believe that this is what broke the build.

PR #263, opened 2020-09-04
Fix workflow not triggering from forked repo issue

erogluorhan:
Adds `pull_request_target` event into `.github/workflows/build.yml` per the following improvement from this [Github anouncement](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) :

"GitHub Actions has always been about more than just continuous integration. Our goal is to enable repository maintainers to automate a variety of workflows and reduce manual effort. In order to protect public repositories for malicious users we run all pull request workflows raised from repository forks with a read-only token and no access to secrets. This makes common workflows like labeling or commenting on pull requests very difficult.

In order to solve this, we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the `pull_request` event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request. This event can be used in combination with the private repository settings as well."

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
     - '*'
+  pull_request_target:
+    branches:
+    - '*'
 
 jobs:
   build-and-run:


erogluorhan:
@pilotchute the reason for workflow not being triggered by PR events from forked repo might be that we removed `pull_request` event from `.github/workflows/build.yml` recently and left only `push` event there in the workflow. `push` event description does sound like its scope is only the base repo, not forked ones.

In addition, do you have an idea whether we can test if the change in this PR can fix the problem without actually merging it to the `master`?

hCraker (approved):

anissa111 (approved):

erogluorhan:
@pilotchute , since several PRs of @hCraker have already been blocked for a long while by the issue this PR would fix, let me merge this PR thanks to two review-approvals. However, depending on its results and your thoughts, I can revert it back since it indeed consists of a tiny change.

pilotchute:
@erogluorhan this usage of `pull_request_target` is correct and is a good improvement. Thanks!

erogluorhan:
> @erogluorhan this usage of `pull_request_target` is correct and is a good improvement. Thanks!

Thanks for your thoughts! I will have a deeper look on these events and will investigate workflows in order (especially aiming for yapf formatter).

PR #267, opened 2020-09-15
Replacing single character colors with their full color name

### LINKED ISSUE #265: Addressing single character color deprecation
`matplotlib` will no longer be supporting using single character color names (i.e. `'r'` instead of `'red'`). This change will take effect in two minor releases from now as stated [here](https://matplotlib.org/3.2.1/api/api_changes.html#colors). This means that we will need to go through the entire gallery and make sure that we switch from `'r'` to `'red'`, `'k'` to `'black'`, `'b'` to `'blue'`, etc.

hCraker:
Closes #265 

This PR replaces colors specified by single characters (i.e. `'k'`, `'b'`, `'r'`) with the full name of the color. This is to avoid issues in the future when Matplotlib no longer supports colors being referenced in this way.


michaelavs (approved):
Thanks for going through and updating these!

anissa111 (approved):
Looks great, thank you!

PR #269, opened 2020-09-21
panel_15

### LINKED ISSUE #266: panel_15
Description: Paneling three plots vertically with the top two sharing a color bar and the bottom most one having its own color bar.

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/panel_15.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/panel_15_lg.png

Category: Panel

hCraker:
Closes #266 
Original NCL Plot: https://www.ncl.ucar.edu/Applications/Images/panel_15_lg.png

Python recreation:
image


Plots/Panels/NCL_panel_15.py
```diff
@@ -0,0 +1,101 @@
...
+# Select an appropriate colormap
+cmap = 'magma'
+
+# Plot data
+C = ax1.contourf(t_1['lon_t'], t_1['lat_t'], t_1.data, levels=np.arange(0, 30, 2), cmap=cmap, extend='both')
+ax2.contourf(t_2['lon_t'], t_2['lat_t'], t_2.data, levels=np.arange(0, 30, 2), cmap=cmap, extend='both')
+C_2 = ax3.contourf(t_6['lon_t'], t_6['lat_t'], t_6.data, levels=np.arange(0, 22, 2), cmap=cmap, extend='both')
+
+# Add colorbars, by specifying two axes for `ax` the colorbar will span both of them
+plt.colorbar(C, ax=[ax1, ax2], ticks=range(0, 30, 2), extendrect=True, extendfrac='auto', shrink=0.85, aspect=13)
```
anissa111, line 98: Could we add ``drawedges=True`` to both of the colorbar functions?

anissa111 (commented):
This looks great, Rose!

I've noted a few small tweaks below, but the main concern I have is the way the plot looks when generating the html. I've pasted what it generates for me below:
image

I thought at first the colorbar spanning the first two plots was forcing a greater spread in the plots, but even if I shrink it, the spread remains the same.  I also tried to manually reduce the subplots spacing, but that did not improve it, either.

Let me know what you think!

michaelavs:
> This looks great, Rose!
> 
> I've noted a few small tweaks below, but the main concern I have is the way the plot looks when generating the html. I've pasted what it generates for me below:
> image
> 
> I thought at first the colorbar spanning the first two plots was forcing a greater spread in the plots, but even if I shrink it, the spread remains the same. I also tried to manually reduce the subplots spacing, but that did not improve it, either.
> 
> Let me know what you think!

I noticed the spacing issue too, and found that increasing the fig size was able to reduce the space between when in my IDE. However, I couldn't test this with the html build because when I run `make html` the webpage doesn't load past the ncl links. I think this is something completely local, so no need to worry about it, but if you try increasing fig size, does this work for reducing space between plots?

hCraker:
This is the output from `make html` now. There is a lot of white space around the edges that I'm working on removing, but for now the subplots are positioned better.
image

anissa111 (approved):
This looks great, thank you for addressing the html generation issues!

michaelavs (approved):
Looks good!

PR #283, opened 2020-10-22
skewt_2_2

### LINKED ISSUE #282: skewt_2_2
Description: Skew-T plot with temperature, dewpoint, and wind data

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/skewt_2.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/skewt_2_2_lg.png

Category: Skew-T

hCraker:
Closes #282 
Original Plot: https://www.ncl.ucar.edu/Applications/Images/skewt_2_2_lg.png
Python Plot:
image


hCraker:
I'm having issues getting the windbarbs to be offset from the plot. It's only an aesthetic difference, but I'd like to make this example consistent with the others on the GeoCAT gallery. @michaelavs do you have any thoughts on how to do this? `xloc` doesn't seem to be working for `skew.plot_barbs()`.
image

michaelavs:
> I'm having issues getting the windbarbs to be offset from the plot. It's only an aesthetic difference, but I'd like to make this example consistent with the others on the GeoCAT gallery. @michaelavs do you have any thoughts on how to do this? `xloc` doesn't seem to be working for `skew.plot_barbs()`.
> image

I've been testing out a couple of theories on this and so far, I haven't found anything glaringly wrong in the code. The short version of what I've tried so far is: testing to see if the wind data is causing the issue (this uses dummy whereas others read from data) by using the dataset from skewt_3_2 just for the wind data, disabling the parcel path plotting calls, changing zorder of the line incase that was causing problems, and trying the main body code from skewt_3_2 (from 'Plot' down). After the last attempt (body change), I think the problem is coming from the data itself. I only say this because if the actual body of the code was the issue, we would've seen a change in outcome when using a code that is currently working. I did also restart/clear out the kernel and running the code with a different body again and had the same result. With that being said, unless you have already found a resolution to this, maybe try changing the file type of the data into a csv file and see if that solves anything.  

If that doesn't help anything, my next guess is something with the axes and maybe the order the data is read in/plotted? It's definitely a weird problem that is going on with this plot, so I'm interested to find out what ends up being the issue.

michaelavs:
> > I'm having issues getting the windbarbs to be offset from the plot. It's only an aesthetic difference, but I'd like to make this example consistent with the others on the GeoCAT gallery. @michaelavs do you have any thoughts on how to do this? `xloc` doesn't seem to be working for `skew.plot_barbs()`.
> > image
> 
> I've been testing out a couple of theories on this and so far, I haven't found anything glaringly wrong in the code. The short version of what I've tried so far is: testing to see if the wind data is causing the issue (this uses dummy whereas others read from data) by using the dataset from skewt_3_2 just for the wind data, disabling the parcel path plotting calls, changing zorder of the line incase that was causing problems, and trying the main body code from skewt_3_2 (from 'Plot' down). After the last attempt (body change), I think the problem is coming from the data itself. I only say this because if the actual body of the code was the issue, we would've seen a change in outcome when using a code that is currently working. I did also restart/clear out the kernel and running the code with a different body again and had the same result. With that being said, unless you have already found a resolution to this, maybe try changing the file type of the data into a csv file and see if that solves anything.
> 
> If that doesn't help anything, my next guess is something with the axes and maybe the order the data is read in/plotted? It's definitely a weird problem that is going on with this plot, so I'm interested to find out what ends up being the issue.

Just a quick update, the issue is coming from the pressure data variable `p`. I've looked at a direct comparison of the code using `skewt_2` data mixed with `skewt_3` data (and slicing data for correct lengths) and the `t` and `tdc` variables invoke no changes to the the plot, but when combined with the pressure data from `skewt_2`, the barbs will not plot off of the diagram. When using the sliced `skewt_3` pressure data, the barbs will be plotted off of the diagram. I'm not sure why this is the case, but I think that somehow the windbarb data is being tied to something with the pressure data which isn't allowing it to be moved via the `xloc` kwarg.

hCraker:
I figured out the issue with the wind barbs. The pressure data extends up to the 20 hPa level. Given that this figure caps out at 100 hPa, I believe `MetPy` doesn't know how to handle wind barbs that extend beyond the upper pressure level of the chart. By reducing the data to just the pressures greater than or equal to 100 hPa, the `xloc` argument works as expected. I also thinned the wind data to more closely resemble the NCL version. 
image

michaelavs:
Looks good so far! My only thought is to add the subtitle from the original NCL plot if possible, but if not, no worries! 

Also just a side thought, do you know what the circles on the wind barb line on the original NCL plot are for? I was looking at them both on this example and my example and have no idea what they could be representing! Not that I want you to add them, I'm just genuinely curious as to what they were for and if you maybe have an idea. My best guess was maybe they were station readings for cloud cover?

hCraker:
> Looks good so far! My only thought is to add the subtitle from the original NCL plot if possible, but if not, no worries!
> 
> Also just a side thought, do you know what the circles on the wind barb line on the original NCL plot are for? I was looking at them both on this example and my example and have no idea what they could be representing! Not that I want you to add them, I'm just genuinely curious as to what they were for and if you maybe have an idea. My best guess was maybe they were station readings for cloud cover?

Regarding the subtitle on the original plot: That subtitle contains calculated values that we don't have functions for. I opened an [issue](https://github.com/NCAR/geocat-comp/issues/89) in `geocat-comp` so that this functionality can be added. Do you think I should add a subtitle now with arbitrary values or should I leave the plot as is and add the subtitle later once we have functions to calculate those values?

Regarding the circles on the wind barb line: I have no idea what those are supposed to represent. They appear in examples [with](https://www.ncl.ucar.edu/Applications/Images/skewt_2_2_lg.png) and [without](https://www.ncl.ucar.edu/Applications/Images/skewt_2_1_lg.png) wind data. They're also on the [default NCL skew-t chart](https://www.ncl.ucar.edu/Applications/Images/skewt_1_1_lg.png) that has no data of any kind on it. I'm assuming the circles are supposed to be markers of some kind, but I'm not sure.

michaelavs:
> > Looks good so far! My only thought is to add the subtitle from the original NCL plot if possible, but if not, no worries!
> > Also just a side thought, do you know what the circles on the wind barb line on the original NCL plot are for? I was looking at them both on this example and my example and have no idea what they could be representing! Not that I want you to add them, I'm just genuinely curious as to what they were for and if you maybe have an idea. My best guess was maybe they were station readings for cloud cover?
> 
> Regarding the subtitle on the original plot: That subtitle contains calculated values that we don't have functions for. I opened an [issue](https://github.com/NCAR/geocat-comp/issues/89) in `geocat-comp` so that this functionality can be added. Do you think I should add a subtitle now with arbitrary values or should I leave the plot as is and add the subtitle later once we have functions to calculate those values?
> 
> Regarding the circles on the wind barb line: I have no idea what those are supposed to represent. They appear in examples [with](https://www.ncl.ucar.edu/Applications/Images/skewt_2_2_lg.png) and [without](https://www.ncl.ucar.edu/Applications/Images/skewt_2_1_lg.png) wind data. They're also on the [default NCL skew-t chart](https://www.ncl.ucar.edu/Applications/Images/skewt_1_1_lg.png) that has no data of any kind on it. I'm assuming the circles are supposed to be markers of some kind, but I'm not sure.

Up to you! I was under the impression those values were manually added, but because that isn't the case, it definitely would seem fair to not include them until the function is added to geocat-comp

hCraker:
> Up to you! I was under the impression those values were manually added, but because that isn't the case, it definitely would seem fair to not include them until the function is added to geocat-comp

I decided to add a note at the beginning of the file explaining why the subtitle is missing and a link to the issue in geocat-comp. I think I'm done making changes so this PR is actually ready for review now

anissa111 (commented):
This looks good!

My only comment is that the font size on the axes tick marks seems really small. In case this is something that's just occurring on my side, this is the image I get through the make html call: 
image

michaelavs (approved):

anissa111 (approved):
This looks good, thank you for making that change!

PR #289, opened 2020-11-10
wrf_interp_3

### LINKED ISSUE #288: wrf_interp_3
Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/wrf_interp_3.ncl
Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/wrf_interp_3_lg.png

michaelavs:
closes #288 
Original NCL Plot: https://www.ncl.ucar.edu/Applications/Images/wrf_interp_3_lg.png


michaelavs:
@hCraker @anissa111 I am having an issue with the y-axis on this projection and wanted to see if either of you maybe have some insight on how to fix it. As it is now, the output is, for lack of better words, not attaching the vertical variables to the y-axis. When I declare the y-axis range of heights, the projection essentially shrinks to 1/1000 the correct scale (or so it seems). I've left the three attempts I've made in the code as three options that I've tried so far for reference. 

-The first option is taken directly from the [wrf-python website ](https://wrf-python.readthedocs.io/en/latest/plot.html) for a vertical cross section. The issue with this route is that while the y-axis is in the same scale, the values are not the same as the NCL version and cannot be edited or pulled from the data set to match.
-The second option I tried to use the convenience functions to maybe trick the code into using the defined variables but allow the tick marks to be the correct numbers. This results in the projection shrinking/not displaying correctly.
-The third option is a variation of the first, and relies on simply defining the ticks and tick marks and going from there.This also results in the incorrect display.

michaelavs:
The minor tick marks on this projection were not formatting correctly. I tried a couple different methods, but all of them were not matching where they were meant to be. If you have any suggestions, let me know, otherwise they may need to be left off of this example

hCraker:
@michaelavs I think the issues you're having might be caused by the fact that the data is in a Lambert Conformal projection versus a rectangular one. I'm not very familiar with WRF datasets so I'm not sure if this is the case.

hCraker:
Okay so I think I have fixed it, but I had to change a bunch of lines. If you set `meta=False` in the `vertcross` function then `qv_cross` will have the meta data about the height included. Then you can use the `xarray` function `qv_cross.plot.contourf` to plot the data and you can set the location of the ticks with `yticks=np.arange(0, 2000, 3000)` and `xticks=x_ticks[::20]`. This will require you to switch the order you do things in the script since you currently define `x_ticks` after you call the plotting function. Then you should be able to use `gvutil_add_major_minor_ticks` to add the minor ticks. Then you can finish this off by using `ax.set_xticklabels(x_lables[::20], rotation-45, fontsize=8)` to make the x tick labels look how you want.

After making my changes, this is what is produced with `make html`. Do you care if I push these changes since I've already got it working? @michaelavs 

image

michaelavs:
> Okay so I think I have fixed it, but I had to change a bunch of lines. If you set `meta=False` in the `vertcross` function then `qv_cross` will have the meta data about the height included. Then you can use the `xarray` function `qv_cross.plot.contourf` to plot the data and you can set the location of the ticks with `yticks=np.arange(0, 2000, 3000)` and `xticks=x_ticks[::20]`. This will require you to switch the order you do things in the script since you currently define `x_ticks` after you call the plotting function. Then you should be able to use `gvutil_add_major_minor_ticks` to add the minor ticks. Then you can finish this off by using `ax.set_xticklabels(x_lables[::20], rotation-45, fontsize=8)` to make the x tick labels look how you want.
> 
> After making my changes, this is what is produced with `make html`. Do you care if I push these changes since I've already got it working? @michaelavs
> 
> image

That would be perfect, thank you so much for looking into and solving this!

hCraker:
> That would be perfect, thank you so much for looking into and solving this!

No problem! I'm happy to help. Let me know if you have questions about anything I changed. One thing I didn't change is the height of the left and right titles. You'll need to raise them up a bit so they don't overlap with the tick marks.

hCraker (approved):
This looks good to me now. The only change you may want to consider is making the font size bigger for the titles and the tick labels in order to match the NCL plot more closely. It's a minor detail so I'm approving this PR anyway.

anissa111 (approved):
This looks good!

michaelavs:
> This looks good!
> 
> EDIT: I just noticed that a file is left over (`wrfout_d03_2012-04-22_23.nc`) in the WRF directory after I ran the make html, is there a way that the file could be removed from the examples directory? If not, it's not a huge deal.

I'm going to push a newer version that should remove the file from the directory after running. I have run `make html` and it looks like there isn't a wrfout file in the directory and the code still runs correctly, so I think this should work overall, but definitely feel free to double check that after the push!

Issue #292, opened 2020-11-17
Github actions update causing build failures


anissa111:
Github deprecated two commands used in ```actions/checkout@v1```,  ```add-path``` and ```set-env``` on November 16.

A GitHub Blog post can be found about the change here: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

erogluorhan:
Yes, I was going to talk about this today, thanks a lot for bringing it together, Anissa! Apparently they took this action due to a security concern, but there seems to be a quick solution by setting some variables in Github actions config file.

PR #293, opened 2020-11-17
Github actions build failures

anissa111:
Addresses #292 

There is a ticket working its way through the action on github now, so I assume the issue will eventually be resolved by itself, but for now, to allow the commands ```add-path``` and ```set-env``` to run as before, I added the ```ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true' ``` tag to the build.yml file as suggested in the error read out. This allows the two deprecated commands to still be used in our build process while the action we are using updates.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Create conda environment
-      uses: goanpeca/setup-miniconda@v2
+      uses: goanpeca/setup-miniconda@v1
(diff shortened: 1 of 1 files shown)


anissa111:
Note that the checks on this PR appear to be failing because one of them is running against master instead of this branch.

michaelavs (approved):
Thanks again for looking into this!

erogluorhan (approved):
Thanks a lot for looking into this!

anissa111:
I'm going to force permissions and merge!

PR #305, opened 2021-01-27
Update actions

pilotchute:


.github/workflows/build.yml
```diff
@@ -1,30 +1,27 @@
...
-        ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
-      uses: goanpeca/setup-miniconda@v1
-      with:
-         activate-environment: geocat-examples
-         environment-file: conda_environment.yml
-    - name: Build sphinx docs
-      shell: bash -l {0}
-      run: |
-        conda info
-        conda list
```
erogluorhan, line 29: Unless there is a specific reason for deletion, I'd suggest keeping `conda info` and `conda list` in the config, maybe in the "conda_env" step, as they have been useful especially when we get errors with dependencies.

.github/workflows/build.yml
```diff
@@ -1,30 +1,27 @@
...
+name: Build Examples
 on:
   push:
-    branches:
-    - '*'
-#  pull_request_target:
-#    branches:
-#    - '*'
-
+    branches-ignore:
```
erogluorhan, line 4: I get why you ignored `main` here, but there were several occasions where we broke the `main` in the past. Are we 100% secure not to break `main` again with this new config flow? One more thing is that our Github Actions Status Badge in `README.md` is working on `main`, so I am not sure it will continue working after merging this change.

erogluorhan (commented):
Looks great overall! Please address a few inline comments though.

anissa111 (changes requested):
I might be reading this wrong, but these changes would push yapf changes upon every push to every branch, correct? Last time this was set up, it was kind of disruptive and often just pushed empty commits. Could we configure it to only push upon merge to main? I don't think every single push to the repo needs to trigger formatting, though I agree that formatting all the code in main is ideal.

Also, while we're at it, could we add a periodically triggered build to main? As mentioned by Orhan, sometimes dependencies have broken our main without any obvious indicator.

erogluorhan:
> I might be reading this wrong, but these changes would push yapf changes upon every push to every branch, correct? Last time this was set up, it was kind of disruptive and often just pushed empty commits. Could we configure it to only push upon merge to main? I don't think every single push to the repo needs to trigger formatting, though I agree that formatting all the code in main is ideal.
> 
> Also, while we're at it, could we add a periodically triggered build to main? As mentioned by Orhan, sometimes dependencies have broken our main without any obvious indicator.

At line 8 in the `yapf.yml` file, you can see that this only runs when the commit message contains a string "'trigger_yapf'". That said, this action is supposed to be triggered manually by the developer. With its current form, it seems to be working nicely, at least only with the initiative of the developer; however, I am still curious if there are other best practices of doing such a formatting among the Github users (I did not spend time searching yet).

I am absolutely backing  your comment on periodically checking the `main` branch to ensure it is not broken at any time, not only for this repo, but also for `geocat-comp`. We have several branch protection rules already defined, but we are still not guaranteed that the main branch never breaks; it is best to see it periodically with an automated test run.

anissa111:
> At line 8 in the `yapf.yml` file, you can see that this only runs when the commit message contains a string "'trigger_yapf'". That said, this action is supposed to be triggered manually by the developer. With its current form, it seems to be working nicely, at least only with the initiative of the developer; however, I am still curious if there are other best practices of doing such a formatting among the Github users (I did not spend time searching yet).
> 
> I am absolutely backing your comment on periodically checking the `main` branch to ensure it is not broken at any time, not only for this repo, but also for `geocat-comp`. We have several branch protection rules already defined, but we are still not guaranteed that the main branch never breaks; it is best to see it periodically with an automated test run.

Oh! I see now, okay! Could the periodic triggering be added to this PR?

.github/workflows/build.yml
pilotchute: We have nightly builds now, so we should continue to ignore main on pushes.

.github/workflows/build.yml
pilotchute: they were only there for testing, I will add them back as needed if there is are any errors, but I want to keep this to only the instructions that are actually needed, so that it self-documents a little better.

michaelavs (approved):
Looks good to me!

erogluorhan (changes requested):
I believe we should revise this PR, being inspired by this [GeoCAT-comp PR](https://github.com/NCAR/geocat-comp/pull/110)

erogluorhan:
@pilotchute I will work on integrating the recent updates on GeoCAT-comp actions to here (which the team seemed to like such as pre-commit). If you don't mind I can take over this branch and PR, or work on a new branch. Let me know.

pilotchute:
@erogluorhan given the changes on geocat-comp, I would suggest making a new branch to implement those changes, rather than use this branch.

erogluorhan:
> @erogluorhan given the changes on geocat-comp, I would suggest making a new branch to implement those changes, rather than use this branch.

Sounds good, thank you!

PR #306, opened 2021-02-08
NCL_panel_9.py

### LINKED ISSUE #284: panel_9
Description: Paneling a polar map projection and an XY plot in the same figure

NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/panel_9.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/panel_9_lg.png

Category: Panel

hCraker:
Closes #284 
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/panel_9_lg.png

image


hCraker:
The original NCL script uses [wgt_runave_Wrap](https://www.ncl.ucar.edu/Document/Functions/Contributed/wgt_runave_Wrap.shtml) to calculate a running average for the data in the XY plot and then plots the resulting line. Is there an equivalent for this in Python yet? I have been unable to find something that is sufficient. @michaelavs @anissa111 Thanks!

hCraker:
I plotted the rolling average using `xarray.DataArray.rolling()` and it is similar to what is in the NCL plot, but the Python version is still jagged. The NCL version is also a weighted rolling average which hasn't been implemented in xarray yet. I'm going to continue to look for an improved way of doing this, but let me know if any of you have suggestions.

hCraker:
For future reference, the weighted rolling mean of xarray data can be calculated by first making a `DataArray` object with the weights. Then, call `xarray.rolling()` on the data in question. Using `.construct('window')` on the rolling object and then doing the dot product between it and the array of weights returns the weighted average. Note that the weights have to sum to 1 for this to work properly.

The edit to this [stack overflow post](https://stackoverflow.com/a/48512802) explains it better.

hCraker:
Current state of the plot using `make html`
image

hCraker:
To properly fill the xy plot between y=0 and the data, I had to interpolate points and make that into a new array. Is there a better way to do this? I want to make sure this isn't terribly inefficient. I do this starting at line 118.

michaelavs:
I think everything looks good overall. I looked into your question about the xy plot and I couldn't find a better way to do the plotting, but I also don't have much of a problem with how the code is now as long as it doesn't take a super long time to build the html file (which it seems it isn't). The only thing I see needing changed is removing native_1 from the PR/branch because I think it's just sticking from another branch currently. After that, I should be ok with approving this PR!

hCraker:
Thanks for catching that @michaelavs. I accidentally did some work in the wrong branch. This PR should be good to go now.

Plots/Panels/NCL_panel_9.py
```diff
@@ -0,0 +1,144 @@
...
+                    vmax=3.5,
+                    levels=19,
+                    colors='black',
+                    linewidths=0.25,
+                    linestyles='solid')
+
+# Add mean temperature over time data to XY plot
+line = ax2.plot(xyarr.time, xyarr, linewidth=0.25, color='black')
+
+# Retreive data points and interpolate to make the fillcolor more complete
```
anissa111, line 117:
```suggestion
# Retrieve data points and interpolate to make the fillcolor more complete
```

Plots/Panels/NCL_panel_9.py
```diff
@@ -0,0 +1,144 @@
...
+x, y = line[0].get_data()
+num_pts = 400  # number of points to interpolate
+x_interp = np.linspace(x[0], x[-1], num_pts)
+y_interp = np.interp(x_interp, x, y)
+
+# Fill above and below the zero line
+ax2.fill_between(x_interp, y_interp, where=y_interp > 0, color='red')
+ax2.fill_between(x_interp, y_interp, where=y_interp < 0, color='blue')
```
anissa111, line 125:
```suggestion
ax2.fill_between(x, y, where=y > 0, color='red', interpolate=True)
ax2.fill_between(x, y, where=y < 0, color='blue', interpolate=True)
```

Plots/Panels/NCL_panel_9.py
```diff
@@ -0,0 +1,144 @@
...
+                    linestyles='solid')
+
+# Add mean temperature over time data to XY plot
+line = ax2.plot(xyarr.time, xyarr, linewidth=0.25, color='black')
+
+# Retreive data points and interpolate to make the fillcolor more complete
+x, y = line[0].get_data()
+num_pts = 400  # number of points to interpolate
+x_interp = np.linspace(x[0], x[-1], num_pts)
+y_interp = np.interp(x_interp, x, y)
```
anissa111, line 121:
```suggestion
```

anissa111 (changes requested):
> To properly fill the xy plot between y=0 and the data, I had to interpolate points and make that into a new array. Is there a better way to do this? I want to make sure this isn't terribly inefficient. I do this starting at line 118.

There is actually a better way of doing this, though it essentially just does what you did on the back end. There's an option on the ```fill_between(...)``` call that will do this for you!

I've included the changes as suggestions below along with a typo I found. Other than that, this looks good to go, thank you!

hCraker:
> There is actually a better way of doing this, though it essentially just does what you did on the back end. There's an option on the `fill_between(...)` call that will do this for you!
> 
> I've included the changes as suggestions below along with a typo I found. Other than that, this looks good to go, thank you!

Thank you @anissa111! I had no idea I could use that kwarg. This should be good to go once all of the checks complete.

anissa111 (approved):
This looks good, thank you for addressing those changes so quickly!

michaelavs (approved):
Looks good to me!

PR #318, opened 2021-03-26
Update Actions and Repo

erogluorhan:
WORK IN PROGRESS FOR THE TIME!

This PR is aimed at making this repo look as much as needed/applicable like GeoCAT-comp repo with respect to actions, folder structure, etc. Therefore it makes the following changes:

- Moves `readthedocs` related files into newly created `/docs` folder as in `geocat-comp`.


pilotchute:
Was preparing to give my review, when I saw that some automated function had modified the fortran files. While I do not expect that any of the fortran files have been broken, I do not like that they have been touched at all. They need to remain untouched, in order to preserve our confidence that they do in fact provide the exact desired behavior from NCL.

erogluorhan:
> Was preparing to give my review, when I saw that some automated function had modified the fortran files. While I do not expect that any of the fortran files have been broken, I do not like that they have been touched at all. They need to remain untouched, in order to preserve our confidence that they do in fact provide the exact desired behavior from NCL.

It does not apply to here (Though it is a great catch for GeoCAT-f2py), and I believe this PR is ready for review unless you have another point. 

Fyi: Since this is the first run of our `pre-commit hooks` on all the files, there are unfortunately many files changed, but most of those changes are insignificant.

anissa111 (approved):
This looks good. Thank you for making these changes. I think we should consider moving the RTD configuration files in geocat-examples into a docs file like in this PR, too.

erogluorhan:
> I think we should consider moving the RTD configuration files in geocat-examples into a docs file like in this PR, too.

Did you meant some other repository other than GeoCAT-examples? That is because this PR is already going to do that once it is merged in GeoCAT-examples. Am I missing something?

erogluorhan:
@michaelavs @pilotchute If you are not planning to give this PR a review, can I merge this?

erogluorhan:
> Apologies for the late review, I think I had mixed this PR with the one in f2py! I am ok with these changes and merging when you are ready!

No worries at all! I had already mentioned I did not intend to get the actual review of you all but aimed at having you all know what I changed overall about our CI actions and so.

PR #319, opened 2021-04-08
Added citing info

erogluorhan:
Adds "how to cite" info into README.md and RTD docs.

Also adds `dependabot` as it has been working smoothly with geocat-comp.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      # Check for updates once a week
+      interval: 'weekly'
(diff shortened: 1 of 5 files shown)


michaelavs (approved):
Looks good to me, thanks for adding this!

anissa111 (commented):
This is not the point of this PR, but are we not allowing for different DOIs to cite different versions of our packages? The DOI I see listed in this PR links to the geocat homepage instead of any sort of repository. Is there going to be a way to cite the project as a whole vs different releases?

erogluorhan:
> This is not the point of this PR, but are we not allowing for different DOIs to cite different versions of our packages? The DOI I see listed in this PR links to the geocat homepage instead of any sort of repository. Is there going to be a way to cite the project as a whole vs different releases?

Great question! For now, we hope that a single overarching DOI would work for all of our software assets and our reporting purposes. If you check the citation text of GeoCAT-comp, you will see that we are recommending to cite a specific version number in the text. But for GeoCAT-examples, there is no version we keep track of, so we are only recommending to cite that it is "GeoCAT-examples" package.

If in the future, we need separate DOIs for each version of each software tool, we can do that.

anissa111 (approved):
Looks good, thank you!

PR #321, opened 2021-04-12
Fixed spelling error in ReadTheDocs description header paragraph

lkailynncar:
The word "functionality" was misspelled in the GeoCAT documentation file that generates the GeoCAT example homepage on ReadTheDocs.  This has been fixed.


michaelavs (approved):
Thanks for catching this!

michaelavs:
Before merging this branch, I wanted to ask: do you want me to keep this open and you can push each change you find/make, or would you prefer I merge it and you can create new PRs for each change? Either route is fine, but I wanted to make sure I didn't merge anything too early for your preference.

michaelavs:
@anissa111 would you mind looking at the changes proposed to animate_1, specifically lines 95 and 99? We just want to make sure the updates to the comments are correct and I believe that you may have a better understanding of the functions used than I do.

Plots/Animations/NCL_animate_1.py
```diff
@@ -92,8 +92,8 @@ def animate(i):
         ylabel="")
 
 
-# runs the animation initiated with the frame from init and progressed with the animate function
+# Iterates over the animation initiated with the frame int and progressed with the animate function
```
anissa111, line 95: "Iterates over" is better wording, but the phrase "from init" refers to the initialization of the plot, not an int. Its shortened as init because that is the commonly used shortening in this context from the keyword argument in the FuncAnimation call that is called init. But as this example actually got rid of the explicit init, I can see the confusion. May I suggest the following rephrase:

```suggestion
# Iterates over the initiated animation initiated when progressed with the animate function
```

Plots/Animations/NCL_animate_1.py
```diff
@@ -43,7 +43,7 @@
 ax.coastlines(linewidths=0.5)
 ax.set_extent([-180, 180, -90, 90], ccrs.PlateCarree())
 
-# Use geocat.viz.util convenience function to set axes limits & tick values
+# Use geocat.viz.util convenience function to set axis limits & tick values
```
anissa111, line 46: Axes is actually the correct version here, as the util function corrects multiple axes.

```suggestion
# Use geocat.viz.util convenience function to set axes limits & tick values
```

lkailynncar: Yes, but the phrase "axes limits" is incorrect grammatically.  Maybe change it to "tick values, and limits for axes"?

michaelavs: So I think both of you are right to some extent. Grammatically, yes it doesn't make sense, however, in matplotlib axes is an object, not the plural of axis as it would be used in math. That being said, I think the correct way to fix this sentence is to add "the" before axes which would make the sentence "Use geocat.viz.util convenience function to set the axes limits & tick values". I think this helps to show that axes is a singular object instead of the inferred plural of axis.

anissa111: At the risk of sounding pedantic, I still think axes is grammatically correct? I'm totally open to somebody explaining it differently, but the convenience function is setting a lower and upper limit for both the x and y axis, so with multiple axes and multiple limits being set per axis, isn't "axes limits" grammatically fine?

anissa111: Additionally, the util function itself is called ```set_axes_limits_and_ticks```, which did not originate with me.

michaelavs: I see what you're saying Anissa. I wasn't reading the sentence with that in mind, I was reading it in terms of when we set the ax variable for the matplotlib axes. But what you pointed out also makes sense here, the function is being used for multiple axes so it would be fine grammatically as well. I think the confusion really does come down to the word choice of matplotlib for making axes mean the whole plot that also contains an x and y axis that are unrelated to the 'axes' name/object.

Plots/Bar/NCL_bar_11.py
```diff
@@ -40,7 +39,7 @@
 panel = 0
 for row in range(0, 2):
     for col in range(0, 2):
-        # Use geocat.viz.util convenience function to set axes parameters
+        # Use geocat.viz.util convenience function to set axis parameters
```
michaelavs, line 42:
```suggestion
        # Use geocat.viz.util convenience function to set axes parameters
```
This function is setting the parameters for both axes at once, so 'axes' would be correct here

Plots/Bar/NCL_bar_7.py
```diff
@@ -93,7 +93,7 @@
 # Add the legend
 plt.legend(bars_reversed, labels_reversed)
 
-# Use geocat.viz.util convenience function to set axes limits & tick values without calling several matplotlib functions
+# Use geocat.viz.util convenience function to set axis limits & tick values without calling several matplotlib functions
```
michaelavs, line 96:
```suggestion
# Use geocat.viz.util convenience function to set the axes limits & tick values without calling several matplotlib functions
```
This will also be axes because it is applying to both axes at once

PR #324, opened 2021-04-24
Downgrade cartopy to 0.18.0 to fix CI errors

erogluorhan:
After `cartopy`'s version 0.19.0 release, shape file errors occurred in two scripts, which are not occurring in local: 

- `NCL_polyg_2`: KeyError: "There is no item named 'ne_110m_admin_1_states_provinces_lakes_shp.cpg' in the archive" 
- `NCL_WRF_zoom_1_2`: KeyError: "There is no item named 'ne_50m_admin_1_states_provinces_shp.cpg' in the archive"

Above errors can be found in scheduled CI actions, e.g. [see this one](https://github.com/NCAR/GeoCAT-examples/runs/2442188579?check_suite_focus=true).

This PR downgrades `cartopy` to 0.18.0, and it seems to have worked. 

It also fixes CI link in Badge, which was accidentally linking to `geocat-comp`'s badge before.

### DIFF
--- a/Plots/Polygons/NCL_polyg_2.py
+++ b/Plots/Polygons/NCL_polyg_2.py
@@ -85,7 +85,7 @@ ax = plt.axes([.05, -.05, .9, 1],
 ax.set_extent([-119, -74, 18, 50], ccrs.Geodetic())
 
 # Set shape name of map (which depicts the United States)
-shapename = 'admin_1_states_provinces_lakes_shp'
+shapename = 'admin_1_states_provinces_lakes'
 states_shp = shpreader.natural_earth(resolution='110m',
                                      category='cultural',
                                      name=shapename)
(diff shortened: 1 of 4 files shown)


pilotchute (approved):
While I don't like pinning versions, and we will want to investigate unpinning after about 6 months, I approve this for now.

erogluorhan:
> While I don't like pinning versions, and we will want to investigate unpinning after about 6 months, I approve this for now.

Sounds good! I also opened an issue in cartopy repo, so if we can get a solution there, we could unpin this sooner.

erogluorhan:
@pilotchute You can see from the files changed that I now figured out the actual issue. So, I reverted cartopy version pinning. We are good with the newest 0.19.0 version. I will merge this version

Issue #348, opened 2021-05-21
Error in RTD installation page


anissa111:
[Installation instructions on RTD](https://geocat-examples.readthedocs.io/en/latest/install.html) specify an old version of geocat-viz

PR #350, opened 2021-05-24
Update RTD installation page

anissa111:
Addresses #348 

This PR updates the RTD installation page by replacing it with the correct installation instructions in INSTALLATION.md.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -14,8 +14,10 @@ environment and then building file of which is provided in this repo as follows:
 
 From the root directory of the cloned geocat-examples repository, run the following commands:
 
-    conda env create -f conda_environment.yml -n geocat-examples
-    conda activate geocat-examples
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


erogluorhan (commented):
Thanks for fixing this! A quick question:

Neither this document nor INSTALLATION.md mentions `make html` (it is only mentioned in CONTRBITUING.md). Do you think we should include that information in these INSTALLATION documents as well, e.g. saying something like "if you want to generate all the plots at once, ..."

anissa111:
> Thanks for fixing this! A quick question:
> 
> Neither this document nor INSTALLATION.md mentions `make html` (it is only mentioned in CONTRBITUING.md). Do you think we should include that information in these INSTALLATION documents as well, e.g. saying something like "if you want to generate all the plots at once, ..."?

Given that it takes a fair amount of time to run and that you kind of have to go looking for where they were all generated, I don't think we need to include any of the doc generation stuff.

Issue #329, opened 2021-05-12
maponly_2


anissa111:
Description: Plot of only maps for several subregions

See previously implemented [maponly_1](https://geocat-examples.readthedocs.io/en/latest/gallery/MapProjections/NCL_maponly_1.html#sphx-glr-gallery-mapprojections-ncl-maponly-1-py)

NCL Page: https://www.ncl.ucar.edu/Applications/maponly.shtml
NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/maponly_2.ncl
NCL Plots: https://www.ncl.ucar.edu/Applications/Images/maponly_2_1_lg.png
https://www.ncl.ucar.edu/Applications/Images/maponly_2_2_lg.png![image](https://user-images.githubusercontent.com/38434768/118023111-60d49c00-b31a-11eb-803f-400638c3b4c8.png)

![image](https://user-images.githubusercontent.com/38434768/118023009-426ea080-b31a-11eb-9bd8-a64d03c62cb3.png)

![image](https://user-images.githubusercontent.com/38434768/118023046-4c909f00-b31a-11eb-9616-731c8e19d266.png)

![image](https://user-images.githubusercontent.com/38434768/118023126-66ca7d00-b31a-11eb-939c-d58a93495ed3.png)

PR #352, opened 2021-05-25
maponly_2

erinlincoln:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/maponly_2.ncl
- Original NCL plots: https://www.ncl.ucar.edu/Applications/Images/maponly_2_1_lg.png
                          https://www.ncl.ucar.edu/Applications/Images/maponly_2_2_lg.png
                          https://www.ncl.ucar.edu/Applications/Images/maponly_2_3_lg.png  
- Issue #329 
image

### DIFF
--- a/Plots/MapProjections/NCL_maponly_2.py
+++ b/Plots/MapProjections/NCL_maponly_2.py
@@ -29,7 +29,7 @@ from geocat.viz import util as gvutil
 
 def map_plot(scale, long_min, long_max, lat_min, lat_max, long_labels,
              lat_labels):
-    """Plots a map-only figure with continent borders, country borderes, and
+    """Plots a map-only figure with continent borders, country borders, and
     lakes at a certain longitude and latitude.
 
     Parameters
--- a/Plots/MapProjections/NCL_maponly_2.py
+++ b/Plots/MapProjections/NCL_maponly_2.py
@@ -28,9 +28,6 @@ from geocat.viz import util as gvutil
 ###############################################################################
 # Plot:
 
-# Generate figure with (width, height) in inches
-fig = plt.figure(figsize=(12, 24))
-
 
 # Define a utility function that takes in the scale of the continent borders,
 # the number of rows and columns and position within the subplot, the latitude
@@ -41,7 +38,9 @@ def map_plot(scale, row, col, pos, long_min, long_max, lat_min, lat_max,
 
     # Define the projection and generate axes using Cartopy
     projection = ccrs.PlateCarree()
-    ax = fig.add_subplot(row, col, pos, projection=projection)
+    # Generate figure with (width, height) in inches
+    fig = plt.figure(figsize=(12, 8))
+    ax = plt.axes(projection=projection)
 
     # Add in country borders, continent borders, and lakes
     ax.add_feature(
@@ -77,6 +76,7 @@ def map_plot(scale, row, col, pos, long_min, long_max, lat_min, lat_max,
                                      yticks=np.linspace(lat_min, lat_max, 4),
                                      xticklabels=long_labels,
                                      yticklabels=lat_labels)
+    plt.show()
 
 
 # Latitude and longitude labels for the top subplot
--- a/Plots/MapProjections/NCL_maponly_2.py
+++ b/Plots/MapProjections/NCL_maponly_2.py
@@ -1,6 +1,6 @@
 """
 NCL_maponly_2.py
-===========
+================
 This script illustrates the following concepts:
     - Outlining continents
     - Turning on map country boundaries
--- a/Plots/MapProjections/NCL_maponly_2.py
+++ b/Plots/MapProjections/NCL_maponly_2.py
@@ -10,9 +10,7 @@ This script illustrates the following concepts:
 
 See following URLs to see the reproduced NCL plot & script:
     - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/maponly_2.ncl
-    - Original NCL plots: https://www.ncl.ucar.edu/Applications/Images/maponly_2_1_lg.png
-                          https://www.ncl.ucar.edu/Applications/Images/maponly_2_2_lg.png
-                          https://www.ncl.ucar.edu/Applications/Images/maponly_2_3_lg.png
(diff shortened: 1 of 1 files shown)


erinlincoln:
Quick note - here are a few lakes in the second and third subplots that I can't get to show without showing many more lakes. I can't tell if this is a quirk between the different scales between NCL and Cartopy or if there is a fix.

anissa111:
> Quick note - here are a few lakes in the second and third subplots that I can't get to show without showing many more lakes. I can't tell if this is a quirk between the different scales between NCL and Cartopy or if there is a fix.

I believe this is a difference in the shapefiles that NCL uses vs what cartopy uses. There are several examples where I remember not being able to recreate the maps exactly to what NCL used. I would suggest using the option to show more lakes on the third plot. @michaelavs, you're much more familiar with cartopy than I am, what do you think?

michaelavs:
I agree with Anissa. When I was making projections, I would notice that Cartopy had more lakes/lake features than what NCL would show. I think Anissa is correct that it's nothing more than a difference between the shapefiles each software uses for the lake features and not necessarily something to worry about. A similar thing you may run across is some slight differences in the data when it is plotted in Python, but this is also just a difference in how the two softwares resolve the data for plotting and not something to worry about (unless the difference is sizable, then definitely bring it to our attention).

erinlincoln:
I changed the lake resolution to be higher for the final graph. There are many more lakes now, so let me know if you think there's too many and I can make the resolution higher!

Plots/MapProjections/NCL_maponly_2.py
```diff
@@ -0,0 +1,96 @@
+"""
+NCL_maponly_2.py
+===========
```
anissa111, line 3:
```suggestion
NCL_maponly_2.py
================
```

anissa111 (changes requested):
I think the way you handled the lakes here is appropriate. My only concern is that the original plots are three separate plots and not stacked subplots as you've created here. I suggest making three individual plots for this example.

erinlincoln:
I separated the subplots into three different plots:
maponly_2_fig_3
maponly_2_fig_2
maponly_2_fig_1

anissa111:
This is a small point, but you could probably reduce the individual figsize so that there isn't so much white space around each plot

erinlincoln:
Here are the updated graphs with a tight layout!
image
image
image

erinlincoln:
I changed the function definition to be within the function like the example and make the links in a line. I am going to see how it looks once built but then I think everything should be good!

erinlincoln:
The links display correctly on the build!

Plots/MapProjections/NCL_maponly_2.py
```diff
@@ -0,0 +1,110 @@
...
+        lat_max : int
+            maximum latitude for plotting
+        long_labels : list
+            list of tick labels for x axis
+        lat_labels : list
+            ist of tick labels for y axis
+
+    Plots a map-only figure with continent borders, country borderes, and lakes
+    at a certain longitude and latitude.
+    '''
```
anissa111, line 52: Make sure you're using double quotation marks for docstrings.

Also @michaelavs & @erogluorhan, do we have a standardized way of formatting docstrings for helper functions in geocat-examples?

anissa111 (commented):
You're so close! I like the way these plots look as opposed to the ones with more whitespace around them.

Plots/MapProjections/NCL_maponly_2.py
erogluorhan: I full agree with @michaelavs .

Plots/MapProjections/NCL_maponly_2.py
```diff
@@ -0,0 +1,121 @@
...
+plt1_lat_labels = ["", "30S", "0", "30N"]
+
+# Latitude and longitude labels for the middle and bottom subplots
+plt23_long_labels = ["0", "30E", "60E", ""]
+plt23_lat_labels = ["", "30N", "60N", "90N"]
+
+# Plot all three subplots
+map_plot('110m', 90, 160, -50, 30, plt1_long_labels, plt1_lat_labels)
+map_plot('110m', 0, 70, 10, 90, plt23_long_labels, plt23_lat_labels)
+map_plot('10m', 0, 70, 10, 90, plt23_long_labels, plt23_lat_labels)
```
michaelavs, line 121: @anissa111 I noticed that the minor tick marks on the projections over Africa are weird between 10N and 30N. I was try to re-select the data to make the minor per major line up on the same scale as 30N-60N and 60N-90N. Right now, the minor ticks should be at every 10 degrees of latitude, but below 30N they are at ~6.67 degrees each. Do you have any ideas on how to maybe ignore the 3 min-per-major for this location?

anissa111: @michaelavs I don't think I'm understanding the difference you're seeing. Can you show me at/after the intern weekly meeting tomorrow?

michaelavs: > @michaelavs I don't think I'm understanding the difference you're seeing. Can you show me at/after the intern weekly meeting tomorrow?

Absolutely, because I knew there was a very good chance my comment made no sense! lol

Plots/MapProjections/NCL_maponly_2.py
```diff
@@ -0,0 +1,121 @@
...
+                   facecolor="none")
+
+    # Set extent to show particular area of the map
+    ax.set_extent((long_min, long_max, lat_min, lat_max), crs=projection)
+
+    # Use geocat.viz.util convenience function to add minor and major tick lines
+    gvutil.add_major_minor_ticks(ax,
+                                 x_minor_per_major=3,
+                                 y_minor_per_major=3,
+                                 labelsize=10)
```
michaelavs, line 93:
```suggestion
                                 labelsize=15)
```
Increasing the label size to be closer to NCL but also to make the projections a bit easier to read. After this change I should be ready to approve this PR!

michaelavs (approved):

anissa111 (approved):
Looks good, thank you!

Issue #346, opened 2021-05-12
box_2


anissa111:
Description: boxplot with altered thickness and colors

Related: [GeoCAT-examples boxplots](https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#boxplots)

NCL Page: https://www.ncl.ucar.edu/Applications/box.shtml
NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/box_2.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/box_2_lg.png

![image](https://user-images.githubusercontent.com/38434768/118046684-e36b5480-b336-11eb-92f7-a25b50acb8cf.png)

Jiaqi-beep:
Is the NCL plot link correct? It links to a text plot instead of box plot

anissa111:
Oops! You're right it's a copy paste error on my part. I'll fix it. You should still be able to find the right plot from the NCL Page link

anissa111:
It's fixed now!

PR #357, opened 2021-06-08
Box 2

Jiaqi-beep:
Description: boxplot with altered thickness and colors

- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/box_2.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/box_2_lg.png
- Issue #346

box2


Plots/Boxplots/NCL_box_2.py
```diff
@@ -0,0 +1,95 @@
+"""
+NCL_box_1.py
+===============
```
anissa111, line 3:
```suggestion
NCL_box_1.py
============
```

anissa111 (commented):
This looks good so far! I think what you mentioned as looking a little funny is that the original script modifies the width of the boxplot, which I don't think you're recreating. Note that in the original, both the box itself and the top and bottom range markers are reduced in width.

Also, when you make a new branch, make sure you're branching off of main so that extra files don't get pulled in! It looks like you've got your xy_6 plot attached to this PR, too.

Plots/Boxplots/NCL_box_2.py
Jiaqi-beep: I see. Does that mean I should be on the main branch when I do git checkout -b newbranch?

anissa111: Yup exactly that!

Jiaqi-beep: Got it. Should I try and delete those files from this branch? Also by length of the range markers, do you mean the length of the whisker? (I'm assuming using the whis argument)

anissa111: `whis` is not actually the argument you want. I think you'll actually want to modify `cap`

Plots/Boxplots/NCL_box_2.py
```diff
@@ -0,0 +1,95 @@
+"""
+NCL_box_1.py
```
anissa111, line 2:
```suggestion
NCL_box_2.py
```

michaelavs (approved):
Looks good to me!

anissa111 (approved):
Looks good to me, too!

Issue #330, opened 2021-05-12
xy_6


anissa111:
Description: Default NCL XY plot with explicit axes labeling

See GeoCAT-examples [existing XY Plots](https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#xy-plots)

NCL Page: https://www.ncl.ucar.edu/Applications/xy.shtml
NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/xy_6.ncl
NCL Plot: https://www.ncl.ucar.edu/Applications/Images/xy_6_lg.png

![image](https://user-images.githubusercontent.com/38434768/118026022-9dee5d80-b31d-11eb-99f3-526ab7287fa0.png)

PR #358, opened 2021-06-08
Xy 6

Jiaqi-beep:
Original NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/xy_6.ncl
Original NCL Plot: https://www.ncl.ucar.edu/Applications/Images/xy_6_lg.png
Issue: #330 

xy6

### DIFF
--- a/Plots/XY/NCL_xy_6.py
+++ b/Plots/XY/NCL_xy_6.py
@@ -70,11 +70,17 @@ gvutil.set_axes_limits_and_ticks(ax,
                                  xlim=(1951, 1991),
                                  ylim=(-1.5, 0.6),
                                  yticks=np.arange(-1.2, .6, .4),
-                                 xticks=np.linspace(1951, 1991, 11),
+                                 xticks=np.array([
+                                     1951, 1954, 1957, 1961, 1965, 1968.5, 1972,
+                                     1977, 1982, 1987, 1991
+                                 ]),
                                  xticklabels=[
                                      '1951', '', '1957', '', '1965', '', '1972',
                                      '', '1982', '', '1991'
                                  ])
+
+# Turn off minor ticks on y axis
+ax.tick_params(axis='x', which='minor', bottom=False, top=False)
 
 # Use geocat.viz.util convenience function to set titles and labels without calling several matplotlib functions
 gvutil.set_titles_and_labels(ax,
--- a/Plots/XY/NCL_xy_6.py
+++ b/Plots/XY/NCL_xy_6.py
@@ -55,7 +55,7 @@ plt.figure(figsize=(7, 6.5))
 ax = plt.gca()
 
 # Plot
-plt.plot(warm_yrs, y)
+plt.plot(warm_yrs, y, color='grey', linewidth=1)
 
 # Use geocat.viz.util convenience function to add minor and major tick lines
 gvutil.add_major_minor_ticks(ax,
--- a/Plots/XY/NCL_xy_6.py
+++ b/Plots/XY/NCL_xy_6.py
@@ -3,7 +3,6 @@ NCL_xy_6.py
 ===========
 This script illustrates the following concepts:
    - Changing the labels and tickmarks on the X axis in an XY plot
-
 See following URLs to see the reproduced NCL plot & script:
     - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/xy_6.ncl
     - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/xy_6_lg.png
@@ -71,7 +70,7 @@ gvutil.set_axes_limits_and_ticks(ax,
                                  yticks=np.arange(-1.2, .6, .4),
                                  xticks=np.array([
                                      1951, 1954, 1957, 1961, 1965, 1968.5, 1972,
-                                     1977, 1982, 1987, 1991
+                                     1977, 1982, 1986.5, 1991
                                  ]),
                                  xticklabels=[
                                      '1951', '', '1957', '', '1965', '', '1972',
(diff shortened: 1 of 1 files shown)


Jiaqi-beep:
The plot isn't exactly the same. I think it's because of how the xticks are assigned. I'm not exactly sure how the original plot creates equal spacing on the tickmarks on x-axis while the actual values(years) are not equally spaced apart.

michaelavs:
> The plot isn't exactly the same. I think it's because of how the xticks are assigned. I'm not exactly sure how the original plot creates equal spacing on the tickmarks on x-axis while the actual values(years) are not equally spaced apart.

So I think I have made a (very) hacky work around for this issue with ```ax.xaxis.set_minor_locator``` and also defining the xticks in line 72 and removing the null ('') labels from xticklabels list. Like I said, it's hacky, so @anissa111 if you know of a better route than this, definitely let me know but, either way, here is what I have inserted on line 77:
```
ax.xaxis.set_minor_locator(FixedLocator([1954, 1961, 1968.5, 1977, 1987]))
```
this will put exactly one tick in middle point of each major tick at the specified value. You can use decimals, so it can be precise enough. But what I don't love about this version is that it specified by hand and I have a feeling it may not be the true blue best approach. 

Let me know what you think!

Jiaqi-beep:
> > The plot isn't exactly the same. I think it's because of how the xticks are assigned. I'm not exactly sure how the original plot creates equal spacing on the tickmarks on x-axis while the actual values(years) are not equally spaced apart.
> 
> So I think I have made a (very) hacky work around for this issue with `ax.xaxis.set_minor_locator` and also defining the xticks in line 72 and removing the null ('') labels from xticklabels list. Like I said, it's hacky, so @anissa111 if you know of a better route than this, definitely let me know but, either way, here is what I have inserted on line 77:
> 
> ```
> ax.xaxis.set_minor_locator(FixedLocator([1954, 1961, 1968.5, 1977, 1987]))
> ```
> 
> this will put exactly one tick in middle point of each major tick at the specified value. You can use decimals, so it can be precise enough. But what I don't love about this version is that it specified by hand and I have a feeling it may not be the true blue best approach.
> 
> Let me know what you think!

I tried to follow what you said but I don't seem to be able to

Jiaqi-beep:
Oops sorry, I don't know what I did that closed the PR.

Jiaqi-beep:
> > The plot isn't exactly the same. I think it's because of how the xticks are assigned. I'm not exactly sure how the original plot creates equal spacing on the tickmarks on x-axis while the actual values(years) are not equally spaced apart.
> 
> So I think I have made a (very) hacky work around for this issue with `ax.xaxis.set_minor_locator` and also defining the xticks in line 72 and removing the null ('') labels from xticklabels list. Like I said, it's hacky, so @anissa111 if you know of a better route than this, definitely let me know but, either way, here is what I have inserted on line 77:
> 
> ```
> ax.xaxis.set_minor_locator(FixedLocator([1954, 1961, 1968.5, 1977, 1987]))
> ```
> 
> this will put exactly one tick in middle point of each major tick at the specified value. You can use decimals, so it can be precise enough. But what I don't love about this version is that it specified by hand and I have a feeling it may not be the true blue best approach.
> 
> Let me know what you think!

I tried to edit my code according to this but I still couldn't recreate the original plot. The plot itself doesn't seem to change for me when I change the ticks locations. Do you mind sharing your code from line 72-77? Thank you!

michaelavs:
@Jiaqi-beep 

The code starts on line 66 and goes through 77:
```
# Use geocat.viz.util convenience function to set axes parameters without calling several matplotlib functions
# Set axes limits, tick values, and tick labels
gvutil.set_axes_limits_and_ticks(ax,
                                 xlim=(1951, 1991),
                                 ylim=(-1.5, 0.6),
                                 yticks=np.arange(-1.2, .6, .4),
                                 xticks=[1951, 1957, 1965, 1972, 1982, 1991],
                                 xticklabels=[
                                     '1951', '1957', '1965', '1972',
                                     '1982', '1991'
                                 ])
ax.xaxis.set_minor_locator(FixedLocator([1954, 1961, 1968.5, 1977, 1987]))
```

I will note that the new minor tick marks are pretty faint, but that might be something that can be changed with a kwarg later on.

This is the plot output that I have when running locally:
image

Jiaqi-beep:
@michaelavs Your version totally makes sense to me! That's why I was confused how so it seems that the x-axis tick marks in the original plot seem to space out evenly, which was what I tried to mimic. Am I correct in thinking that we should modify original ncl plot when needed? 
Also, when @anissa111 and I talked briefly the other day, we both noticed that the bend on the plot (between 1957-65) in our version points toward the right, but it points toward the left in the original plot. Are we supposed to fix little differences like these?

anissa111:
I'm not concerned about the slight difference in that one point. Also, I like the idea of moving the years around to make physical sense, even if that's not happening in the original NCL. I have no idea what the original plot was doing with regards to the labels and spacing.

Jiaqi-beep:
updated.
Figure_1

Plots/XY/NCL_xy_6.py
```diff
@@ -0,0 +1,93 @@
...
+    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/xy_6_lg.png
+"""
+
+###############################################################################
+# Import packages:
+
+import numpy as np
+import xarray as xr
+import matplotlib.pyplot as plt
+from matplotlib.ticker import FixedLocator
```
anissa111, line 18:
```suggestion
```

Unused import

anissa111 (commented):
This looks good to me! I There's an unused input, but I'll approve this after that!

michaelavs (approved):
This looks good to me! I changed the location of the tick mark between 1982 and 1991 to 1986.5 instead of 1987 to be the true halfway point, but I didn't think making you do the full change on your end was worth it before approving on my side!

anissa111 (approved):
Looks good to me, too!

Issue #338, opened 2021-05-12
bar_3


anissa111:
Description: basic bar plot with size of bars altered

Related: [GeoCAT-examples bar plots](https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#bar)

NCL Page: https://www.ncl.ucar.edu/Applications/bar.shtml
NCL Script: https://www.ncl.ucar.edu/Applications/Scripts/bar_3.ncl
NCL Plots: https://www.ncl.ucar.edu/Applications/Images/bar_3_lg.png

![image](https://user-images.githubusercontent.com/38434768/118039375-81f2b800-b32d-11eb-8af1-a1b2346bb676.png)

PR #368, opened 2021-06-15
bar_3

erinlincoln:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/bar_3.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/bar_3_lg.png
- Issue #338 
bar_3_fig_1


Plots/Bar/NCL_bar_3.py
```diff
@@ -0,0 +1,88 @@
...
+colors = ['red' if (value > 0) else 'blue' for value in dsoik[::8]]
+
+# Plot bar chart with defined width
+plt.bar(date_frac[::8],
+        dsoik[::8],
+        align='edge',
+        edgecolor='black',
+        color=colors,
+        width=.25,
+        linewidth=.6)
```
anissa111, line 61:
```suggestion
plt.bar(date_frac[::8],
        dsoik[::8],
        color=colors,
        width=.25)
```

You don't actually need the `align='edge'` here, as that was specific to the previous example. I would also suggest removing the black outlines as it doesn't appear to be on the original I think. The lines are very small though, so feel free to disagree with me on that point.

anissa111 (commented):
This looks good so far. I've added some comments below. 

I have a question about the widths of the bars, though. I agree that you've seem to set them appropriately, but when zoomed in on, some seem to be slightly wider?

image

It is only barely noticeable on the overall plot, but it's something that might be interesting to poke around in. @michaelavs have you ever seen anything like this?

erinlincoln:
I also noticed the slightly wider bars and I tried to fix it by setting the dpi but it did not end up helping.

michaelavs:
@anissa111 @erinlincoln 
So I'm not 100% sure if this is why it was happening, but I recalled something similar happening last summer. Looking at bar_2, the plt.bar() call is as follows:
```
plt.bar(date_frac[::8],
        dsoik[::8],
        align='edge',
        edgecolor='black',
        color=colors,
        width=8 / 12,
        linewidth=.6)
```
I did notice that there are still some widths that are larger or smaller in bar_2, but the difference seems to be a little bit less than it was in this version. But based on a little bit of researching on the width argument, it seems like this may be contributed to there being data overlap between each time step? However, I'm not 100% certain if this is the cause. If there isn't a full resolution on this, I'm not too concerned about it because of bar_2 having the same problem and it being approved/merged previously.

erinlincoln:
I played around with the width for a while and nothing made the bars the same size, unfortunately.

Plots/Bar/NCL_bar_3.py
```diff
@@ -0,0 +1,82 @@
+"""
+NCL_bar_3.py
+============
+This script illustrates the following concepts:
+    - Drawing bars instead of curves in an XY plot
+    - Changing the aspect ratio of a bar plot
```
michaelavs, line 6:
```suggestion

```
The aspect ratio isn't changed in the code

michaelavs (commented):
Overall, code is in good shape. Just one thing to take out of "concepts illustrated" and I'm good with approving!

michaelavs (approved):

PR #391, opened 2021-07-13
Polar 8

Jiaqi-beep:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/polar_8.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/polar_8_lg.png

**### Issue: quiver/streamplot is not supported when projection is Geodetic, but vectors are completely wrong when projection is PlateCarree**

(temporary) fix: Added in rotation and scaling of U and V wind components.

image

### DIFF
--- a/Gallery/Contours/NCL_polar_8.py
+++ b/Gallery/Contours/NCL_polar_8.py
@@ -20,7 +20,7 @@ import matplotlib.pyplot as plt
 import matplotlib.ticker as mticker
 
 import geocat.datafiles as gdf
-from geocat.viz import util as gvutil
+import geocat.viz as gv
 
 ###############################################################################
 # Read in data:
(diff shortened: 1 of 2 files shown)


kafitzgerald:
Just noting that it looks like the problem with the `quiver` directions may be related to this issue https://github.com/SciTools/cartopy/issues/1179 over on cartopy.  There's an open PR https://github.com/SciTools/cartopy/pull/1926 as well.

kafitzgerald:
The fix over on cartopy is still an open PR, but it looks like there's been some activity.  

If we end up with extra time, we might be able to work around this, but for the time being I think we should wait to see what happens w/ the cartopy PR.

jukent (approved):

PR #395, opened 2021-07-16
Panel 41

Jiaqi-beep:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/panel_41.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/panel_41_lg.png
image

### DIFF
--- a/Plots/Panels/NCL_panel_41.py
+++ b/Plots/Panels/NCL_panel_41.py
@@ -5,6 +5,8 @@ This script illustrates the following concepts:
    - Paneling six plots on a page
    - Adding a common title to paneled plots using a custom method
    - Adding left, center, and right subtitles to a panel plot
+   - Using a different color scheme to follow `best practices <https://geocat-examples.readthedocs.io/en/latest/gallery/Colors/CB_Temperature.html#sphx-glr-gallery-colors-cb-temperature-py>`_ for visualizations
+
 
 See following URLs to see the reproduced NCL plot & script:
     - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/panel_41.ncl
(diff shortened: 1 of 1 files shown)


anissa111 (commented):
This looks almost perfect! My only comment is that I think the left and right titles should be raised a little bit as to not touch the top tick marks:

image

michaelavs:
This looks good overall!

One thing I noticed was that there were white spots where data is missing on some of the plots. After looking into this, I found that the levels don't encompass all of the data points for this set (even though they match what the NCL version provided). However, if we change the levels to match the data, then the colorbar gets thrown off/doesn't match the NCL version. If anyone knows of a work around for this where the filled contours can have one range of levels and the colorbar has another range of levels, I think that would be the best route for this plot. However, I'm not 100% if there is a clean and easy way to make that happen. So if there isn't, I am totally fine approving this PR with the contour levels as is, but wanted to check on if anyone else maybe knew of a way to fix this that I don't.

Jiaqi-beep:
> This looks good overall!
> 
> One thing I noticed was that there were white spots where data is missing on some of the plots. After looking into this, I found that the levels don't encompass all of the data points for this set (even though they match what the NCL version provided). However, if we change the levels to match the data, then the colorbar gets thrown off/doesn't match the NCL version. If anyone knows of a work around for this where the filled contours can have one range of levels and the colorbar has another range of levels, I think that would be the best route for this plot. However, I'm not 100% if there is a clean and easy way to make that happen. So if there isn't, I am totally fine approving this PR with the contour levels as is, but wanted to check on if anyone else maybe knew of a way to fix this that I don't.

I think creating a mappable allows us to customize color bar, or maybe simply specifying ticks argument for plt.colorbar()? I'll work on this and thank you for spotting the data missing.

anissa111 (approved):
This looks good!

PR #401, opened 2021-07-27
Station 3

Jiaqi-beep:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/station_3.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/station_3_1_lg.png and https://www.ncl.ucar.edu/Applications/Images/station_3_2_lg.png
image
image


Jiaqi-beep:
There are still overlapping texts on the second plot. I'd love to hear some suggestions for this! I believe that this is because the location coordinates I used to calculate the distance is in degrees format, while the projection is Mercator, so the coordinates should be converted into UTM, but I haven't figured out how yet.

Plots/Station/NCL_station_3.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Draw gridlines
+    gl = ax.gridlines(crs=ccrs.PlateCarree(),
+                      draw_labels=True,
+                      dms=False,
+                      x_inline=False,
+                      y_inline=False,
+                      linewidth=1,
+                      color="black",
+                      alpha=0.25)
```
anissa111, line 67: Why are you drawing gridlines only to remove them starting on line 73?

Plots/Station/NCL_station_3.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Set frequency of gridlines in the x and y directions
+    gl.xlocator = mticker.FixedLocator(np.arange(26, 45, 2))
+    gl.ylocator = mticker.FixedLocator(np.arange(36, 43, 1))
```
anissa111, line 71: Similarly, the original NCL plot has tick marks while this recreation does not. It seems like you're creating tick marks here and removing them unnecessarily later on. 

Also, a lot of this helper functions looks like it could be accomplished with -viz functions instead of like this.

Plots/Station/NCL_station_3.py
```diff
@@ -0,0 +1,157 @@
...
+# Create an array of booleans denoting if station would be removed
+remove = np.full(npts, False)
+
+# Currently minimum distance is calcualted through finding distance between two suitable stations
+# In the future we would like to find mindist by finding the width and height of texts in pixel coodinates
+mindist = np.sqrt(np.sum(np.square(location[123] - location[124])))
+
+# Tag station to be removed using array `remove`
+# Loop through every pair of stations and calculate distances between them
+for i in range(npts):
```
anissa111, line 134: I'm not sure this is the best way to do this, but I cannot think of a better solution, @michaelavs any ideas?

Plots/Station/NCL_station_3.py
```diff
@@ -0,0 +1,157 @@
...
+import cartopy.feature as cfeature
+from matplotlib import pyplot as plt
+import matplotlib.ticker as mticker
+
+import geocat.datafiles as gdf
+
+###################################################
+# Generate data:
+
+# Open a netCDF data file using xarray default engine and load the data into xarrays
```
anissa111, line 28: I think you should consider reading this in with xarray instead of pandas, like we do in newer examples.

Plots/Station/NCL_station_3.py
```diff
@@ -0,0 +1,157 @@
...
+import pandas as pd
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+from matplotlib import pyplot as plt
+import matplotlib.ticker as mticker
+
+import geocat.datafiles as gdf
+
+###################################################
+# Generate data:
```
anissa111, line 26:
```suggestion
# Read in data:
```

Not actually generating any data here

anissa111 (changes requested):
Overall this is a good start! I've left some comments to be addressed below. I think you might have based this example on a similar older example that doesn't use some of our more updated techniques.

Plots/Station/NCL_station_3.py
michaelavs: My only other thought would be maybe an if else statement, but I also don't know if that would necessarily be more efficient or not.

Plots/Station/NCL_station_3.py
Jiaqi-beep: @anissa111 I wasn't able to use viz function set_axes_limits_and_ticks because the projection used for this example is Mercator, so the data is in UTM coordinates, but the latitude and longitude datasets are in degrees. So when I tried to directly set ticks to x axis and y axis, the locations of ticks are wrong because they do not correspond to the Mercator projection's coordinate system. I checked with the other examples that use another projection, [proj_2](https://geocat-examples.readthedocs.io/en/latest/gallery/MapProjections/NCL_proj_2.html#sphx-glr-gallery-mapprojections-ncl-proj-2-py), [native_2](https://geocat-examples.readthedocs.io/en/latest/gallery/MapProjections/NCL_native_2.html#sphx-glr-gallery-mapprojections-ncl-native-2-py), they both use gridlines as a workaround. [native_1](https://geocat-examples.readthedocs.io/en/latest/gallery/MapProjections/NCL_native_1.html#sphx-glr-gallery-mapprojections-ncl-native-1-py) is another example that uses another projection and has to add tickmarks and labels with gridlines. I removed the gridlines and keep the labels to make it similar to the original plot, but both native_1 and native_2 keep the gridlines even though the original plots don't. Let me know if this makes sense to you and if there is a better way to do this!

Jiaqi-beep:
@anissa111 Do you have an old example in mind? I didn't actually base this off of an old one, wondering if there is an example that does something similar to model after.

Plots/Station/NCL_station_3.py
anissa111: Oh, okay! I'll look into a better way. You should add this workaround to the "concepts illustrated" section!

Plots/Station/NCL_station_3.py
Jiaqi-beep: I'm wondering, if we have to compare each point against every other point then the runtime has to be O(N*2) or am I missing something?

Plots/Station/NCL_station_3.py
Jiaqi-beep: I don't think xarray's open_dataset is meant for txt files. I had erros opening txt files with it:
 _did not find a match in any of xarray's currently installed IO backends ['netcdf4', 'scipy']. Consider explicitly selecting one of the installed backends via the ``engine`` parameter to xarray.open_dataset(), or installing additional IO dependencies:_

Another function to use for opening txt files are numpy's load_txt, but I think it is more cumbersome because we have to specify datatypes for each column when both floats and strings exist.

Let me know if I'm wrong about this!

Plots/Station/NCL_station_3.py
Jiaqi-beep: I found the function transform_points in Cartopy that can transform points between coordinate systems.

Plots/Station/NCL_station_3.py
anissa111: You're right I apologize! However, could you change the comment to reflect that you are using pandas instead of xarray

anissa111: Nevermind, I see you've already done this!

michaelavs (approved):
This looks good to me!

anissa111 (approved):
Looks great!

PR #405, opened 2021-08-03
fix polar_1 url

Jiaqi-beep:
- Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/polar_1.ncl
- Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/polar_1_lg.png


michaelavs (approved):
Thanks for catching that!

PR #409, opened 2021-08-12
conOncon1 with add_right_hand_axis

Jiaqi-beep:
Current plot:
image

Plot with add_right_hand_axis:
image


erogluorhan:
@michaelavs @anissa111 is the next step to add this utility function into geocat-viz and to use from there instead?

erogluorhan:
@anissa111 @michaelavs What is our plan with this?

pilotchute:
duplicate of #464 ?

jukent:
So this function is already in geocat-viz, going to remove the function block from the example and call in geocat-viz

jukent (changes requested):
I will fix this PR to use the new pressure height function before merging. The right y-axis is wrong.

jukent:
image

New plot using new height from pressure axis function

jukent:
Will fail until next GeoCAT-Viz release

jukent (approved):

kafitzgerald:
Do we know why the right axis is still slightly different from the original NCL plot?  I just noticed when looking through the history here.  

Looks like it might be a difference in the calculation coming from metpy, but I don't know why there'd be a diff.  

https://www.ncl.ucar.edu/Applications/Images/conOncon_1_lg.png

jukent:
Yes it is because there is a difference in the "standard atmosphere" used for conversion from pressure to height in metpy's function and I'm NCL. It doesn't seem worth supporting our own pressure height conversion at this time.

jukent:
If we do chose to change the height-pressure conversion, that would be in viz but we'd use the same function call here.

PR #412, opened 2021-08-24
Set up Google Analytics for RTD documentation page

erogluorhan:
This PR adds `docs/_templates/layout.html` that refers to a Google Analytics tracking ID to track analytics for https://geocat-examples.readthedocs.io/en/latest/

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,0 +1,15 @@
+{% extends "!layout.html" %} {% block extrahead %}
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script
+        async
+        src="https://www.googletagmanager.com/gtag/js?id=G-BY9T11S8QG"
+></script>
+<script>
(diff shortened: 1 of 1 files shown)


erogluorhan:
> Overall, this PR looks good to me. But before approving, Anissa and I were wondering if enabling the analytics via RTD admin pages (outlined here: https://docs.readthedocs.io/en/stable/guides/google-analytics.html) is different than the method used with this PR?

Great catch, thank you! I missed this. Technically, the builtin support of RTD and the changes in this PR should be doing the same thing: Adding a Google Analytics snippet into the header of the html. However, using builtin RTD support seems to be more convenient than changing our codes. 

FYI: I configured our RTD as suggested in the source you linked. Once the new page is generated and I see Google Analytics starts working, we could close this PR.

erogluorhan:
Closing this PR as @michaelavs and @anissa111 suggestion of the builtin RTD feature worked.

PR #413, opened 2021-09-07
Build failures

michaelavs:
Removing overlay_11b seems to fix build failure, testing with this branch for the "why?"


michaelavs:
@erogluorhan 
I've done some back and forth testing on how to get the RTD builds to fail and my final conclusion is that ```overlay_11b``` is causing the builds to fail, but I can't pinpoint an exact reason why. The tests I've done are as follows:
1. remove ```overlay_11b``` and see if RTD builds successfully (it did).
2. Try unpinning the matplotlib version in the environment file and have ```overlay_11b``` active in the repo. Build failed but gave a pointer to metograms
3. Remove meteograms and leave overlay and unpinned conda environment. Build failed on overlay.
4. Repin matplotlib version in environment. Build failed on overlay.

The reason given for```overlay_11b``` failing is due to variable "cf" is not being defined. The variable is defined in a for loop and when brought out of the for loop, the projection does not work. Upon looking more into what the goal of ```overlay_11b``` and ```overlay_11a``` were, I think we could completely take out ```overlay_11b``` and turn ```overlay_11a``` into just "overlay_11". The reason I believe this is that 11a and 11b were meant to show two ways of how to crop and mask data. The method in 11b is actually the same method used in all of the other examples in Overlays (using ax.extent) while the method used in 11a is not replicated in the other examples in Overlays (using masks).

However, due to the nature of what is causing the build failure from this file, I would definitely like a second set of eyes on the file to make sure I haven't overlooked anything that would lead to RTD thinking "cf" isn't defined.

erogluorhan:
Thank you for thorough testing of this @michaelavs ! It was very helpful for me.

I tried several ways to overcome the undefined `cf` variable issue in `NCL_overlay_11b` (defining it as a global variable, etc.) but I couldn't succeed. 

Your suggestion of turning `overlay_11a` into `overlay` would make sense to me, but I'd first like to do a fact check about `overlay_11b` with you:

I agree with you that `overlay_11a` and  `overlay_11b` are performing "masking" and "clipping", respectively, for the same overlay plot. However, I don't think `overlay_11b` is performing the clipping via `ax.extent`; instead, the method for this seems to be `set_clip_path`. In addition, when I check that entire for-loop, it is performing the "clipping" in a unique way that is described well by the comments. With that said, I think `overlay_11b` is unique among other overlay examples. However, I am still fine with getting rid of it if we can't fix the `cf` issue soon (as we spent reasonable time on this and the whole repo fails). At the least, do not hessitate to disable `NCL_overlay_11b` with this PR and merge it for helping the rest of the repo start working again. While it is disabled, you or someone else could continue digging into the `cf` issue.

michaelavs:
@erogluorhan Requesting your review because GitHub won't let me merge without one approval

erogluorhan (approved):

PR #419, opened 2022-01-04
Bumped oldest python version from 3.6 to 3.7

hCraker:
Changed oldest supported Python version from 3.6 to 3.7. Updated dependency list accordingly. Changed ci workflow to check against Python versions 3.7 and 3.9 rather than 3.6 and 3.9.


erogluorhan (approved):
Thanks for this!

PR #421, opened 2022-01-18
Adding calendar_average example

hCraker:
This PR adds an example script for the `geocat-comp` function `calendar_average`. The produced plot is shown below, and there is an explanation of what the plot shows in the beginning notes of the example.
image

### DIFF
--- a/GeoCAT-comp-examples/matplotlib/calendar_average_example.py
+++ b/GeoCAT-comp-examples/matplotlib/calendar_average_example.py
@@ -13,7 +13,7 @@ data:
 
 Dependencies:
     - geocat.comp
-    - geocat.datafiles (Not necessary for figure but used for accessing the data file)
+    - geocat.datafiles (for accessing data file only)
     - geocat.viz
     - cftime (installed with geocat.comp)
     - matplotlib (installed with geocat.viz)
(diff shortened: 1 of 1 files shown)


michaelavs (approved):
Looks good to me!

GeoCAT-comp-examples/matplotlib/calendar_average_example.py
```diff
@@ -0,0 +1,123 @@
...
+    - For GeoCAT functions: https://github.com/NCAR/geocat-comp
+    - For atm.20C.hourly6-1990-1995.nc file: https://github.com/NCAR/geocat-datafiles/tree/main/netcdf_files
+
+Dependencies:
+    - cftime
+    - geocat.comp
+    - geocat.datafiles (Not necessary for figure but used for accessing the data file)
+    - geocat.viz
+    - matplotlib
+    - xarray
```
erogluorhan, line 20: cftime and xarray is automatically installed when you have geocat-comp dependency (from a conda environment setup perspective), so we may want to put this information in parenthesis next to each of these packages here.

erogluorhan (commented):
It looks good to me except a few inline comments.

GeoCAT-comp-examples/matplotlib/calendar_average_example.py
hCraker: That makes sense. Should I also remove matplotlib from this list since it is automatically installed with geocat.viz?

erogluorhan: Maybe instead of removing these packages from the list, just give the information that through which package these packages are auto-installed.

erogluorhan (approved):
Looks great!

PR #422, opened 2022-01-18
Adding climatology_average example

hCraker:
This PR adds an example script for the `geocat-comp` function `calendar_average`. The produced plot is shown below, and there is an explanation of what the plot shows in the beginning notes of the example.

image

### DIFF
--- a/GeoCAT-comp-examples/matplotlib/climatology_average_example.py
+++ b/GeoCAT-comp-examples/matplotlib/climatology_average_example.py
@@ -1,5 +1,5 @@
 """
-climatology_example.py
+climatology_average_example.py
 =========================
 This script illustrates the following concepts:
     - Usage of geocat-comp's climatology_average function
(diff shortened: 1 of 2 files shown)


GeoCAT-comp-examples/matplotlib/climatology_average_example.py
```diff
@@ -0,0 +1,117 @@
...
+
+See following GitHub repositories to see further information about the function and how to access
+data:
+    - For GeoCAT functions: https://github.com/NCAR/geocat-comp
+    - For atm.20C.hourly6-1990-1995.nc file: https://github.com/NCAR/geocat-datafiles/tree/main/netcdf_files
+
+Dependencies:
+    - cftime
+    - geocat.comp
+    - geocat.datafiles (Not necessary for figure but for accessing the data file)
```
michaelavs, line 17: Would it be more accurate to say "necessary for figure but not for function" because you need the data file in order to create the figure below, but you don't necessarily need the file to use the function? Or possibly just changing the comment to be "for accessing datafile"?

hCraker: I'm trying to express that a user does not need `geocat.datafiles` if they have a different way to access their data. I'll replace it with "for accessing datafile only"

anissa111 (approved):
I agree with the previous comment, but once the "for accessing datafile only" change is made, this looks good to go, thank you!

michaelavs (approved):

PR #438, opened 2022-05-27
Panel 4

doqhne:
Closes issue #437

### DIFF
--- a/Gallery/Panels/NCL_panel_4.py
+++ b/Gallery/Panels/NCL_panel_4.py
@@ -1,6 +1,6 @@
 """
 NCL_panel_4.py
-================
+==============
 This script illustrates the following concepts:
    - Paneling three plots vertically on a page
    - Adding a common title to paneled plots
--- a/Gallery/Panels/NCL_panel_4.py
+++ b/Gallery/Panels/NCL_panel_4.py
@@ -8,7 +8,7 @@ Note: The colormap has been changed from the original NCL colormap in order to f
 This script illustrates the following concepts:
    - Paneling three plots vertically on a page
    - Adding a common title to paneled plots
-   - Adding a common labelbar to paneled plots
+   - Adding a common colorbar to paneled plots
    - Adding additional text at the bottom of a series of paneled plots
    - Subsetting a color map
 
--- a/Gallery/Panels/NCL_panel_4.py
+++ b/Gallery/Panels/NCL_panel_4.py
@@ -1,6 +1,10 @@
 """
 NCL_panel_4.py
 ==============
+Note: The colormap has been changed from the original NCL colormap in order to follow
+      best practices for colormaps. See more examples here:
+      https://geocat-examples.readthedocs.io/en/latest/gallery/index.html#colors
+
 This script illustrates the following concepts:
    - Paneling three plots vertically on a page
    - Adding a common title to paneled plots
@@ -23,7 +27,6 @@ import matplotlib as mpl
 from mpl_toolkits.axes_grid1.inset_locator import inset_axes
 import numpy as np
 import xarray as xr
-import cmaps
 
 import geocat.datafiles as gdf
 import geocat.viz as gv
@@ -75,15 +78,15 @@ def plot_labelled_filled_contours(data, ax=None):
     # Import an NCL colormap, truncate it, and save the colormap
 
     # Import the colormap
-    cmap = cmaps.gui_default
+    cmap = mpl.cm.get_cmap('OrRd', 13)
     # Create a linear segmented colormap using the colormap
-    newcmp = mpl.colors.LinearSegmentedColormap.from_list(
-        name="newcmap", colors=cmap(np.linspace(0.03, 0.9, 100)))
+    # newcmp = mpl.colors.LinearSegmentedColormap.from_list(
+    #     name="newcmap", colors=cmap(np.linspace(0.03, 0.9, 13)))
 
     handles = dict()
     handles["filled"] = data.plot.contourf(
         ax=ax,  # this is the axes we want to plot to
-        cmap=newcmp,  # our special colormap
+        cmap=cmap,  # our colormap
         levels=levels,  # contour levels specified outside this function
         transform=projection,  # data projection
         add_colorbar=False,  # don't add individual colorbars for each plot call
@@ -101,7 +104,7 @@ def plot_labelled_filled_contours(data, ax=None):
         add_labels=False,  # again turn off automatic labels
     )
 
-    #Label the contours
+    # Label the contours
     ax.clabel(
         handles["contour"],
         levels=np.arange(-10, 50, 10),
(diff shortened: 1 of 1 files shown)


Gallery/Panels/NCL_panel_4.py
```diff
@@ -30,14 +30,17 @@
 ###############################################################################
 # Read in data:
 
-# Open a netCDF data file using xarray default engine and load the data into
-# xarrays, choosing the 2nd timestamp
-ds = xr.open_dataset(gdf.get("netcdf_files/uv300.nc")).isel(time=1)
+# Open a netCDF data file using xarray default engine and save as a variable
+ds = xr.open_dataset(gdf.get("netcdf_files/uv300.nc"))
 
 ###############################################################################
```
michaelavs, line 36:
```suggestion

```

I think you should combine the "Read in data" and "save zonal and meridional..." into one block underneath "Read in data"

Gallery/Panels/NCL_panel_4.py
```diff
@@ -30,14 +30,17 @@
...
-# Utility Function: Labelled Filled Contour Plot:
 
-# Define a utility plotting function in order not to repeat many lines of codes
+# save the zonal and meridional wind separately, select July data
+zonal = ds.U.isel(time=1)
+meridional = ds.V.isel(time=1)
+
+###############################################################################
+# Define a utility plotting function in order not to repeat many lines of code
```
michaelavs, line 43: This is left over from the original, but I feel like the flow is not quite the best. My thought would be to either make this under the header of "Define Plotting Helper function" and then have individual comments, like we see in line 43, then keep line 134 and on how they are. What do you think, @anissa111?

michaelavs (commented):
Looks good overall! My only two comments are more about overall formatting (one of which is from how this notebook was formatted before). Pending anything Anissa may see, I think this is in good shape to be approved once everything is addressed!

Gallery/Panels/NCL_panel_4.py
```diff
@@ -10,12 +10,12 @@
 
 See following URLs to see the reproduced NCL plot & script:
     - Original NCL script: https://www.ncl.ucar.edu/Applications/Scripts/panel_4.ncl
-    - Original NCL plot: https://www.ncl.ucar.edu/Applications/Images/panel_4_lg.png
+    - Original NCL plots: https://www.ncl.ucar.edu/Applications/Images/panel_4_lg.png
```
anissa111, line 13: Why this change here?

anissa111, line 13: Why this change here?

Gallery/Panels/NCL_panel_4.py
anissa111: I also agree that the flow is a little off. I think the header you suggested is way better

Gallery/Panels/NCL_panel_4.py
```diff
@@ -128,7 +131,7 @@ def plot_labelled_filled_contours(data, ax=None):
 
 
 ###############################################################################
-# Plot:
+# Create plot
```
anissa111, line 134: I agree here actually, but the previous version is actually pretty consistent across all examples. Maybe this could be a separate PR where we update some section headings?

anissa111 (commented):
I've left several inline comments below and a couple additional suggestions in this comment

## Remove colorbar truncation util function from script?
``` python
# Import an NCL colormap, truncating it by using geocat.viz.util convenience function
newcmp = gv.truncate_colormap(cmaps.gui_default, minval=0.03, maxval=0.9)
```
I think this section probably does not need to be hidden in a util function. The actual process to make this color bar is probably something useful to the end user. I'm not against keeping the util function, but I think we could do this part manually in the example. We could also add it to the top doc string in 'concepts illustrated'

## Check on updated methods for specifying `edgecolors`
``` python
# matplotlib's "contourf" doesn't let you specify "edgecolors",
# instead we use matplotlib's "contour" to plot contour lines on top of the filled contours
handles["contour"] = data.plot.contour(
    ax=ax,
    levels=levels,
    colors="black",  # note plurals in this and following kwargs
    linestyles="-",
    linewidths=0.5,
    add_labels=False,  # again turn off automatic labels
    )
```
I swear there's a better way to handle this now. @michaelavs, do you know what I'm talking about? I feel like we don't use this in more recent examples.

Gallery/Panels/NCL_panel_4.py
anissa111: I agree here

Gallery/Panels/NCL_panel_4.py
anissa111: Yeah, the heading you suggested is much better and more consistent with later examples. We might want to make sure that messing with the "section" doesn't produce any strange Sphinx results

doqhne:
I did not find a way to add edgecolors to the filled contour lines, and did not see another method in any other examples, so I left the contour section as is.

Gallery/Panels/NCL_panel_4.py
```diff
@@ -92,7 +101,7 @@ def plot_labelled_filled_contours(data, ax=None):
         add_labels=False,  # again turn off automatic labels
     )
 
-    # Label the contours
+    #Label the contours
```
anissa111, line 104: Why was this change made?

Gallery/Panels/NCL_panel_4.py
```diff
@@ -68,8 +72,13 @@ def plot_labelled_filled_contours(data, ax=None):
...
 
-    # Import an NCL colormap, truncating it by using geocat.viz.util convenience function
-    newcmp = gv.truncate_colormap(cmaps.gui_default, minval=0.03, maxval=0.9)
+    # Import an NCL colormap, truncate it, and save the colormap
+
+    # Import the colormap
+    cmap = cmaps.gui_default
+    # Create a linear segmented colormap using the colormap
+    newcmp = mpl.colors.LinearSegmentedColormap.from_list(
+        name="newcmap", colors=cmap(np.linspace(0.03, 0.9, 100)))
```
anissa111, line 81: We might want to change the colormap here to reflect best practices as in other examples

anissa111 (commented):
This looks almost ready, just a few very minor comments!

Gallery/Panels/NCL_panel_4.py
```diff
@@ -68,13 +75,18 @@ def plot_labelled_filled_contours(data, ax=None):
         tick labels.
     """
 
-    # Import an NCL colormap, truncating it by using geocat.viz.util convenience function
-    newcmp = gv.truncate_colormap(cmaps.gui_default, minval=0.03, maxval=0.9)
+    # Import an NCL colormap, truncate it, and save the colormap
+
+    # Import the colormap
+    cmap = mpl.cm.get_cmap('OrRd', 13)
+    # Create a linear segmented colormap using the colormap
```
michaelavs, line 82: Line 82 through 84 can be removed (or just double check that it isn't accidentally being commented out if it is required to address a different topic)

Gallery/Panels/NCL_panel_4.py
```diff
@@ -68,13 +75,18 @@ def plot_labelled_filled_contours(data, ax=None):
         tick labels.
     """
 
-    # Import an NCL colormap, truncating it by using geocat.viz.util convenience function
-    newcmp = gv.truncate_colormap(cmaps.gui_default, minval=0.03, maxval=0.9)
+    # Import an NCL colormap, truncate it, and save the colormap
```
michaelavs, line 78: Are we still importing the colormap from MCL or is it coming from matplotlib?

anissa111 (approved):

PR #441, opened 2022-06-01
Vector 1

doqhne:
Changed tick marks to match NCL plot and changed colorbar to match NCL plot.

Closes issue #440

Previous version:
previous_version

New version: 
new_version

Original NCL plot:
NCL_original

### DIFF
--- a/Gallery/Vectors/NCL_vector_1.py
+++ b/Gallery/Vectors/NCL_vector_1.py
@@ -69,7 +69,7 @@ plt.figure(figsize=(8, 5))
 ax = plt.axes(projection=proj)
 
 # add land feature and zoom in on desired location
-ax.add_feature(cfeature.LAND, facecolor="lightgrey", zorder=1)
+ax.add_feature(cfeature.LAND, facecolor="lightgrey", zorder=2)
 ax.set_extent((66, 96, 5, 25), crs=ccrs.PlateCarree())
 
 # Create the filled contour plot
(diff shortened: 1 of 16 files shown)


Gallery/Vectors/NCL_vector_1.py
```diff
@@ -72,80 +95,68 @@
...
-
-# Add color bar
+# Add minor tick marks
+gv.add_major_minor_ticks(ax,
+                         x_minor_per_major=4,
+                         y_minor_per_major=4,
+                         labelsize=14)
+
+# Add and customize colorbar
 cbar_ticks = np.arange(24, 29.1, .3)
```
michaelavs, line 131:
```suggestion
cbar_ticks = np.arange(24, 29, .3)
```
I know you didn't edit this part, but for some reason the endpoint of this range was being included in the colorbar. Just switching it to 29 will resolve this.

michaelavs (commented):
Looks good overall, just a couple of comments to be addressed!

anissa111:
A couple points:

On my side, the background for the label is not showing up. Also, if you look at the screenshot below, it seems like we might be missing a level above 28.8, which would also make sense because it seems like the original plot has a section of darker purple that the recreation does not. 

Also, we might want to consider switching colormaps to reflect best practices as in the other examples

image

doqhne:
I could not get the color bar edges to extend to the end of the colorbar, but the levels on the actual plot are correct now.

anissa111 (approved):
Looks good!!

Issue #446, opened 2022-06-14
Coneff 11


doqhne:
Recreate coneff_11 NCL plot

Link to original NCL plot:
https://www.ncl.ucar.edu/Applications/Images/coneff_11_2_lg.png

PR #447, opened 2022-06-14
Coneff 11

doqhne:
Resolves issue #446 

Original NCL plot:

coneff_11_NCL

Recreated version with python:

coneff_11_python

### DIFF
--- a/Gallery/Contours/NCL_coneff_11.py
+++ b/Gallery/Contours/NCL_coneff_11.py
@@ -19,9 +19,8 @@ See following URLs to see the reproduced NCL plot & script:
 
 import numpy as np
 import xarray as xr
-import cartopy.crs as ccrs
 import matplotlib.pyplot as plt
-import cmaps
+import matplotlib as mpl
 
 import geocat.datafiles as gdf
 import geocat.viz as gv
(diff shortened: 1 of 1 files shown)


Gallery/Contours/NCL_coneff_11.py
```diff
@@ -0,0 +1,127 @@
...
+
+###############################################################################
+# Plot
+
+# Generate figure (set its size (width, height) in inches)
+plt.figure(figsize=(8, 10))
+ax = plt.axes()
+
+# Choose hatches to fill the contours
+hatches = ['/////', '/////', '/////', '/////', None, '..', '.', '.', '.']
```
hCraker, line 46: The original plot has different densities of the dots for the different regions. Can you see if there is a way to do that? It's okay if the densities in matplotlib are different from NCL. As long as we're showing the different options, I think it'll be good.

Gallery/Contours/NCL_coneff_11.py
```diff
@@ -0,0 +1,127 @@
...
+                          https://www.ncl.ucar.edu/Applications/Images/coneff_11_2_lg.png
+"""
+
+###############################################################################
+# Import packages:
+
+import numpy as np
+import xarray as xr
+import matplotlib.pyplot as plt
+import matplotlib as mpl
```
hCraker, line 24:
```suggestion
```
Unused import

Gallery/Contours/NCL_coneff_11.py
```diff
@@ -0,0 +1,127 @@
+"""
+NCL_coneff_11.py
+================
+
+This script illustrates the following concepts:
+   - Filling contours with multiple shaded patterns
+   - Filling contours with stippling (solid dots)
+   - Changing the size of the dot fill pattern in a contour plot
+   - Overlaying a stipple pattern to show area of interest
+   - Changing the density of contour shaded patterns
```
hCraker, line 10: I recommend updating the language here to be consistent with Matplotlib (i.e. "shaded patterns" --> "hatches") and to mention the work arounds we had to find (i.e. creating a color bar with colored hatch/stippling patterns). This will make it easier for NCL users converting to Python to learn the new terminology and for all users to find examples they're searching for.

hCraker (changes requested):
This looks really good! I've made a few minor comments about the stippling pattern and the file description at the beginning

Gallery/Contours/NCL_coneff_11.py
doqhne: The densities are different in Matplotlib, and unfortunately have a large increase/decrease in density with each change in level. The '...' level almost looks like a solid color, so I had to stick to just the '..' and '.' options.

hCraker: Ah that's understandable. I'll go ahead and approve this then

hCraker (approved):

PR #449, opened 2022-06-21
Unpin matplotlib version

hCraker:
In `conda_environment.yml`, the version for matplotlib has been pinned at version 3.3.0 since commit https://github.com/NCAR/GeoCAT-examples/commit/d154346015520edfa52575a236718b502a39d1e4 on August 21, 2020. Since then, matplotlib has had numerous updates. To stay current with our dependencies, this PR unpins the matplotlib version and resolves any deprecation issues that arrive as a result.


hCraker:
This is what the conda environment now looks like when installed using the gallery instructions.
[conda_list.txt](https://github.com/NCAR/GeoCAT-examples/files/8952078/conda_list.txt)
A consequence of removing the pin is that Python 3.10 is automatically installed by conda. We currently only check for compatibility with Python 3.8 and 3.9. Do we want to expand our support to Python 3.10?

anissa111 (approved):

hCraker:
> > This is what the conda environment now looks like when installed using the gallery instructions. [conda_list.txt](https://github.com/NCAR/GeoCAT-examples/files/8952078/conda_list.txt) A consequence of removing the pin is that Python 3.10 is automatically installed by conda. Currently, we only have CI checks for Python 3.7 and 3.9 (not for 3.8 for some reason). Do we want to expand our support to Python 3.10?
> 
> I think it's reasonable to include tests for 3.10 as well

@anissa111 @erogluorhan I haven't made CI tests before. Could either of you walk me through how to do it in a separate PR? We need tests for both 3.8 and 3.10.

anissa111:
> @anissa111 @erogluorhan I haven't made CI tests before. Could either of you walk me through how to do it? We need tests for both 3.8 and 3.10. I don't want to merge this PR until we provide tests for at least 3.10.

Yeah! This is actually a really easy CI thing. Can I push to this branch to show you the change?

PR #450, opened 2022-06-21
Update citation texts per Zenodo DOI implementation per-version

erogluorhan:
We are starting to generate DOIs for each GeoCAT-comp version via the Zenodo minting service. Texts regarding citation are updated to reflect this change.


hCraker (approved):

anissa111 (approved):
Looks good!

PR #469, opened 2022-07-13
Bug fix for multiple axes

hCraker:
`NCL_vector_1.py`, `NCL_vector_3.py`, and `NCL_vector_4.py` all had an issue with multiple axes. This PR fixes that bug by replacing `plt.subplots` with `plt.figure` since none of the scripts required multiple axes.

This bug is recent and the cause of it is still uncertain given that `matplotlib` has not had recent updates and the recently released `cartopy` [v0.20.3](https://github.com/SciTools/cartopy/releases/tag/v0.20.3) has no changes to `GeoAxes` in it's change log.


michaelavs (approved):
Thank you!

Issue #473, opened 2022-07-15
mac-os tests bottleneck


anissa111:
Should we consider reducing the number of tests that run on mac-os? We have 4 mac-os runs per run and a concurrency limit of 5. 

I am unclear about the purpose of testing all the versions on mac-os as well as ubuntu, as the differences between the two versions would likely be cosmetic rather than errors, which GitHub Actions do not account for. Is there any scenario where mac-os would outright fail where ubuntu would not?

anissa111:
@hCraker @erogluorhan @michaelavs

erogluorhan:
I'd say testing for both macos and ubuntu is important because in the dependency list, we have platform-specifically built tools such as geocat-comp and WRF-Python and likely a few others. They have different builds for different platforms, and there is a chance that at some point we'd have errors in a platform but not on the other. 

Regarding the concurrency limit, 5 is too low :) Anyways, if that's really an issue, we'd think of testing macos only for the lowest and highest supported Python versions for push/pull-requests but creating a new workflow to test macos only for all the versions (what we currently do) to be run nightly on the main branch only (this would mean splitting the current `ci.yml` to seperate the cron-job from the others and to specify branch name.

anissa111:
> I'd say testing for both macos and ubuntu is important because in the dependency list, we have platform-specifically built tools such as geocat-comp and WRF-Python and likely a few others. They have different builds for different platforms, and there is a chance that at some point we'd have errors in a platform but not on the other.

Ahh, makes sense!

> Regarding the concurrency limit, 5 is too low :) Anyways, if that's really an issue, we'd think of testing macos only for the lowest and highest supported Python versions for push/pull-requests but creating a new workflow to test macos only for all the versions (what we currently do) to be run nightly on the main branch only (this would mean splitting the current `ci.yml` to separate the cron-job from the others and to specify branch name.

Yeah, that's essentially what I was thinking. 

Normally it probably won't matter that much, but occasionally there will be times (like tomorrow) where I anticipate multiple PRs being approved/merged/updated in a fairly short timespan where the concurrency limit will really get us.

PR #474, opened 2022-07-15
Reduce Concurrent Tests

anissa111:
Addresses #473

Summary of changes:
- adds separate `ci-nightly.yml` workflow that runs on all `os` and `python-version`s at 1 AM nightly
- modifies `ci.yml` to cancel previous workflows and to run `macos-latest` on oldest and newest versions of python we support (3.7 and 3.10) to help concurrency limit (all versions run nightly)

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -1,0 +1,37 @@
+name: CI
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 1 * * *' # Daily “At 01:00”
+
+jobs:
+  build-and-run:
+    name: Build Sphinx Docs (${{ matrix.python-version }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ "ubuntu-latest", "macos-latest"]
+        python-version: [ "3.7", "3.8", "3.9", "3.10" ]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Create conda environment
+      env:
+        ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
+      uses: mamba-org/provision-with-micromamba@main
+      with:
+         micromamba-version: "latest"
+         environment-name: geocat-examples
+         environment-file: conda_environment.yml
+    - name: Make html
+      shell: bash -l {0}
+      working-directory: ./docs
+      run: |
+        conda info
+        conda list
+        make html
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@ name: CI
 on:
   pull_request:
   workflow_dispatch:
-  schedule:
-  - cron: '0 0 * * *' # Daily “At 00:00”
 
 jobs:
   build-and-run:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ "ubuntu-latest", "macos-latest"]
+        os: [ "ubuntu-latest"]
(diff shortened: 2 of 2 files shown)


anissa111:
Note that previous tests on a branch can now be automatically cancelled as shown last commit

michaelavs (approved):
Approving as it looks like all checks will pass with these changes (I assume the sphinx one would/will pass if given enough time)

anissa111:
That... actually shouldn't be there. Huh. I wonder if it's because it's not in main? If you click on the actions for the PR, it shows them all as passing. This PR specifically removed macos-latest 3.9

hCraker:
> Approving as it looks like all checks will pass with these changes (I assume the sphinx one would/will pass if given enough time)

@michaelavs Do you think we should wait for it to pass before merging, or is it worth risking? I highly doubt a preemptive merge will break everything (I could regret saying that)

anissa111:
@hCraker We don't have the permissions to do that anymore haha

hCraker:
> @hCraker We don't have the permissions to do that anymore haha

Oh darn I didn't even notice we lost that since I used it so rarely

anissa111:
@erogluorhan We think this might need your admin privileges. The checks on the PR start with just those we've made in the PR and then adds at least one extra expected required test (from main?) that is one of the tests that we've removed in this PR and on it indefinitely, which blocks merging.

Here is a screenshot of the tests right after I closed and reopened the PR, which lists the expected number of tests
image

Here is a screenshot taken a few minutes afterward that shows that two extra tests have been added. I think one is `docs/readthedocs.org:geocat-examples` and the other is `macos-latest 3.9`
image1

I am unclear why only `macos-latest 3.9` is acting like this and not `macos-latest 3.8`, which was also removed in this PR, but I suspect it is because 3.8 tests were added relatively recently and might not be part of the branch protection rules for main, but we don't have the permissions to see those, either.

If you look at the actions tab, it shows [all tests pass](https://github.com/NCAR/GeoCAT-examples/actions/runs/2678655490) and lists the expected combinations of os and python version.

When you have a minute, could you check this out and maybe lend this PR your magic admin privileges?

erogluorhan:
> > Approving as it looks like all checks will pass with these changes (I assume the sphinx one would/will pass if given enough time)
> 
> @michaelavs Do you think we should wait for it to pass before merging, or is it worth risking? I highly doubt a preemptive merge will break everything (I could regret saying that)

It should (almost) never be worth risking the `main` branch without seeing all of them pass.

erogluorhan:
> @hCraker We don't have the permissions to do that anymore haha

As soon as we have package leaders listed, let us give them the admin privileges, too (but, with they swearing not to risk the `main` branch in any circumstances haha)

erogluorhan:
@anissa111 sorry if I am too late, but I figured why only 3.9 was acting like that. Here is what the current state of our branch protection rule was (3.8 was not listed there):

image

anissa111:
@erogluorhan No problem! I figured it was something like that. I suggest removing `macos-latest 3.9` and adding `macos-latest 3.10`, `ubuntu-latest 3.8`, and `ubuntu-latest 3.10`

anissa111:
@erogluorhan Great, thank you!!

.github/workflows/ci.yml
```diff
@@ -16,8 +18,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ "ubuntu-latest", "macos-latest"]
+        os: [ "ubuntu-latest"]
```
erogluorhan, line 21: Instead of removing macos-latest at all, I'd do ubuntu and macos but only for versions 3.7 and 3.10 for this pull-request CI

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,37 @@
+name: CI
```
erogluorhan, line 1: I'd name this CI Nightly

anissa111: I agree! See https://github.com/NCAR/GeoCAT-examples/pull/475

Issue #481, opened 2022-07-28
Nightly CI Fails


anissa111:
Nightly CI failed across the board on `NCL_skewt_2_2.py` and `NCL_skewt_3_2.py`

See [GH Actions logs](https://github.com/NCAR/GeoCAT-examples/runs/7551835878?check_suite_focus=true#step:4:2211)

PR #482, opened 2022-07-28
Fix for skewt plots

anissa111:
Addresses #481 

Summary of changes:
- remove units from `p` for `metpy.plot_barbs` call
- comments about making `p` unitless

Notes:
- Not clear why this started failing, other than it _might_ be connected to the newest sphinx (5.5.1) release on Wednesday.
- This change is, as far as I can tell, only reproducible by running `make html`. Neither `skewt_2_2` or `skewt_3_2` have had issues running locally outside of sphinx generation
- Might be a bug we should report to metpy.


michaelavs (approved):
Assuming all of the checks pass, this looks good to me!

Issue #484, opened 2022-09-15
Link in the docs page is broken


erogluorhan:
**Describe the bug**
A link at the Installation page of the documentation is broken.

**To Reproduce**
1. Go to https://geocat-examples.readthedocs.io/en/latest/install.html
2. Click on the link "[Contributor’s Guide](https://geocat-examples.readthedocs.io/en/latest/(https://geocat.ucar.edu/pages/contributing.html#3-creating-a-development-environment))", which is at the very end of the first section, "Installation"
3. It has a broken link

**Expected behavior**
Should have  linked to https://geocat.ucar.edu/pages/contributing.html

Issue #320, opened 2021-04-12
GeoCAT ReadTheDocs Example Homepage Spelling Error


lkailynncar:
On the GeoCAT examples homepage on ReadTheDocs, the word "functionalitie" is used in the sentence "In addition, geocat-datafiles is used as a dataset storage and geocat-viz is used for a higher level implementation for low level matplotlib functionalitie."  Did you perhaps mean "functionality" or "functionalities"?

PR #487, opened 2022-11-19
Fixes for #320 #484

pilotchute:


hCraker (approved):
Thanks for fixing these typos!

Issue #495, opened 2023-03-27
Unpin xarray version once root cause is found


hCraker:
The checks for geocat-examples are currently [failing](https://github.com/NCAR/GeoCAT-examples/actions/runs/4527728492/jobs/7973849215#step:4:2287) with the latest xarray release [v2023.03.0](https://docs.xarray.dev/en/stable/whats-new.html#v2023-03-0-march-22-2023). A PR has been made to pin xarray to the prior version, but that is only a temporary solution. We will need to solve the root cause and unpin the version at a later date.

kafitzgerald:
Noting that geocat-examples is pinned, because of this issue in geocat-comp: https://github.com/NCAR/geocat-comp/issues/381

kafitzgerald:
Once xarray is (fully) pinned for comp in this upcoming release we should be able to remove our version pin here and just take care of things in comp instead of tracking elsewhere as well.   Same goes for viz and I'll make a note there as well.

PR #496, opened 2023-03-27
Pin xarray in conda_environment.yml

hCraker:
This PR pins the xarray version to <=2023.02.0 as the latest versoin of xarray is causing checks to fail. This is a temporary fix, see #495 for more details


anissa111 (approved):
Approving even though checks are failing because failures are not caused by this PR.

PR #500, opened 2023-04-07
Right vertical axis height be calculated

### LINKED ISSUE #499: Help needed to calculate height for NCL_h_lat_6.py
I am working on the visualization of the CESM output with the GeoCAT template and one type of plot is the vertical distribution (see example at https://geocat-examples.readthedocs.io/en/latest/gallery/Contours/NCL_h_lat_6.html#sphx-glr-gallery-contours-ncl-h-lat-6-py).

According to the notes in the example above, the height values on the right y-axis are not calculated based on the model output, but just arbitrary data. Is there any existing function I could use to calculate the height values based on my model output first and plot them correspondingly? Thanks.

jukent:
Uses MetpY's [`pressure_to_height`](https://unidata.github.io/MetPy/latest/api/generated/metpy.calc.pressure_to_height_std.html) method to calculate rightmost y-axis height from standard atmosphere, instead of arbitrarily.

This fix is made in 4 plots NCL_h_lat_6, NCL_h_lat_7, and NCL_h_long_5, NCL_vector_5.
Closes #499


Gallery/Contours/NCL_h_lat_6.py
```diff
@@ -104,12 +105,17 @@
...
+# Use MetPy's pressure_to_height_std function to get standard atmosphere height conversion
+heights = mpcalc.pressure_to_height_std(ds.lev).values
+min_height = min(heights)
+max_height = max(heights)
+
 # Use geocat.viz.util convenience function to set axes tick values
-gv.set_axes_limits_and_ticks(axRHS, ylim=(0, 32), yticks=np.arange(4, 32, 4))
+gv.set_axes_limits_and_ticks(axRHS,
+                             ylim=(min_height, max_height),
+                             yticks=np.arange(4, max_height, 4))
```
anissa111, line 118: Wait, here shouldn't we be doing something like getting the positions (in height) we want to label and then labeling those positions? Definitely correct me if I'm reading this wrong but won't this just evenly space some ticks between the max and min position?

jukent: You are correct that this will evenly space ticks, but now those ticks and their labels are at an accurate location corresponding to the left vertical axis (pressure). Whereas before, they were arbitrary numbers that did not match the pressure values.

The left ticks are not in identical spots to the right ticks though, but I don't think that is what the user would want and the numbers wouldn't be as pretty.

jukent: I think if the user has specific height values they want to plot (as opposed to integers within the height range at set intervals), it would be pretty straightforward for them to use a numpy array with those values for yticks, as long as min_height and max_height are calculated from pressure.

jukent: Or maybe I have to think about this longer

anissa111: I think I'm confused about the positions of the ticks between the min and max values. I don't think we can assume that they'll be evenly spaced, right?

jukent: https://www.ncl.ucar.edu/Applications/Images/h_lat_6_lg.png Do these look evenly spaced to you? They look slightly off. Well it is at least closer to true than what we had -- but I think you're right. Now the top and bottom are pinned correctly, but the heights are not necessarily 1:1 along the axis.

anissa111: Also, I'm a little concerned that axis on the new plot looks pretty different from the original ncl.

anissa111: > https://www.ncl.ucar.edu/Applications/Images/h_lat_6_lg.png Do these look evenly spaced to you? They look slightly off. Well it is at least closer to true than what we had -- but I think you're right. Now the top and bottom are pinned correctly, but the heights are not necessarily 1:1 along the axis.

Yeah, I don't think those are evenly spaced, but just barely. Especially the top two tick marks. I'm assuming this was to get nice, evenly spaced values for the labels?

jukent: If I change the ticks to 

```
# Create second y-axis to show geo-potential height.
axRHS = ax.twinx()

# Use MetPy's pressure_to_height_std function to get standard atmosphere height conversion
heights = mpcalc.pressure_to_height_std(ds.lev).values
min_height = min(heights)
max_height = max(heights)

axRHS.set_ylim(min_height, max_height)
axRHS.set_yticks(heights)
```

I get 
image
is this closer? The numbers aren't pretty, but are they at least at the correct spaces?

jukent: In this [stackoverflow question](https://stackoverflow.com/questions/14153816/matplotlib-axis-labels-how-to-find-out-where-they-will-be-located) people seem to imply that altitude is linear, but I don't think they're right.

jukent: NCL seems to do the conversion here: https://github.com/NCAR/ncl/blob/8a96101fe14d0cf0f5ed66a5e6b1733084bc69df/ni/src/examples/gsun/gsn_csm.ncl#L2659

I will try to recreate this in Python

anissa111: So, I think the thing to do might be to think of this the other way around. 

What if we set both y-axes to be the same scale. Then, use `height_to_pressure_std` instead of `pressure_to_height_std` to get the pressure values that correspond to `np.arange(4,24,4)` (or something like that). Then, set the RHS tick values to be the corresponding calculated pressures and set the labels to be the height values.

This is convoluted enough it might make sense to make this into a geocat-viz function

jukent: image

I followed the NCL steps and produced code that should probably be in its own function to be reused in other plots that have altitude and pressure plotted. It still doesn't match the NCL so I'd love to confer with someone on the math next week:

```
# Create second y-axis to show geo-potential height.
axRHS = ax.twinx()

# Use MetPy's pressure_to_height_std function to get standard atmosphere height conversion
heights = mpcalc.pressure_to_height_std(ds.lev).values

min_height = min(heights)
max_height = max(heights)

# Auto select "nice" heights based on range
height_range = max_height - min_height
if (height_range < 35):
    step = 4
else:
    if (height_range < 70):
        step = 7
    else:
        step = 10
  
heights_nice = np.arange(int(min_height), int(max_height), step)

# Use a cubic spline interpolation to send "nice" height values back to pressure tick locations
cubic_spline = scipy.interpolate.CubicSpline(heights, np.flip(ds.lev))
pressures_nice = cubic_spline(heights_nice)

gv.set_axes_limits_and_ticks(axRHS,
                            ylim=ax.get_ylim()[::-1],
                            yticks=pressures_nice,
                            yticklabels=heights_nice)
axRHS.tick_params(labelsize=12)  # manually set tick label size
```

jukent: > ght be to think of this the other way around.

Just saw this. I like this idea. We might still need to do both conversions to get our "nice" heights range, and we can make this a fx where the user inputs their step size

anissa111: > Just saw this. I like this idea. We might still need to do both conversions to get our "nice" heights range, and we can make this a fx where the user inputs their step size

Yeah! I imagine the user could provide an array of their intended tick locations as an `np.arange` or list or something

jukent: Okay trying your suggestion (I like the idea of using metpy again instead of scipy) I have the following code:
```
# Create second y-axis to show geo-potential height.
axRHS = ax.twinx()

# Use MetPy's pressure_to_height_std function to get standard atmosphere height conversion
heights = mpcalc.pressure_to_height_std(ds.lev).values
heights_nice = np.arange(int(min(heights)), int(max(heights)), 4)

# Send "nice" height values back to pressure tick locations
pressures_nice = mpcalc.height_to_pressure_std(heights_nice * units('km')).magnitude

gv.set_axes_limits_and_ticks(axRHS,
                            ylim=ax.get_ylim(),
                            yticks=pressures_nice,
                            yticklabels=heights_nice)
```
It is flipped somehow, which I keep going in circles trying to change
image

But I will keep working on this next week.

jukent: The biggest problem was that I had to set the second axis to log as well (even though the plt as a whole was set to it).

It ended up only being a few lines to get the "nice" height pressure values and setting it to log. So I'm not sure if we need a new function at this stage.

jukent: image
Just saw that I didn't post the new image with the logarithmic right y-axis

jukent:
> So, if what's connecting the scale of the two axes is the min and max values, why aren't the min and max values visually aligned? Does that make sense??

Thanks for your comment @anissa111 and your help with figuring this out. But I'm not sure if I understand what you mean by the min/max values not visually aligning.

I'll outline what I do understand:

Here is the current code for the right axis
```
# Create second y-axis to show geo-potential height.
axRHS = ax.twinx()

# Select heights to display as tick labels
heights_nice = np.arange(0,32,4)
# Send "nice" height values back to pressure  as tick locations
pressures_nice = mpcalc.height_to_pressure_std(heights_nice * units('km')).magnitude

axRHS.set_yscale('log')
gv.set_axes_limits_and_ticks(axRHS,
                            ylim=ax.get_ylim(),
                            yticks=pressures_nice,
                            yticklabels=heights_nice)
axRHS.minorticks_off()
```

So technically both the left and right y-axis are in Pressure, and the limits of the right are inherited from the left. We just picked pretty height values, and found their corresponding pressure values to indicate where they go.

Setting the yscale to logrithmic is vital, but the log has auto minor tick labels that are at natural pressure values so they look confusing with our selected height values (because they are subdividing a different unit)

The min value is technically ~0.1 and isn't displayed because the "pretty" value of 0 is off the chart. I'm no longer actively working with the min values. I'd be happy to discuss this in a meeting. It is a tad confusing, especially since I think our "height to pressure" doesn't seem to be based on the exact same standard atmosphere as NCL's (the max value is under 28).

As far as I can tell the NCL code works by:
1. converting pressures to standard atmosphere heights
2. inferring an array of pretty values by taking the integer min and max values and picking a step size
3. using a cubic spline interpolation to turn these back to pressure (we can just use a height to pressure conversion just as easily)
4. plot with tick marks at pressures and labels at heights.

I'm fairly confident the discrepancy is in step 1.

I hope that makes sense! Happy to meet to discuss more.

anissa111:
Oh, I missed that you'd stopped working with min and max values, sorry! That explanation does super help, thank you for taking the time to write it out. I think I'm convinced now and will approve.

I'm going to blame Monday brain 😆

anissa111 (approved):

jukent:
No worries. Thanks @anissa111 I think at the end of the day, the extra steps are so few lines that I'm not sure if we still need a separate function

PR #501, opened 2023-04-26
update examples to use height-pressure-axis function

jukent:
After this commit this should be a valid simplification of these examples: https://github.com/NCAR/geocat-viz/commit/00917c70ee18b8df34c02b1b9685d5c476462c7d

Does our GeoCat-examples environment grab the main branch of Viz? That would be very helpful for development. Or does it grab from the latest release version? I need to learn more about how this is set up.


jukent:
pre-commit passed before I had to rerun all the tests because readthedocs timed out and there was no option to re-run just that test. Nothing has changed since it passed so even with this one test lagging, I think this is safe to merge.

kafitzgerald (approved):

kafitzgerald:
Looks good.  

Thanks for cleaning this up!  So much less code.

jukent:
Of course since pulling from main the tests have to run all over again!

PR #506, opened 2023-05-11
Updates to Gallery to use new function names and kwargs

jukent:
Functions and kwargs changed to match flake8 and snake_case rules in https://github.com/NCAR/geocat-viz/pull/127 here are the respective changes in calling these fxs in the examples gallery.


kafitzgerald:
Shoot, it looks like the readthedocs build is still failing.  

"Command killed due to timeout or excessive memory consumption"

Guessing we're running up against [resource limits](https://docs.readthedocs.io/en/stable/builds.html), since it was running for almost 20 minutes and I would guess we're on the community plan which allows for 15 min 😬 .

kafitzgerald (approved):
Looks good!

PR #515, opened 2023-06-07
Update cm calls

### LINKED ISSUE #505: deprecated matplotlib colormaps functions in NCL_vector_4.py
`plt.cm.register_cmap` and `plt.cm.get_cmap`, both in [NCL_vector_4.py](https://github.com/NCAR/geocat-examples/blob/main/Gallery/Vectors/NCL_vector_4.py), are both [deprecated as of version 3.7 of matplotlib](https://matplotlib.org/stable/api/cm_api.html).  We should probably update these and clean up that section of the script.  The re-registration of colormaps is likely also problematic in more recent versions of matplotlib.

kafitzgerald:
Clean up the colormaps portion of `NCL_vector_4.py` script to remove deprecated matplotlib functions and unnecessary code.

Closes #505.


kafitzgerald:
Ok, so I removed the commit unpinning matplotlib for now (until I can address the re-registration issue in viz). 

I'm not quite sure what happened here with the readthedocs build failing...  Going to rerun these tests in a minute.  Doesn't look related to this PR and everything should pass w/ this minor (tested) change.

I did also poke around a bit and I'm not sure if the GitHub actions configuration options/behavior changed at some point, but we aren't actually assigning the Python version on which to run the tests (non-unique to this PR).  I think I can fix that relatively quickly though.

kafitzgerald:
And the readthedocs fail was just a fluke 🎉 .  Thanks @anissa111 for helping me get that sorted!  

I still need to put in that CI PR to fix the Python version thing, but this should be good.  Happy to wait thought if we'd rather fix CI first.

jukent (approved):

jukent:
Great! I'm curious why it's now not getting failures when your commit didn't touch those files.

kafitzgerald:
Wait, which failures?  The new gv function ones?

I pulled out the unpinning of matplotlib, which caused the initial failures.  CI was then using Python 3.11 and a newer matplotlib for everything so must be related to something in that env.  With the matplotlib version pin now it's using Python 3.10 and passing.

jukent:
Yes the new geocat viz ones. Glad it's working!

kafitzgerald:
Yeah, I imagine those might come back up once we unpin matplotlib and the tests run on a broader range of Python/dep versions.

PR #521, opened 2023-06-21
Adjust notebook plot sizing

### LINKED ISSUE #519: Notebook example's format seems like broken after the new docs style
**Describe the bug**
The plots in this [notebook](https://geocat-examples.readthedocs.io/en/latest/gallery-notebooks/Datashader/MPAS_Datashader_Trimesh.html) needs to have a better formatting for better readibility

**To Reproduce**
Steps to see the behavior:
1. Go to https://geocat-examples.readthedocs.io/en/latest/gallery-notebooks/Datashader/MPAS_Datashader_Trimesh.html
2. Check the plots through the end of the page, e.g. check out the screenshot below.

**Expected behavior**
Maybe fit the plot into the screen/column size?

**Screenshots**
image

kafitzgerald:
Adjust notebook plot sizing to look cleaner with the new theme.  

For context: I was hoping to be able to use the responsive options that would adjust with the window size, but unfortunately those don't seem to work well in this case.  I could only get the responsive width to work, but that was distorting the visualizations.  Setting the width in pixels to something that should look reasonable in most cases was a compromise.  

This PR also adds needed dependencies (e.g. geoviews, holoviews, datashader) to the environment so that everything builds with the clean notebook (no cells run ahead of time).  This may take a little longer to build for testing and obviously adds some dependencies so open to pulling this from the PR.  

Closes #519.

### DIFF
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -5,10 +5,12 @@ dependencies:
   - python>=3.8, <3.11  # minimum support 3.8, 3.11 not yet fully supported
   - geocat-comp
   - cf_xarray>=0.3.1
+  - datashader
   - geocat-datafiles
   - eofs
   - fiona
   - geocat-viz
+  - geoviews
   - holoviews
   - cartopy
   - geographiclib
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - eofs
   - fiona
   - geocat-viz
+  - holoviews
   - cartopy
   - geographiclib
   - jupyter
--- a/docs/gallery-notebooks/Datashader/MPAS_Datashader_Trimesh.ipynb
+++ b/docs/gallery-notebooks/Datashader/MPAS_Datashader_Trimesh.ipynb
@@ -66,8 +66,8 @@
     "hv.extension(\"bokeh\",\"matplotlib\")\n",
     "\n",
     "opts.defaults(\n",
-    "    opts.Image(responsive=True, aspect='equal'),\n",
-    "    opts.RGB(responsive=True, aspect='equal'))\n"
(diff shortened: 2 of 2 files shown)


kafitzgerald:
Unfortunately (and perhaps unsurprisingly) this didn't render as well in the docs.  https://geocat-examples--521.org.readthedocs.build/en/521/gallery-notebooks/Datashader/MPAS_Datashader_Trimesh.html

Marking as a draft until I have a chance to look into other options.

kafitzgerald:
A side note - this notebook isn't in the most intuitive location.  

Seems like it should probably live outside of docs w/ (or alongside) the other content, but maybe I'm missing something

anissa111 (approved):
Surprisingly, this does not look like it has significantly increased the build times for CI. I was worried about the regeneration and additional dependencies, but I don't think it's actually making that large of an impact

kafitzgerald:
> Surprisingly, this does not look like it has significantly increased the build times for CI. I was worried about the regeneration and additional dependencies, but I don't think it's actually making that large of an impact

Yeah, I was a bit surprised as well.  It's non-zero, but not significant.  Does make me wonder where most of the time is spent, but I'll save that question for later.

erogluorhan:
> A side note - this notebook isn't in the most intuitive location.
> 
> Seems like it should probably live outside of docs w/ (or alongside) the other content, but maybe I'm missing something

You are right! When initially adding these notebooks into the gallery, I couldn't get it to be rendered from under a folder in the root directory (e.g. a "Notebooks" directory just like "Plots"). If that can be fixed, such a location is ideal.

erogluorhan (approved):

jukent:
Thanks for this notebook @kafitzgerald !

PR #540, opened 2023-09-22
Remove unneeded xarray version pin

### LINKED ISSUE #495: Unpin xarray version once root cause is found

kafitzgerald:
Closes #495.

### DIFF
--- a/conda_environment.yml
+++ b/conda_environment.yml
@@ -32,7 +32,7 @@ dependencies:
   - sphinx-design
   - sphinx-gallery
   - sphinx_rtd_theme
-  - xarray<=2023.02.0
+  - xarray
   - wrf-python
   - pip:
       - pre-commit


kafitzgerald:
It looks like there are a few things going on here. 

For the Python 3.8 build on Ubuntu, it looks like we're getting sphinx 7, which is causing this error: https://github.com/executablebooks/sphinx-book-theme/issues/749

And this is not a problem, but worth noting - currently we don't have Python 3.8 version of geocat-comp so the Python 3.8 builds here are grabbing an older version of geocat-comp (and thankfully xarray).

kafitzgerald:
In the other builds it's not making it that far and failing instead on the contour labeling in this example: gallery/Contours/NCL_coneff_8

I haven't sorted out why this is the case yet.

jukent (approved):

kafitzgerald:
Confirming this is only failing because of our known RTD issue.  @jukent we can probably merge whenever.

PR #542, opened 2023-09-27
Clean up environment file to mitigate sphinx version problem

kafitzgerald:
Clean up environment file to mitigate sphinx version problem:
* Removes old sphinx theme from the environment file and related template file
* Removes jupyter and sphinx from the environment file so they're just installed as dependencies (hopefully this mitigates the sphinx version issue we are having)

Closes #541.

PR #545, opened 2023-09-27
Add environment cache key to CI config file

### LINKED ISSUE #543: Fix environment caching for CI
Right now we're caching (and reusing) some environments for longer than intended.  

I can't recall if maybe the defaults were different once upon a time or where I got the current setup from, but regardless it needs an update.

kafitzgerald:
Add environment cache key to ci.yml configuration file to fix our environment caching trouble.

Closes #543.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,9 @@ jobs:
     steps:
     - name: checkout
       uses: actions/checkout@v3
+    - name: Set environment variables
+      run: |
+        echo "TODAY=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
     - name: setup micromamba
       uses: mamba-org/setup-micromamba@v1
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,7 @@ jobs:
          environment-name: geocat-examples
          environment-file: conda_environment.yml
          cache-environment: true
+         cache-environment-key: "${{runner.os}}-${{runner.arch}}-py${{matrix.python-version}}-${{env.TODAY}}"
          create-args: >-
            python=${{matrix.python-version}}
     - name: make html


kafitzgerald:
Failing because #542 needs to get merged.

.github/workflows/ci.yml
```diff
@@ -37,6 +37,7 @@ jobs:
          environment-name: geocat-examples
          environment-file: conda_environment.yml
          cache-environment: true
+         cache-environment-key: "${{runner.os}}-${{runner.arch}}-py${{matrix.python-version}}-${{env.TODAY}}"
```
anissa111, line 40: It doesn't look like all the info in the cache key is making it to the cache tag.

This is a cache from this brach:
image

A cache from xarray using a similar key format looks like this:
image

It looks like neither the python version or the date is actually making it in.

I think we need to explicitly set the date in the github environment, see [this part in xarray's workflow file](https://github.com/pydata/xarray/actions/runs/6309646247/workflow?pr=8229#L66-L68). They also similarly [set the python version](https://github.com/pydata/xarray/actions/runs/6309646247/workflow?pr=8229#L92) in the github environment, though I'm not sure why your `matrix.python-version` didn't work here.

Also, forget what I said about the environment file hash, the caching automatically appends the file hash to the end anyway, so we don't need to specify it.

kafitzgerald: Alright, looks like this is working better now. The OS bit is duplicated in the key, but it's nice to have up front and I don't think it hurts anything.  xarray is set up the same way.   

I think it was just failing back to the default key before. 

I'll try to keep an eye on how this is working for us as well moving forward.

anissa111: > I think it was just failing back to the default key before.

Ahh, that makes sense.

anissa111 (approved):
Looks good assuming the tests go through as expected!

PR #549, opened 2023-10-03
Contributor's guide

kafitzgerald:
Still working on porting over some items from CONTRIBUTING.md, but can get this in soon.  

In the meantime feedback welcome.


kafitzgerald:
We should probably squash this when we merge (the commit history is pretty rough looking and there's not a lot of value in keeping it as is) or I can do that ahead of time if you'd prefer.

jukent:
Squashed and merged

Issue #528, opened 2023-07-10
Change text on gallery homepage


jukent:
Think about the direction we want geocat-examples to go in and make sure this text is in line with that.

kafitzgerald:
Might want to update the README and repo description as well (unless they're general enough already).

Issue #524, opened 2023-06-29
Move notebooks out from the docs folder


kafitzgerald:
Example notebooks currently live inside the docs folder in not the most intuitive location.  We might consider moving these out (i.e. alongside the other examples).

jukent:
Move these to a Project Pythia Cookbook

jukent:
Keeping these notebooks through October 9th for the ECMWF conference.

PR #556, opened 2023-11-06
Change text, and remove bad notebooks

jukent:
Closes https://github.com/NCAR/geocat-examples/issues/528 and https://github.com/NCAR/geocat-examples/issues/524


kafitzgerald:
Did we want to hang on to the GeoCAT-comp examples until they live in GeoCAT-applications?

jukent:
Okay, let's try to move those over on Wednesday?

jukent:
Decided we only need to move over the fourier example because MPAS goes to Pythia, linint1 and 2 are unsupported / from f2py and are going to be completely redone in geocat-applications (that shouldn't hold this up), and the comp only examples are already in comp.

kafitzgerald (approved):

PR #557, opened 2023-11-08
Remove Gallery/Topography/NCL_topo_1.py

kafitzgerald:
Remove NCL_topo_1.py temporarily so we can get Read The Docs building again.  

Other topography examples should have similar enough content for the time being.


kafitzgerald:
🎉

jukent (approved):

PR #559, opened 2023-11-09
Panel_35 to use data from geocat.datafiiles instead of creating on the fly with generate_2d_array

### LINKED ISSUE #558: Change panel_35 to load data from geocat-datafiles
Not implementing generate_2d_array as in https://github.com/NCAR/geocat-viz/pull/175 because it is too niche and doesn't need to be re-run.

Instead adding data to geocat datafiles in https://github.com/NCAR/geocat-datafiles/pull/57

Move this code to geocat-scratch and change example to load data.

jukent:
Closes #558

Currently doesn't like how the data is formatted. Need to figure that out. If my datafiles need to be regenerated the code is here: https://github.com/NCAR/geocat-scratch/blob/main/geocat-examples/generate_2d_array.py


kafitzgerald:
> Currently doesn't like how the data is formatted. Need to figure that out. If my datafiles need to be regenerated the code is here: https://github.com/NCAR/geocat-scratch/blob/main/geocat-examples/generate_2d_array.py

Should this get updated as well?

kafitzgerald:
> > Currently doesn't like how the data is formatted. Need to figure that out. If my datafiles need to be regenerated the code is here: https://github.com/NCAR/geocat-scratch/blob/main/geocat-examples/generate_2d_array.py
> 
> Should this get updated as well?

Otherwise, this looks good!

jukent:
No that was added there- thrown into scratch in case we need it again for something similar. No update needed

kafitzgerald:
Oh, I see.   That was the original code that this removed and not the code specifically to create the netCDF files themselves (that are now being used).

Sorry for the confusion on that.

kafitzgerald (approved):
Looks good!

Makes this so much more clear!

jukent:
It was the data saving step I had to adjust which isn't represented, but I think that's fine.

Issue #562, opened 2023-11-09
Repeated CI failures due to timeouts


kafitzgerald:
**Describe the bug**
Our CI workflows not infrequently fail by timing out after a download from geocat-datafiles hangs.  I'm not sure if it's always on the `wrfout_d03_2012-04-22_23_00_00_QV.nc` file, but it often seems to be.  

It seems to be the MacOS runners most commonly, but a linux one failed today as well in a similar fashion.  It seems like the failures are getting more frequent as well.  

Usually re-running the failed workflow will get it to pass.  

**To Reproduce**
Trigger a CI workflow and likely watch one or two fail.  

**Screenshots**
<img width="699" alt="Screen Shot 2023-11-09 at 1 09 19 PM" src="https://github.com/NCAR/geocat-examples/assets/7872563/c6fc4443-1d3e-4711-9258-db00940fc35b">

kafitzgerald:
Something like what @anissa111 did in NCAR/geocat-comp#532 could help with this.  

We should do something similar for the upstream dev CI as well.

kafitzgerald:
The other thing we might think about is caching the test files somehow.

kafitzgerald:
Seems to be helping.  I'm going to close this for now.

PR #581, opened 2024-01-09
Add a retry step to the flaky part of our CI workflow

kafitzgerald:
Add a retry step to the flaky part (creating the HTML) of our CI workflow

Partially addresses #562. 

I only added the retry step in ci.yml and not in the upstream-dev CI, since that doesn't seem to have the same problem.  I think it's the additional data downloads from multiple version/OS CI runs that's problematic here.  

I also am leaving the issue open since I think we should look into some sort of caching strategy to really address this.  That might need to wait a moment though as I'm pretty swamped.


kafitzgerald:
I spoke too soon on the lack of trouble with the upstream-dev-ci.  That just failed last night for a similar reason.  

I'll add a retry step on that as well.

jukent (approved):
I like this approach, we know if it is likely to time out sooner and can rerun it.

jukent:
Want to see if the retry solves our problems here before applying it to upstream-ci?

kafitzgerald:
> Want to see if the retry solves our problems here before applying it to upstream-ci?

Should have referenced it, but I was testing on my fork and the flaky step does time out and the retry step works.  

I can turn this on again and link to it when I add the upstream ci bit.

kafitzgerald:
Here are the links to the test runs on my fork: 

CI Upstream: https://github.com/kafitzgerald/geocat-examples/actions/runs/7492946925
CI: https://github.com/kafitzgerald/geocat-examples/actions/runs/7492945543

The 3.11 ubuntu-latest CI run used the retry step - others just ran through.

PR #598, opened 2024-03-25
Taylor Diagram example update with new methods

jukent:
No longer calling add_xgrid() or add_ygrid().

Might have to wait until next release before merge.


jukent:
I think this is good to go @kafitzgerald

kafitzgerald (approved):

PR #599, opened 2024-03-27
Demonstrate how to mask out lakes from land feature in NCL_mask2

jukent:
Adds lake masking to the land mask in https://github.com/NCAR/geocat-examples/pull/189, so that the contourf plot still fills in lake as well as ocean water elements.

Closes https://github.com/NCAR/geocat-viz/issues/27 (won't automatically)

mask2_linewidth


kafitzgerald (commented):
It was bugging me that so much iteration was required here so I looked into shapely a bit and it turns out you can do diffs on GeometryCollections as well as individual geometries.  It requires an import from shapely, but I think it might be worth it to tidy up the example a bit.

You can also use the [`with_scale`](https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.feature.NaturalEarthFeature.html#cartopy.feature.NaturalEarthFeature.with_scale) method to grab datasets w/ different scales with a bit less overhead.  Not sure if that's new or not, but also makes things a bit cleaner looking.

```
from shapely import GeometryCollection

land = GeometryCollection(list(cfeature.LAND.with_scale('110m').geometries()))
lakes = GeometryCollection(list(cfeature.LAKES.with_scale('110m').geometries()))
land_no_lakes = land.difference(lakes)

ax.add_geometries(land_no_lakes,
                  crs=ccrs.PlateCarree(),
                  facecolor='lightgray',
                  edgecolor='black',
                  linewidth=0.5,
                  zorder=1)
```

jukent:
Thanks @kafitzgerald! I was hoping something like that existed but wasn't finding it. Thanks, that's way better.

kafitzgerald (changes requested):
Looks a notebook got edited / added instead of the script?

jukent:
Oh no. fixed. I have to give my notebooks weirder names next time.

kafitzgerald (approved):
No worries.  Looks great!

PR #610, opened 2024-07-25
Placeholder PR to highlight code changes

jukent:
Doing dev work in Jupyter notebooks. Making PR to share curly vector code.


kafitzgerald (commented):
This is really cool!  

It looks like the length scales of the vectors don't match the originals though: https://www.ncl.ucar.edu/Applications/Images/vector_1_lg.png

I think this is because of more conceptual differences between streamplot (integrates forward or backward until some constraint) and curly vectors (length varies with vector magnitude).  It's handled length normalization in the [various approaches here](https://stackoverflow.com/questions/51843313/flow-visualisation-in-python-using-curved-path-following-vectors), but I don't know how robust that is.

PR #615, opened 2024-10-09
Address SciPy interp2d deprecation

### LINKED ISSUE #607: Remove deprecated SciPy functionality
It looks like we need to update NCL_vector_5.py. 

See here: https://github.com/NCAR/geocat-examples/actions/runs/9239185399/job/25418034302

kafitzgerald:
Addresses the SciPy `interp2d` deprecation by using xarray's `interp` method instead.

Closes #607

This doesn't address the bigger issues I stumbled upon in looking into this of matplotlib's streamlines requiring input data on a regular grid (this should probably be addressed as an upstream contribution) and that these (streamlines vs NCL curly vectors) are different ways of representing data.  While they look similar, they aren't showing information in the same ways and that should probably be documented.  The reference vector is also incorrect.  I'll log this as an issue.

### DIFF
--- a/Gallery/Vectors/NCL_vector_5.py
+++ b/Gallery/Vectors/NCL_vector_5.py
@@ -119,19 +119,15 @@ colors = T.plot.contourf(ax=ax,
 
 # We attempt to recreate curly vectors with Matplotlib's streamplot function.
 # streamplot requires the input parameter x, y to be evenly spaced strictly increasing arrays.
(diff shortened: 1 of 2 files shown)


jukent (approved):
Thanks for fixing the deprecation. We have a mountain of vector issues ahead of us.

kafitzgerald:
> Thanks for fixing the deprecation. We have a mountain of vector issues ahead of us.

Indeed 😞

Issue #314, opened 2021-02-27
Curly vector plot ?


fcodron:
Hello,

Is there any plan to implement "curly vector" plots in some way for "standard" Python (matplotlib, cartopy...) ?
I guess through a modification of streamlines ?

This plot type is fairly unique to NCL (as far as I know), and very useful especially for oceanography plots, where strong currents tend to flow along the coastlines, and vectors ending over land are not very nice...  

Thanks,

Francis

michaelavs:
Hi Francis, 
I don't think there is an explicit plan in place for implementing the curly vector, but it could become part of later releases of GeoCAT software. I've been looking into if there is an already established Python version for curly vectors and it looks like currently the closest would be streamlines (which as you probably know, is slightly different than the curly vector from NCL). I'll bring this issue up with the rest of the team later this week and see what they think about it and get back to you then!

fcodron:
Hello,

Thanks for replying !

I am only aware of this try by Kieran Hunt (U. Reading):
https://github.com/kieranmrhunt/curved-quivers
Which is I think a modification of standard MPL streamlines. I am not sure
if it can be made to work with maps. It was inspired by this stackoverflow
thread :
https://stackoverflow.com/questions/51843313/flow-visualisation-in-python-using-curved-path-following-vectors

Otherwise the curly vect seems pretty unique to NCL...

I'm attaching the figure of IPSL currents in the Arctic so you see what I
mean, perfect for ocean eddies too !

Thanks,

Francis

[image: Arctic.png]

pilotchute:
@fcodron I've just been assigned to look into this. 
From my understanding, you curly vectors are arrows in a grid, that are scaled by the magnitude and curved based on the curl derivative of the field. 
So I think, it's the curl of the arrows that differentiates it from a standard quiver plot.
Though looking closely at the IDL output for curly vectors it actually seems more similar to broken up lines of a streamplot(), with the broken up arrows scaled by magnitude. 

Could you in your own words (or with a sketch) show what you'd be looking for in a curly vector plot?

fcodron:
Hello,

Nice to hear you'll get into this !

Curly vectors are originally part of NCL visualization (don't know about IDL), where it's an option in vector plots. You can see an example image here : [https://www.ncl.ucar.edu/Applications/Images/vector_4_lg.png](url)

[...]

PR #618, opened 2024-10-14
Temporarily remove incorrect NCL_vector_5.py example

### LINKED ISSUE #616: NCL_vector_5.py currently misrepresents the data
At a minimum the reference vector for the NCL_vector_5.py visualization is incorrect, but I'm a little concerned the overall visualization might be misrepresenting the data to some degree.

The original NCL example leveraged the curly vectors functionality which the python example here tries to reproduce using matplotlib's streamplot.  However, streamlines are a rather different type of visualization and that doesn't seem to be accounted for or well documented here.

kafitzgerald:
Removes the (incorrect) NCL_vector_5.py example until we have a working curly vector implementation.  

Closes #616
Relates to #314

If we'd rather not remove this for now, we could also look into using `quiver` from Matplotlib like some of the other vector examples, but I don't think that's a great option for this particular example.  The vector field is rather complex and I think it'd be difficult to represent with that plot type.  A `streamplot` is another option, but that's really a different type of visualization.


kafitzgerald:
This is failing pre-commit because of https://github.com/PyCQA/docformatter/issues/293

jukent (approved):

jukent:
I'm in favor of removing it, as you showed here. Or adding some sort of annotation that the example is not displaying scientifically significant vector lengths. I wonder if there are people using this example as a template for other visualization needs (pretty sure they're captured in a different example).

kafitzgerald:
> I'm in favor of removing it, as you showed here. Or adding some sort of annotation that the example is not displaying scientifically significant vector lengths. I wonder if there are people using this example as a template for other visualization needs (pretty sure they're captured in a different example).

Yeah, we have other streamplot examples already so I think the best thing to do here is remove it while we address the issue.

PR #624, opened 2024-11-08
Theme configuration fix

kafitzgerald:
Merges theme configuration dictionaries to address issues cause by the second dictionary overwriting the first.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -104,26 +104,20 @@ html_favicon = '_static/images/GeoCAT_square.svg'
 html_static_path = ['_static']
 html_css_files = ["theme_overrides.css"]
 
-html_theme_options = dict(
-    # canonical_url="",
-    repository_url="https://github.com/NCAR/geocat-examples",
-    repository_branch="main",
-    path_to_docs="docs",
-    use_edit_page_button=True,
-    use_repository_button=True,
-    use_issues_button=True,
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -104,26 +104,29 @@
...
+    "home_page_in_toc":
+        False,
+    "navbar_footer_text":
+        "",
     "logo": {
         "image_light": '_static/images/NSF_NCAR_light.svg',
         "image_dark": '_static/images/NSF_NCAR_dark.svg',
-    }
+    },
+    "extra_footer":
```
jukent, line 128: Your PR description sounds like it removes the extra footer but isn't this it? Looks mostly like the formatting of `html_theme_options` changed

kafitzgerald: Yeah, that wasn't the most clear and I don't think you saw this on applications.

It's the other way around.  The extra footer config and other other items weren't being applied because they were overwritten by the second dictionary definition.  

I could also have just modified the dictionary below to address this, but it seem cleaner just to merge it all into one and match how the dictionaries are defined in the theme docs.  That also introduces more changes though so I think it's a little harder to understand what's going on from the diffs.

kafitzgerald: Pre-commit also formatted this in a way I don't love, but that's probably for another day.

jukent (approved):

PR #629, opened 2024-12-10
plt.scatter needs norm argument in Station 2

mnlevy1981:
Per [a comment](https://github.com/NCAR/geocat-examples/pull/49#issuecomment-2532219356) on the original PR to bring in the station 2 example, the colors weren't actually assigned based on the randomized values.

This will change the images produced by this example; originals are in #49, the updated colors are

labels with norm

colorbar with norm

It's been a while since I've contributed to this repo, let me know if I need to update more than just the `.py` script :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .DS_Store
 .ipynb_checkpoints/
 *.grd.gz
+**/__pycache__/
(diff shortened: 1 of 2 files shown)


kafitzgerald (approved):
Thanks for the quick fix!

Issue #619, opened 2024-10-14
Transition to Ruff


kafitzgerald:
Add [docformatter](https://github.com/PyCQA/docformatter) back to pre-commit once there's a v1.7.6 release.

Relates to #618 (where docformatter was removed).

kafitzgerald:
Not a high priority, but we may want to look into Ruff to replace this.

PR #648, opened 2025-05-08
codespell pre-commit

cyschneck:
Fix code misspellings with `codespell`, as part of the transition to ruff

Related to #619


jukent (approved):
Wow that's a lot of spelling errors! Thanks for catching all of these and for the codespell pre-commit.

kafitzgerald (approved):
Looks great.  Thanks for adding this in!

PR #649, opened 2025-05-09
Ruff format

### LINKED ISSUE #619: Transition to Ruff

cyschneck:
Updates pre-commit to transition from `docformatter` to `ruff`

Lot of files, but most of them are just formatting:
- Formatting, nothing functionality
- Adds in `ruff.toml` to handle ruff formatting rules

Ruff ignores `F841` (Local variable name is assigned to but never used) since multiple matplotlib examples setup fig or contour plots, while the variable `fig` is not used

Closes #619

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,17 +1,16 @@
-repos:
--   repo: https://github.com/google/yapf    # To format the code to conform YAPF
-    rev: v0.43.0
-    hooks:
-    - id: yapf
-      args: ['--in-place', '--recursive', '--style', 'google']
-      require_serial: true
+ci:
+    autofix_prs: false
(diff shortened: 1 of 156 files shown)


jukent (approved):
I looked through a good amount of the files (enough to trust that these changes are not more than stylistic ones pushed by ruff). I do prefer this formatting a lot more.

kafitzgerald (approved):
Looks good to me.   Thanks for doing this! 

I think the only thing I noticed that's a little funky while skimming through some was a few of the data arrays are now quite long, but overall I think this is much better and we can handle that later if it gets bothersome.

PR #694, opened 2026-05-17
Fix failure issue generation and remove CI Upstream badge

kafitzgerald:
Fixes the failure issue generation and removes the CI Upstream badge from the README since it's no longer needed and not really important for users.


kafitzgerald:
I'm going to go ahead and merge this so we can make sure the failure issues are working tomorrow when the workflow runs.

